PR #38, opened 2015-05-21
Add scritps to manage assets

jackscotti:
2 scripts:

first: allows to upload a single file on asset-manager
second: allows to update a single file on asset-manager

Work required by this ticket:
https://www.pivotaltracker.com/story/show/94847618

### DIFF
--- a/bin/update_asset
+++ b/bin/update_asset
@@ -1,0 +1,29 @@
+#!/usr/bin/env ruby
+
+require File.expand_path("../../config/environment", __FILE__)
+
+old_asset_id = ARGV.any? ? ARGV.fetch(0) : nil
+filename = ARGV.any? ? ARGV.fetch(1) : nil
+
+unless old_asset_id
+  puts "You need to provide the asset ID as first argument when running this script"
+  abort
+end
+
+unless filename
+  puts "You need to provide a filename as second argument when running this script"
+  abort
+end
+
+file = File.new(filename)
+old_asset = Asset.find(old_asset_id)
+
+if old_asset.update_attributes(file: file)
+  puts "Updated!"
+  puts "Asset id: #{old_asset.id}"
+  puts "Asset name: #{old_asset.file.filename}"
+  puts "Asset basepath: /media/#{old_asset.id}/#{old_asset.file.filename}"
+else
+  puts "not updated, something went wrong"
+  puts "errors: #{old_asset.errors.full_messages}"
+end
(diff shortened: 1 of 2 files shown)


bin/update_asset
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env ruby
+
+require File.expand_path("../../config/environment", __FILE__)
+
+filename = ARGV.any? ? ARGV.fetch(0) : nil
+old_asset_id = ARGV.any? ? ARGV.fetch(1) : nil
```
evilstreak: It would be good to reverse the order of these arguments, to match the [API adapter](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/asset_manager.rb#L33).

bin/update_asset
```diff
@@ -0,0 +1,32 @@
...
+
+unless filename
+  puts "You need to provide a filename as first argument when running this script"
+  abort
+end
+
+unless old_asset_id
+  puts "You need to provide the old ID path as second argument when running this script"
```
evilstreak: I think it would be better to refer to this as the "asset ID" instead of the "old ID path"

evilstreak:
A couple of minor comments, but this looks good :+1:

bin/update_asset
jackscotti: yep, when I wrote that I thought I was going to need the path of the previous asset. I was wrong and forgot to update it!

jackscotti:
Done what you've suggested :+1:

bin/update_asset
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env ruby
+
+require File.expand_path("../../config/environment", __FILE__)
+
+old_asset_id = ARGV.any? ? ARGV.fetch(0) : nil
+filename = ARGV.any? ? ARGV.fetch(1) : nil
+
+unless old_asset_id
+  puts "You need to provide the asset ID as second argument when running this script"
```
evilstreak: This should be "first" now

bin/update_asset
```diff
@@ -0,0 +1,29 @@
...
+  abort
+end
+
+unless filename
+  puts "You need to provide a filename as first argument when running this script"
```
evilstreak: This should be "second" now

jackscotti:
I am sorry, I've been really thick. Fixed now.

evilstreak:
You haven't been thick! It was a pretty easy thing to overlook.

PR #42, opened 2015-10-09
Upgrade to rails 4 2

h-lame:
For: https://trello.com/c/3HCOXMbf/159-upgrade-rails-on-asset-manager

In which we upgrade the app to rails 4.2.  Along the way we have to upgrade pretty much every other dependency of the app in order to get a properly running test-suite and app: Mongoid, RSpec, Delayed Job, Carrierwave, and even swap state_machine for state_machines-mongoid.  There are some extra notes about each upgrade step in the commits.

I guess this may require deployment changes?  Let me know if it does and I'll get a PR in for them too.


h-lame:
FWIW: I've tested this on the VM via travel-advice-publisher and I was able to upload an image and view it again (after I ran the workers).

boffbowsh:
Looks good to me apart from one tiny comment.
What tool did you use to apply the boilerplate from new rails to this app? Or was it just `rails new`?

h-lame:
@boffbowsh I used [railsdiff](http://railsdiff.org/) to see the changes and then copy+paste from there. Basically the same as doing a diff on a `rails new` copy.  Nothing fancy.

issyl0:
This needs a `config.secret_key_base` for production adding to `alphagov-deployment`, which I've done in https://github.gds/gds/alphagov-deployment/pull/1088.

boffbowsh:
In that case :+1: from me but I don't have a huge amount of context on the app itself

h-lame:
Thanks @issyl0!

dsingleton:
It would be worth deploying this branch to preview before merging, to test the production-env deployment config. (apologies if this has already been discussed)

issyl0:
@dsingleton Thanks. It hadn't, and also I don't know how deploying alphagov-deployment to get the new secret keys works.

dsingleton:
I've opened a PR [alphagov-deployment](https://github.gds/gds/alphagov-deployment/pull/1099) updating the `mongoid.yml` config for Mongoid 4, once that is merged this can be tested on preview.

dsingleton:
> I've opened a PR alphagov-deployment updating the mongoid.yml config for Mongoid 4, once that is merged this can be tested on preview.

I've closed that PR and made the change in [puppet](https://github.gds/gds/puppet/pull/3596/files) instead, through ENV vars, which has been merged and i've updated [`mongoid.yml` here](https://github.com/alphagov/asset-manager/commit/9c14c710a89a1ebb15d5c5c5af5803d0e6a5dfca) to use the ENV var approach - note i've not got asset-manager running locally to test that, so it'd be good if someone else could give it a run.

I've created a branch of [alphagov-deployment](https://github.gds/gds/alphagov-deployment/pull/1100) with the asset-manager `mongoid.yml` removed, so it won't overwrite the new approach here.

To test that on preview we'll need to use the branch of `alphagov-deployment` during the deploy to preview. Once we're happy we can merged _everything_.

davidbasalla:
Just tested this branch with travel advice publisher by attaching an image and I couldn't see any problems with it :+1:

dsingleton:
I've tested this on preview, with the [`alphagov-deployment` branch](https://github.gds/gds/alphagov-deployment/pull/1100), which deployed without errors and I was able to upload an image via `travel-advice-publisher` without error.

I've booked a deploy slot, which will requiring merging this, and the above PR, plus making sure the [puppet change](https://github.gds/gds/puppet/pull/3596) has propagated to staging/production.

PR #45, opened 2016-01-08
Improve readme

tadast:
It took me a while to understand why I could see my uploaded assets in dev environtment, so here's an improved readme that should save time for other developers looking at this.

I've also added a MIT licence because this repo didn't seem to have any.

In addition to this PR the repo could benefit from adding a one line description on GitHub (someone with admin access to the repo should be able to do that)


jamiecobbett:
I've opened a PR to run the worker automatically when running asset-manager with bowler: https://github.gds/gds/development/pull/287

PR #51, opened 2016-09-16
Soft delete assets

steventux:
Part of https://trello.com/c/zLrCEavw/263-delete-attachments-feature-medium

Adds the route DELETE #destroy for assets.
Assets are soft deleted using the `mongoid_paranoia` gem. Requests to show a soft deleted asset or download media belonging to a soft deleted asset will respond with 404 as this is the default behaviour for rescued `Mongoid::Errors::DocumentNotFound` [in the application controller](https://github.com/alphagov/asset-manager/blob/master/app/controllers/application_controller.rb#L10).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,6 +26,8 @@ gem 'plek', '1.3.0'
 gem 'logstasher', '0.4.8'
 gem 'rack_strip_client_ip', '0.0.1'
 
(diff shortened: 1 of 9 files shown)


tuzz:
This approach looks fine. 👍

It has the same effect as moving the files to another directory on the server.

PR #52, opened 2016-10-05
Restore assets

steventux:
Part of: https://trello.com/c/Vn2SpooI/279-document-attachments-still-visible-when-document-is-unpublished-large

Soft deleted assets can be restored via the `/assets/:id/restore` route.
This allows withdrawn content with soft deleted assets to be unwithdrawn and their assets restored.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,5 +1,6 @@
 Rails.application.routes.draw do
   resources :assets, :only => [:show, :create, :update, :destroy]
+  post "/assets/:id/restore", to: "assets#restore"
 
   get "/media/:id/:filename" => "media#download", :constraints => { :filename => /.*/ }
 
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -1,5 +1,6 @@
 Rails.application.routes.draw do
   resources :assets, :only => [:show, :create, :update, :destroy]
+  post "/assets/restore/:id", to: "assets#restore"
```
gpeng: Would it be more 'correct' to `post "/assets/:id/restore"` here?

spec/models/asset_spec.rb
```diff
@@ -216,5 +216,10 @@
       asset.destroy
       expect(asset.deleted_at).not_to be_nil
     end
+
+    it "can be restored" do
+      asset.restore
```
gpeng: `asset.deleted_at` is already `nil` here so this test will pass if `asset.restore` does nothing.

steventux:
Thanks @gpeng, the routing suggestion makes much more sense now :+1: Amended that flaky test too.

PR #57, opened 2017-03-03
Add Jenkinsfile

steventux:
Move Asset Manager to new CI.
Depends on https://github.com/alphagov/govuk-puppet/commit/8a2696f6c7572e477cfcc2989a33be7dc525b167 being deployed to CI agents.


Jenkinsfile
```diff
@@ -0,0 +1,67 @@
...
+    [$class: 'RebuildSettings', autoRebuild: false, rebuildDisabled: false],
+  ])
+
+  try {
+    if (!govuk.isAllowedBranchBuild(env.BRANCH_NAME)) {
+      return
+    }
+
+    stage("Checkout") {
+      checkout scm
```
gpeng: Is this where we need to use the SSH checkout? https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy#L18

PR #63, opened 2017-07-12
Upload assets to S3 [DO NOT MERGE]

chrislo:
We're hoping to change asset-manager so that it stores its assets in, and serves its assets from S3. As a first step, this PR implements uploading of assets that have passed virus checking to S3. 

The assets are uploaded to S3 while still being stored in the existing NFS store. Assets are still served from NFS so there should be no external change to how the API works with this PR. This allows us to work out how to integrate with S3 in production without altering the existing functionality. 

I've added an `S3Uploader` class to handle uploading the assets to a bucket when virus checking has successfully completed. To test this in development you need to set the following environment variables:
 - `AWS_ACCESS_KEY_ID`
 - `AWS_SECRET_ACCESS_KEY`
 - `AWS_REGION`
 - `BUCKET_NAME`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,6 +28,8 @@ gem 'rack_strip_client_ip', '0.0.1'
 
 gem 'mongoid_paranoia', '0.2.1'
 
+gem 'aws-sdk', '~> 2'
+
 # bundle exec rake doc:rails generates the API under doc/api.
 gem 'sdoc', '~> 0.4.0', group: :doc
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -41,6 +41,14 @@ GEM
       multi_json
     arel (6.0.3)
     ast (2.3.0)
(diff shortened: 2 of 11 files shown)


chrislo:
@floehopper @chrisroos - I'm fairly happy with this now as a first pass. I think we should hold off merging it to `master` at the moment. It shouldn't break anything if we do, but the `upload_to_s3` jobs will fail until the correct environment variables are set. Could you take a look, and if we're happy perhaps the first thing to do is deploy this branch to integration to test it against the AWS configuration when we have it in place.

floehopper:
@chrislo: I've just added a new commit which allows us to optionally "stream" an asset from S3 vs NFS. I've added quite a lot of notes to the commit note to explain my rationale.

floehopper:
Closing in favour of #74.

PR #70, opened 2017-07-18
Make more idiomatic use of RSpec

floehopper:
I've done a bit of tidying up of the specs, e.g.

* Use `subject` or `let` vs instance variables
* Use `before` vs `before :each`

### DIFF
--- a/spec/controllers/assets_controller_spec.rb
+++ b/spec/controllers/assets_controller_spec.rb
@@ -9,21 +9,21 @@ RSpec.describe AssetsController, type: :controller do
 
   describe "POST create" do
     context "a valid asset" do
-      let(:atts) { { file: load_fixture_file("asset.png") } }
+      let(:attributes) { { file: load_fixture_file("asset.png") } }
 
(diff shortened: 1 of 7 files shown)


spec/controllers/assets_controller_spec.rb
```diff
@@ -220,23 +220,22 @@
     end
 
     describe "cache headers" do
+      let(:asset) { FactoryGirl.create(:asset) }
+
       it "sets the cache-control headers to 0 for an unscanned asset" do
-        asset = FactoryGirl.create(:asset, state: 'unscanned')
```
chrislo: Is `unscanned` the default here, or do we need a second `let` block?

floehopper: Yeah. It's the default. I contemplated leaving the state set explicitly within the `let` block or introducing an `unscanned_asset` factory, but it felt like overkill. I'm happy if you want to look into doing that, but I think there are probably a bunch of other places where we'd need to use it.

PR #74, opened 2017-07-19
Store assets on S3 as well as NFS and optionally serve from S3 on public URLs

floehopper:
This is the first step in changing the Asset Manager to store its assets in and serve its assets from AWS S3, i.e. to avoid using NFS.

This PR means that whenever an asset is successfully virus scanned a new asynchronous job is kicked off to upload the asset to S3. The fact that it's a separate job should mean that the current external behaviour is unaffected; the assets are still stored on NFS as before.

Also if `stream_from_s3=true` is added to the query string of the public URL for an asset, then the asset will be streamed from S3 via the Asset Manager app. The default behaviour is unaffected.

Serving assets through the Rails app and not using `X-Sendfile` and nginx is naïve and may well not be practical for production use, but deploying these changes will allow us to check that we have all the appropriate infrastructure & credentials in place before we consider alternative options. This is explained in more detail in the relevant commit note.

The new behaviour requires the following standard environment variables to be set:

* `AWS_ACCESS_KEY_ID`
* `AWS_SECRET_ACCESS_KEY`
* `AWS_REGION`

And this application specific environment variable:

* `ASSET_MANAGER_S3_BUCKET_NAME`

Supersedes #63.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,6 +28,8 @@ gem 'rack_strip_client_ip', '0.0.1'
 
 gem 'mongoid_paranoia', '0.2.1'
 
+gem 'aws-sdk'
+
 # bundle exec rake doc:rails generates the API under doc/api.
 gem 'sdoc', '~> 0.4.0', group: :doc
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -41,6 +41,14 @@ GEM
       multi_json
     arel (6.0.3)
     ast (2.3.0)
+    aws-sdk (2.10.10)
+      aws-sdk-resources (= 2.10.10)
+    aws-sdk-core (2.10.10)
+      aws-sigv4 (~> 1.0)
+      jmespath (~> 1.0)
+    aws-sdk-resources (2.10.10)
+      aws-sdk-core (= 2.10.10)
+    aws-sigv4 (1.0.1)
     bson (3.2.6)
     builder (3.2.3)
     carrierwave (0.10.0)
@@ -87,6 +95,7 @@ GEM
       scss_lint
     hashie (3.4.3)
     i18n (0.7.0)
+    jmespath (1.3.1)
     json (1.8.3)
     jwt (1.5.2)
     kgio (2.10.0)
@@ -251,6 +260,7 @@ PLATFORMS
 
 DEPENDENCIES
   airbrake (~> 4.0.0)
+  aws-sdk
   carrierwave (~> 0.10.0)
   carrierwave-mongoid (~> 0.8.1)
   database_cleaner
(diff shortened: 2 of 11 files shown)


chrisroos:
I think we should remove the `ASSET_MANAGER` prefix from the `ASSET_MANAGER_S3_BUCKET_NAME` environment variable. I've recently discovered that this sort of namespacing is unnecessary. App-specific environment variables are written to app-specific directories by Puppet and are only loaded/read in the processes that run the app to avoid interfering with other apps on the same server.

floehopper:
@chrisroos, @chrislo: I've renamed the environment variable to `AWS_S3_BUCKET_NAME` to make it non-application-specific and using the `AWS` prefix to associate it with the other related env vars. I'm about to force-push the changes.

chrislo:
This looks good to me. I checkout out this branch to my local development environment to test it. Testing file upload without setting any environment variables gave the following output from `rake jobs:work`

``` example
[Worker(host:munro.local pid:57177)] Job Asset#scan_for_viruses (id=596f70632da79ade16000001) RUNNING
[Worker(host:munro.local pid:57177)] Job Asset#scan_for_viruses (id=596f70632da79ade16000001) COMPLETED after 8.6514
[Worker(host:munro.local pid:57177)] Job Asset#save_to_cloud_storage (id=596f706e2da79adf59000000) RUNNING
[Worker(host:munro.local pid:57177)] Job Asset#save_to_cloud_storage (id=596f706e2da79adf59000000) COMPLETED after 0.0012
[Worker(host:munro.local pid:57177)] 2 jobs processed at 0.2305 j/s, 0 failed
```

Testing file upload with the env variables set as per the explanation above gave

``` example
[Worker(host:munro.local pid:58096)] Job Asset#scan_for_viruses (id=596f715a2da79ae2e3000001) RUNNING
[Worker(host:munro.local pid:58096)] Job Asset#scan_for_viruses (id=596f715a2da79ae2e3000001) COMPLETED after 8.8415
[Worker(host:munro.local pid:58096)] Job Asset#save_to_cloud_storage (id=596f71632da79ae2f0000000) RUNNING
[Worker(host:munro.local pid:58096)] Job Asset#save_to_cloud_storage (id=596f71632da79ae2f0000000) COMPLETED after 0.5874
[Worker(host:munro.local pid:58096)] 2 jobs processed at 0.2113 j/s, 0 failed
```

I was able to view the uploaded asset at `/media/<id>/image.jpeg?stream_from_s3=true` and `/media/<id>/image.jpeg` and the response to the former had the `ETAG` header set as per fe04bbe.

The only thing I could think to change here would be to not run the delayed job at all if the environment variable wasn't set, rather than run a job which suceeds but does not have the expected side effect - this might make it a little easier to see at a glance what's going on if we have misconfigured something. Or perhaps we could log something informative instead? I think we'll be able to work it out either way though so :+1:

floehopper:
@chrislo: I think the AWS logging might help in that scenario, but I take your point.

Issue #75, opened 2017-07-19
Upgrade Ruby to 2.3.1


chrisroos:
Ruby 2.3.1 appears to be the latest version being used for GOV.UK apps.

We should update the [Ruby/Rails versions spreadsheet](https://docs.google.com/spreadsheets/d/1FJmr39c9eXgpA-qHUU6GAbbJrnenc0P7JcyY2NB9PgU/edit#gid=1480786499) when we're done.

PR #83, opened 2017-07-25
Upgrade Ruby 2.2.3 -> 2.3.1

chrislo:
Fixes: #75 

The tests continue to pass with this version of Ruby. I don't think it requires any other changes.


chrislo:
@floehopper thank you for checking that in such detail, much appreciated. I've got a PR incoming for a mongoid upgrade, agreed we should press on with upgrading others. 

`release_100` contains the ruby upgrade in this PR - `release_99` is on production so we can deploy the ruby change on its own tomorrow.

PR #84, opened 2017-07-25
Respond with error if asset requested from S3, but S3 not configured

floehopper:
If the AWS S3 bucket is not configured (i.e. the bucket name is not set) and a request is made to stream
an asset from S3, then the server now responds with a "500 Internal server error" instead of a "200 OK" and an empty file.

I've also done a bunch of tidying up in this pull request. The actual change is in the last commit, "Error if asset requested from S3, but S3 not configured". See that commit note for further details of the choices I made.

Note that I've had to disable a Rubocop rule for one line in the "Extract AWS_S3_BUCKET_NAME env var into app config" commit. See the commit note for details.


floehopper:

chrislo:
This looks good to me! 👍 

I'm not fond of the double negative in `disable_cloud_storage_stub: false` but it wasn't originally introduced in this PR, just used here, so I can address it separately.

floehopper:
@chrislo:

Thanks for reviewing.

> I'm not fond of the double negative in `disable_cloud_storage_stub: false` but it wasn't originally introduced in this PR, just used here, so I can address it separately. 

In fact it **was** introduced in this PR. I understand where you're coming from, but the problem I had was that I want the stubbing behaviour to be the default. Can you think of a better way of doing it?

I think I'm going to get this merged anyway and we can address this separately as you suggest.

PR #88, opened 2017-07-25
Upgrade Mongoid 4.0.2 -> 5.2.1

chrislo:
Upgrade Mongoid to the latest patch release in the 5 series using
`bundle update --conservative mongoid`.

I used `rails g mongoid:config` to update `config/mongoid.yml` and
modified the configuration to use the correct database connection
URI. I created a production config based on other mongoid-based
applications in the organisation, in particular following the pattern
we used for upgrading Mongoid from 4 to 5 in Manuals Publisher[1]

In Mongoid 5 documents are no longer ordered by `_id` by default[2]. I've
changed the specs that previously relied on the implicit ordering by
`_id` when calling the `last` method to make this ordering
explicit. There's probably a more intention-revealing way to write
these expectations, but doing it this way for now makes it clear why
the change has to be made - we can improve it later.

[1] https://github.com/alphagov/manuals-publisher/pull/1127
[2] https://docs.mongodb.com/mongoid/master/tutorials/mongoid-upgrade/


config/mongoid.yml
```diff
@@ -1,27 +1,159 @@
...
+
+    # Application name that is printed to the mongodb logs upon establishing a
+    # connection in server versions >= 3.4. Note that the name cannot exceed 128 bytes.
+    # app_name: MyApplicationName
 test:
-  sessions:
+  clients:
     default:
+      # Don't want this interfering with a concurrent publisher/content_api test run
+      # database: manuals_publisher_test
```
floehopper: I'm not sure these two comment lines are correct/relevant are they? Presumably they've been copied from Manuals Publisher...?

config/mongoid.yml
```diff
@@ -1,27 +1,159 @@
 development:
-  sessions:
+  # Configure available database clients. (required)
+  clients:
+    # Defines the default client. (required)
     default:
+      # Defines the name of the default database that Mongoid can connect to.
+      # (required).
+      # database: manuals_publisher_development
```
floehopper: I'm not sure this comment line is correct/relevant is it? Presumably they've been copied from Manuals Publisher...?

spec/models/asset_spec.rb
```diff
@@ -112,7 +112,7 @@
         asset.scan_for_viruses
       }.to change(Delayed::Job, :count).by(1)
 
-      job = Delayed::Job.last
+      job = Delayed::Job.asc('_id').last
```
floehopper: As you explain in the commit note, these `asc('_id')` calls are a bit ugly. I think it would be worth opening a new issue to capture the idea of fixing them otherwise I think they'll just get left.

chrislo:
Thanks @floehopper. I've decided to remove the comments altogether - they were partly generated by the config generator and partly copied incorrectly from Manuals Publisher. I think we can find these options in the documentation or code if we ever need to change them. 

Rather than raise an issue, I've addressed the `'_id'` ugliness in a9ede4d. Could you take another look?

config/mongoid.yml
```diff
@@ -1,27 +1,23 @@
 development:
-  sessions:
+  clients:
     default:
       uri: <%= ENV["MONGODB_URI"] || "mongodb://localhost/govuk_assets_development" %>
       options:
-        write:
-          w: 1
-        read: primary
+  options:
```
floehopper: Very minor: It seems slightly odd to include this key with no options.

config/mongoid.yml
```diff
@@ -1,27 +1,23 @@
...
       uri: <%= ENV['MONGODB_URI'] %>
       options:
-        write:
-          w: 1
-        read: primary
-        retry_interval: 120
+        read:
+          mode: :primary
+  options:
+    use_activesupport_time_zone: true
```
floehopper: Minor: It seems slightly odd that this option is only included in the `production` environment. Do you know why it's not included in `development` or `test`?

spec/support/delayed_job.rb
```diff
@@ -2,6 +2,22 @@ module DelayedJobHelpers
   def run_all_delayed_jobs
     Delayed::Worker.new(exit_on_complete: true, quiet: true).start
   end
+
+  def most_recently_enqueued_job
+    Delayed::Job.asc('_id').last
```
floehopper: Very minor & subjective: I really like the intention-revealing methods that have been extracted, but ideally I wonder whether we need to access the `_id` attribute at all given that it feels like an implementation detail of Mongoid/MongoDB. I wonder whether we could use `created_at` or `updated_at` instead.

config/mongoid.yml
chrislo: It turns out the default is `true` so it doesn't need to be explicitly set.

chrislo:
Rebasing and force pushing before merging.

PR #89, opened 2017-07-25
Enable "stream from S3" functionality via an environment variable

floehopper:
This allows us to make streaming from S3 the default behaviour in a given environment based on the presence of the new `STREAM_ALL_ASSETS_FROM_S3` environment variable.

I've also added some documentation about the whole "Assets on S3" functionality, mainly in order to document the various environment variables.

Note that this is based on #84 and includes a squash commit of those changes which will be removed before merging.


floehopper:

PR #90, opened 2017-07-25
Remove CSRF protection

### LINKED ISSUE #87: Investigate "Can't verify CSRF token authenticity" warning
Steps to reproduce with app running on developer VM:

```
$ echo `date` > tmp.txt

$ curl http://asset-manager.dev.gov.uk/assets --form "asset[file]=@tmp.txt"                        
{"_response_info":{"status":"created"},"id":"http://asset-manager.dev.gov.uk/assets/59775bc7759b74230c000000","name":"tmp.txt","content_type":"text/plain","file_url":"http://assets-origin.dev.gov.uk/media/59775bc7759b74230c000000/tmp.txt","state":"unscanned"}%  
```

This resulted in the following lines in the Rails `development.log` file:

```
Started POST "/assets" for ::1 at 2017-07-25 14:55:03 +0000
Processing by AssetsController#create as */*
  Parameters: {"asset"=>{"file"=>#<ActionDispatch::Http::UploadedFile:0x007f2c2a9c4a20 @tempfile=#<Tempfile:/tmp/user/1000/RackMultipart20170725-8972-vzax4h.txt>, @original_filename
="tmp.txt", @content_type="text/plain", @headers="Content-Disposition: form-data; name=\"asset[file]\"; filename=\"tmp.txt\"\r\nContent-Type: text/plain\r\n">}}
Can't verify CSRF token authenticity
```

I don't think we should be seeing the "Can't verify CSRF token authenticity" warning.

floehopper:
This is an internal API and doesn't allow direct user access.

c.f. alphagov/publishing-api@40ba7fc1d163fd966500f542009de30b31412f14

Fixes #87.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,10 +1,6 @@
 require 'cloud_storage'
 
 class ApplicationController < ActionController::Base
-  # Prevent CSRF attacks by raising an exception.
-  # For APIs, you may want to use :null_session instead.
-  protect_from_forgery with: :null_session
-
   include GDS::SSO::ControllerMethods
 
   before_filter :require_signin_permission!


floehopper:
We think the log warning might be appearing only when requesting an HTML response and not when requesting JSON. There is a separate issue about [the confusing behaviour when requesting JSON][1] which we should probably tackle before deciding whether it's sensible to remove the call to  `protect_from_forgery`. Closing.

PR #91, opened 2017-07-26
Upgrade govuk-lint from v2.1.0 to v3.1.0

floehopper:
And thus upgrade rubocop from v0.43.0 -> v0.49.1.

I've had to fix a number of rubocop violations which have been introduced by this change. See the individual commit notes for details.

It's worth noting that I've chosen to exclude all specs from the `Metrics/BlockLength` rubocop rule by adding a `.rubocop.yml` file. The default restriction of 25 lines per block doesn't play well with the
DSL style of RSpec as [noted here][1].

Initially I investigated adding [`rubocop-rspec`][2] to the `govuk-lint` gem, but this then introduced a whole load more violations. The change in this PR is a stop-gap solution. I plan to open an issue on the `govuk-lint` repo to highlight the problem.

[1]: https://stackoverflow.com/questions/40934345/rubocop-25-line-block-size-and-rspec-tests#answer-41187163
[2]: https://github.com/backus/rubocop-rspec


floehopper:

Rebasing against master, resolving conflicts and force-pushing in preparation for merging.

PR #109, opened 2017-08-01
Add Cache-Control header to S3 object

### LINKED ISSUE #96: Set Cache-Control HTTP header on S3 objects
We think we'll want/need to set cache control headers on the assets stored in S3 in order to reduce the number of requests made directly to S3.

floehopper:
We want to be able to redirect Asset Manager requests to the assets stored in an S3 bucket. However, we want as much of the externally-visible behaviour to remain the same and so we need to set the 'Cache-Control' header to the same value as in `MediaController#download`, i.e. the value generated by [`ActionController::ConditionalGet#expires_in`][1] using `ApplicationController#set_expiry`.

Closes #96.

[1]: http://api.rubyonrails.org/v4.2.7.1/classes/ActionController/ConditionalGet.html#method-i-expires_in

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -28,7 +28,7 @@ private
 
   def set_expiry(duration)
     unless Rails.env.development?
-      expires_in duration, public: true
+      expires_in duration, **AssetManager.cache_control.options
     end
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -27,6 +27,8 @@ private
   end
 
   def set_expiry(duration)
-    expires_in duration, public: true unless Rails.env.development?
+    unless Rails.env.development?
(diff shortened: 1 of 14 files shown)


floehopper:

floehopper:
@chrislo:

Thanks for reviewing.

> Perhaps a link to [this documentation](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html#RESTObjectPUT-requests) in 807768a might be useful context? 

Good idea. I'll add that and get this merged.

floehopper:
@chrislo & I have just confirmed that the Content-Disposition response header is being set correctly on S3 objects uploaded from integration. We did this by using the Rails console in integration to read the response headers via the AWS SDK. This gives us confidence that they will be being written correctly in staging and production, although we probably won't be able to check this until the S3 bucket is made publicly readable.

PR #111, opened 2017-08-02
Add Content-Disposition header to S3 object

### LINKED ISSUE #95: Set the Content-Disposition HTTP Header for objects on S3
We think we'll want/need to set this header to instruct the clients (e.g. browsers) to download the asset instead of displaying it inline.

floehopper:
We want to be able to redirect Asset Manager requests to the assets stored in an S3 bucket. However, we want as much of the externally-visible behaviour to remain the same and so we need to set the 'Content-Disposition' header to the same value as in `MediaController#download`, i.e. the value generated by [`ActionController::DataStreaming#send_data`][1] or [`ActionController::DataStreaming#send_file`][2].

Note that his is based on the work in [Add Cache-Control header to S3 object][3], so it's probably worth waiting until that is merged before reviewing this.

Closes #95.

[1]: http://api.rubyonrails.org/v4.2.7.1/classes/ActionController/DataStreaming.html#method-i-send_data
[2]: http://api.rubyonrails.org/v4.2.7.1/classes/ActionController/DataStreaming.html#method-i-send_file
[3]: https://github.com/alphagov/asset-manager/pull/109

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -18,9 +18,9 @@ class MediaController < ApplicationController
         set_expiry(AssetManager.cache_control.max_age)
         if stream_from_s3?
           body = Services.cloud_storage.load(asset)
-          send_data(body.read, filename: File.basename(asset.file.path), disposition: 'inline')
+          send_data(body.read, **AssetManager.content_disposition.options_for(asset))
         else
-          send_file(asset.file.path, disposition: 'inline')
+          send_file(asset.file.path, disposition: AssetManager.content_disposition.type)
         end
       end
     end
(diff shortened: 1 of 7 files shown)


floehopper:
@chrislo: Thanks for reviewing. And thanks for searching Rack for a way to avoid having to manually construct the `Content-Disposition` header. I couldn't find any easily re-usable code in Rails to do it either. I'll add a note to the relevant commit to explain all this.

floehopper:
Updated commit note, rebased against master and force-pushed in preparation for merging.

floehopper:
@chrislo & I have just confirmed that the Content-Disposition response header is being set correctly on S3 objects uploaded from integration. We did this by using the Rails console in integration to read the response headers via the AWS SDK. This gives us confidence that they will be being written correctly in staging and production, although we probably won't be able to check this until the S3 bucket is made publicly readable.

PR #112, opened 2017-08-03
Optionally redirect public asset requests to S3

### LINKED ISSUE #93: Add functionality to redirect to the file on S3
We're currently able to stream the asset from S3 but want to compare the performance of redirecting to the asset instead.

We might want to toggle this functionality by using a querystring and/or using an environment variable as we've done for the streaming functionality.

floehopper:
We're currently experimenting with different ways to serve assets and adding this optional behaviour makes it possible to have the Asset Manager *redirect* public requests for assets to S3 rather than serving them from the (NFS) filesystem or streaming them from S3 through the app.

The new behaviour is not enabled by default and can be enabled for a single request using a query string parameter or for all requests in an environment by setting an environment variable. See the changes in the `README` for details.

Eventually we're probably going to want to use a "virtual host" style URL for the assets on S3, but this will require us to setup a DNS CNAME for an appropriate sub-domain of gov.uk and setup a bucket with the bucket name matching the fully-qualified domain name. To make it possible to try out the new redirect behaviour before this is in place, I've made the `virtual_host` option for `Aws::S3::Object#public_url` configurable via an environment variable.

Note that I chose not to store the public URL in the database, because the call to `Aws::S3::Object#public_url` doesn't appear to connect to the remote API, i.e. it already has all the information it needs to construct the URL.

Fixes #93.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,10 +54,13 @@ The following environment variables are only needed if you want to enable this f
 
 * `AWS_S3_BUCKET_NAME` - name of bucket where assets are to be stored
 * `STREAM_ALL_ASSETS_FROM_S3` - causes *all* assets to be served from S3 via the app
+* `REDIRECT_ALL_ASSET_REQUESTS_TO_S3` - causes *all* requests for assets to be redirected to S3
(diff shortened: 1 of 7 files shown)


floehopper:

PR #117, opened 2017-08-09
Add an overview of the migration of assets to S3

chrisroos:
I'm hoping we can keep this document up to date and to have it serve as
a single resource we can point people to when talking about the
migration of assets to S3.


chrisroos:
@floehopper / @chrislo: Can we go through this together to determine whether it's going to be useful.

PR #120, opened 2017-08-09
[DO NOT MERGE] Serve assets from S3 if they exist

chrisroos:
__NOTE. We're not intending for this branch to be merged to master. We're going to use it to test the behaviour of redirecting to S3 in integration.__

---

This is a temporary change that will allow us to test redirecting to and
serving assets from S3 on integration.

This commit causes some tests to fail but we're not worrying about that
as we're not expecting this commit to end up in master.

We've tested the change locally and are happy that it results in a
redirect to assets uploaded to S3. Assets not on S3 continue to be
served directly from the filesystem.

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -31,7 +31,7 @@ class MediaController < ApplicationController
 protected
 
   def redirect_to_s3?
-    AssetManager.redirect_all_asset_requests_to_s3 || params[:redirect_to_s3].present?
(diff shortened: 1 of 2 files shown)


floehopper:
I'm going to close this because I don't think we need it any more and in any case I'll leave the remote branch in place in case we do need it again.

chrisroos:
Thanks @floehopper. I'm going to delete the branch as we can always recreate the changes based on the diff in this PR if we need to.

PR #123, opened 2017-08-09
Add info about destroy/restore endpoints to README

chrisroos:
This was missing from the API docs and examples.

Note that these methods have already been [documented in the gds-api-adapters Gem][1].

[1]:
https://github.com/alphagov/gds-api-adapters/blob/6984336dc73d07a44d409ba138ae7e10d83148d2/lib/gds_api/asset_manager.rb#L94-L138


chrisroos:
Thanks @floehopper. I've just rebased on master and force pushed. Can you approve it again, please?

PR #126, opened 2017-08-14
Optionally instruct nginx to proxy assets from S3

chrisroos:
This complements [the change in the asset-manager nginx config][1].

The app sets the `X-Accel-Redirect` header to instruct nginx to proxy
the request for an asset to S3.

We're generating a presigned URL for the assets on S3 so that we can
avoid making them publicly accessible. The presigned URL is valid for 1
minute which should be enough time for nginx to use the URL to make the
request.

[1]: https://github.com/alphagov/govuk-puppet/pull/6280

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -21,6 +21,10 @@ class MediaController < ApplicationController
         elsif stream_from_s3?
           body = Services.cloud_storage.load(asset)
           send_data(body.read, **AssetManager.content_disposition.options_for(asset))
+        elsif proxy_via_nginx?
+          url = Services.cloud_storage.presigned_url_for(asset)
+          headers['X-Accel-Redirect'] = "/cloud-storage-proxy/#{url}"
+          render nothing: true
         else
           send_file(asset.file.path, disposition: AssetManager.content_disposition.type)
         end
@@ -29,6 +33,10 @@ class MediaController < ApplicationController
   end
 
 protected
+
+  def proxy_via_nginx?
+    params[:proxy_via_nginx].present?
+  end
 
   def redirect_to_s3?
     AssetManager.redirect_all_asset_requests_to_s3 || params[:redirect_to_s3].present?
--- a/lib/s3_storage.rb
+++ b/lib/s3_storage.rb
@@ -31,6 +31,10 @@ class S3Storage
     object_for(asset).public_url(virtual_host: AssetManager.aws_s3_use_virtual_host)
   end
 
+  def presigned_url_for(asset)
+    object_for(asset).presigned_url(:get, expires_in: 1.minute, virtual_host: AssetManager.aws_s3_use_virtual_host)
+  end
+
 private
 
   def object_for(asset)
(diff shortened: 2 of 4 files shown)


app/controllers/media_controller.rb
```diff
@@ -21,6 +21,12 @@ def download
         elsif stream_from_s3?
           body = Services.cloud_storage.load(asset)
           send_data(body.read, **AssetManager.content_disposition.options_for(asset))
+        elsif proxy_via_nginx?
+          url = Services.cloud_storage.presigned_url_for(asset)
+          headers['X-Cloud-Storage-URL'] = url
+          headers['X-Cloud-Storage-Host'] = URI.parse(url).host
```
floehopper: It would be useful to explain in the commit note why you need *both* the full URL (including the host) *and* the host on its own.

lib/s3_storage.rb
```diff
@@ -31,6 +31,10 @@ def public_url_for(asset)
     object_for(asset).public_url(virtual_host: AssetManager.aws_s3_use_virtual_host)
   end
 
+  def presigned_url_for(asset)
+    object_for(asset).presigned_url(:get, expires_in: 1.minute, virtual_host: AssetManager.aws_s3_use_virtual_host)
```
floehopper: Minor: It would be nice to pull the `expires_in` option value into the application configuration c.f. `AssetManager.aws_s3_use_virtual_host`.

app/controllers/media_controller.rb
floehopper: I've just noticed that you've explained this in an example name below, so I think it's less important to explain in the commit note.

app/controllers/media_controller.rb
```diff
@@ -21,6 +21,12 @@ def download
...
+          headers['X-Accel-Redirect'] = "/cloud-storage-proxy/"
```
floehopper: Very minor: It would be nice to be consistent about using single quotes where possible in this file.

app/controllers/media_controller.rb
chrisroos: We're no longer sending the URL and Host so this problem has gone away.

app/controllers/media_controller.rb
chrisroos: The value of the `X-Accel-Redirect` header now includes some string interpolation so I've had to keep the double quotes.

chrisroos:
Thanks for the review, @floehopper.

I've just rebased on master and made some changes based on the updates in https://github.com/alphagov/govuk-puppet/pull/6280.

Can you review it again, please?

spec/lib/s3_storage_spec.rb
```diff
@@ -84,4 +84,28 @@
...
+
+    context 'when configured to use virtual host' do
+      let(:use_virtual_host) { true }
+
+      it 'returns presigned URL for asset on S3 using virtual host' do
+        allow(s3_object).to receive(:presigned_url).with(:get, expires_in: 1.minute, virtual_host: true).and_return('presigned-url')
+        expect(subject.presigned_url_for(asset)).to eq('presigned-url')
+      end
+    end
+  end
```
floehopper: Minor: It would be nice to test the value of the `expires_in` option separately to (a) reduce the brittleness/duplication in these two examples; and (b) explain the choice of value in an example name.

PR #128, opened 2017-08-14
Add example to exercise S3Storage::Null#save

floehopper:
This should avoid problems like the one fixed in #124 happening again.


floehopper:
@chrislo, @chrisroos: I'd be grateful if you could review this when you have a mo. Thanks.

floehopper:

PR #131, opened 2017-08-15
Upgrade airbrake gem v4.0.0 -> v4.3.8

### LINKED ISSUE #122: Upgrade Airbrake Gem
We're using version 4.0.0 of the Airbrake Gem. This version doesn't report on `ActionController::BadRequest` exceptions correctly. They're reported in Errbit as `Notification`s, which isn't very helpful.

[Pull request 271 in the Airbrake project](https://github.com/airbrake/airbrake/pull/271) appears to have fixed this problem.

The pull request was updated on Aug 26 2014 with a comment saying that the fix would be released the following week. Airbrake version 4.1.0 was released on Sep 4 2014 so I suspect we'll want at least this version.

Note that I've copied the text of this description from https://github.com/alphagov/smart-answers/issues/1681 where we encountered the same problem.

floehopper:
I changed the constraint in the `Gemfile` from `~> 4.0.0` to `~> 4.0` and ran the following command:

    bundle update --conservative airbrake

My main motivation for doing this is so that `airbrake` [reports `ActionController::BadRequest` exceptions correctly][1].

However, I've chosen to upgrade to the latest 4.x version, because that should include other bug fixes without breaking compatibility with `errbit`. I'm pretty sure this is safe to do, because I can see that a bunch of other GOV.UK apps are using this version of `airbrake` in production, e.g.

* `calculators`
* `hmrc-manuals-api`
* `manuals-publisher`
* `publisher`
* `rummager`

Closes #122.

[1]: https://github.com/airbrake/airbrake/pull/271

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'airbrake', '~> 4.0.0'
+gem 'airbrake', '~> 4.0'
 gem 'aws-sdk'
 gem 'carrierwave', '~> 0.10.0'
 gem 'carrierwave-mongoid', '~> 0.8.1', require: 'carrierwave/mongoid'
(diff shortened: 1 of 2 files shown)


floehopper:

PR #133, opened 2017-08-16
Always virus scan uploaded files

### LINKED ISSUE #72: Updating assets can bypass virus scanning
Updating an existing asset by uploading a different file but with the same name causes the asset-manager to bypass virus scanning.

Steps to reproduce:

1. Create temporary file

```
$ echo `date` > tmp.txt
$ cat tmp.txt
Tue 18 Jul 2017 17:22:54 BST
```

2. Upload file

```
$ curl http://localhost:3000/assets --form "asset[file]=@tmp.txt"
{"_response_info":{"status":"created"},"id":"http://localhost:3000/assets/596e35e4edf877ece4000003","name":"tmp.txt","content_type":"text/plain","file_url":"http://static.dev.gov.uk/media/596e35e4edf877ece4000003/tmp.txt","state":"unscanned"}
```

3. Request file info

Ensure virus scanning is running (`bundle exec rake jobs:work`) and that the response says the file is clean.

```
$ curl http://localhost:3000/assets/596e35e4edf877ece4000003
{"_response_info":{"status":"ok"},"id":"http://localhost:3000/assets/596e35e4edf877ece4000003","name":"tmp.txt","content_type":"text/plain","file_url":"http://static.dev.gov.uk/media/596e35e4edf877ece4000003/tmp.txt","state":"clean"}
```

4. Request the file

Observe that the content matches the content of our temporary file created in step 1.

```
$ curl http://localhost:3000/media/596e35e4edf877ece4000003/tmp.txt
Tue 18 Jul 2017 17:22:54 BST
```

5. Update the temporary file

```
$ echo `date` > tmp.txt
$ cat tmp.txt
Tue 18 Jul 2017 17:25:09 BST
```

6. Stop the virus scanning job

7. Update the asset

NOTE. The file is already marked as clean in the response.

```
$ curl http://localhost:3000/assets/596e35e4edf877ece4000003 --request PUT --form "asset[file]=@tmp.txt"
{"_response_info":{"status":"success"},"id":"http://localhost:3000/assets/596e35e4edf877ece4000003","name":"tmp.txt","content_type":"text/plain","file_url":"http://static.dev.gov.uk/media/596e35e4edf877ece4000003/tmp.txt","state":"clean"}
```

8. Request the file

Observe that the content matches the content of our updated temporary file that we uploaded in step 7.

```
$ curl http://localhost:3000/media/596e35e4edf877ece4000003/tmp.txt
Tue 18 Jul 2017 17:25:09 BST
```

floehopper:
Previously a virus scan was only triggered if the file had "changed" according to [`ActiveModel::Dirty`][1]. However, this seemed to have been based on whether the filename of the file had changed rather than whether the content had changed which seems to be the more relevant issue. This meant that (as described in #72) it was possible to update an existing asset with a file having the same name, but different content and for the asset to remain in the "clean" state. This seemed like a security hole.

Initially I investigated triggering a virus scan when the file *content* changed. However, this proved awkward due to the way both `state_machines` and `carrierwave` gems work. So after discussion with @chrisroos & @chrislo, I decided to trigger a virus scan whenever the `Asset#file` attribute is changed. While this means virus scans will sometimes be triggered unnecessarily, I don't think it should happen too often and it seems better to do this than allow unscanned files to be served to the citizens.

I'm not sure this is the most elegant solution, but since the `Asset#file=` method already existed, it seemed simplest to make use of it. A better solution might involve persisting an MD5 digest of the file content which has been virus scanned and then only re-scan the content if it has changed. However, the extra complexity doesn't seem warranted.

Fixes #72.

[1]: http://api.rubyonrails.org/v4.2.7.1/classes/ActiveModel/Dirty.html

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -19,7 +19,6 @@ class Asset
 
   mount_uploader :file, AssetUploader
 
-  before_save :reset_state_if_file_changed
   after_save :schedule_virus_scan
 
   state_machine :state, initial: :unscanned do
@@ -41,6 +40,7 @@ class Asset
     super(file).tap {
       filename_history.push(old_filename) if old_filename
     }
+    reset_state
   end
 
   def filename_valid?(filename_to_test)
@@ -90,8 +90,8 @@ protected
     filename_history + [filename]
   end
 
-  def reset_state_if_file_changed
-    self.state = 'unscanned' if self.file_changed?
+  def reset_state
+    self.state = 'unscanned'
   end
 
   def schedule_virus_scan
--- a/spec/models/asset_spec.rb
+++ b/spec/models/asset_spec.rb
@@ -92,6 +92,18 @@ RSpec.describe Asset, type: :model do
       expect(most_recently_enqueued_job).to have_method_name(:scan_for_viruses)
     end
 
+    it "should schedule a scan after save if the file is changed even if filename is unchanged" do
+      a = FactoryGirl.create(:clean_asset)
+      original_filename = a.file.send(:original_filename)
+      a.file = load_fixture_file("lorem.txt", named: original_filename)
+      expect {
+        a.save!
+      }.to change(Delayed::Job, :count).by(1)
+
+      expect(most_recently_enqueued_job).to have_payload(a)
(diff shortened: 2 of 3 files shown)


floehopper:

Merging.

PR #134, opened 2017-08-16
Set Content-Type when uploading asset to S3

### LINKED ISSUE #130: Set a content type header when uploading assets to S3
The current `send_file` method of serving assets from `MediaController#download` infers the `Content-Type` header [from the extension of the uploaded filename](http://api.rubyonrails.org/v4.2/classes/ActionController/DataStreaming.html#method-i-send_file). We will need to set the content type explicitly on the S3 Object metadata when serving assets from S3 (either by proxy or redirect).

floehopper:
We're currently thinking about either redirecting Asset Manager requests to S3 or proxying them to S3. In either case, we want to include the same response headers to keep the externally visible behaviour unchanged.

We've already added Cache-Control and Content-Disposition headers in previous pull requests (#109 & #111) and this commit adds Content-Type based on the [same logic that `ActionController::DataStreaming#send_file` uses][1].

I've set `AssetManager.default_content_type` to [the default value used by Rails][2] to ensure that the existing `send_file` behaviour and any new behaviour which relies on assets being downloaded from S3 are kept in sync.

Closes #130.

[1]: https://github.com/rails/rails/blob/v4.2.7.1/actionpack/lib/action_controller/metal/data_streaming.rb#L146
[2]: https://github.com/rails/rails/blob/v4.2.7.1/actionpack/lib/action_controller/metal/data_streaming.rb#L11

### DIFF
--- a/spec/models/asset_spec.rb
+++ b/spec/models/asset_spec.rb
@@ -164,6 +164,12 @@ RSpec.describe Asset, type: :model do
         asset.save_to_cloud_storage
       end
 
+      it 'sets the Content-Type header on the asset stored in the cloud' do
+        expect(cloud_storage).to receive(:save).with(anything, include(content_type: 'image/png'))
+
+        asset.save_to_cloud_storage
+      end
+
       context 'when an exception is raised' do
         let(:exception_class) { Class.new(StandardError) }
         let(:exception) { exception_class.new }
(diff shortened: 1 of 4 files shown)


spec/models/asset_spec.rb
```diff
@@ -304,4 +310,48 @@
...
+      it "returns content type based on asset file extension" do
+        expect(asset.content_type).to eq(Mime::Type.lookup('image/png').to_s)
+      end
+    end
+
+    context "when asset file has no extension" do
+      let(:asset) { Asset.new(file: load_fixture_file("asset-without-extension")) }
+
+      it "returns default content type" do
+        expect(asset.content_type).to eq(AssetManager.default_content_type)
```
chrislo: I wonder if it might be worth asserting the literal value of `ActionController::DataStreaming::DEFAULT_SEND_FILE_TYPE` here, just in case that constant changes to `nil` or `""` when we upgrade Rails.

PR #141, opened 2017-08-21
Update migrating-assets-to-s3 doc

chrisroos:
So that it reflects our current approach.


chrisroos:
Thanks @floehopper. I'm going to get this merged.

I agree that keeping it in the repo doesn't feel quite right but I'm not sure there's anywhere else that's obviously better.

Issue #151, opened 2017-08-23
Allow diversion of a percentage of traffic to the proxying to S3 solution


floehopper:
We want to be able to set an environment variable to determine the percentage of traffic to use the proxying to S3 solution. The remaining traffic will be served via sendfile as before. This will allow us to gradually rollout the new solution while monitoring its effect on production load.

Note that the environment variable will need to be added to `govuk-puppet`. Ideally we'd like to be able to hard-code the value in different environments within that repo as per #136.

## Todo

- [x] Update asset-manager app
- [x] Update govuk-puppet
- [ ] Test the change on integration

floehopper:
I'm un-assigning myself from this now that I've merged #152. I think the plan is that @chrisroos will pick up the `govuk-puppet` aspect of this issue.

chrisroos:
I've opened a PR to add the environment variable to govuk-puppet - https://github.com/alphagov/govuk-puppet/pull/6351.

chrisroos:
I've now merged the changes in https://github.com/alphagov/govuk-puppet/pull/6351. This should allow us to set the `PROXY_PERCENTAGE_OF_ASSET_REQUESTS_TO_S3_VIA_NGINX` environment variable to control the percentage of requests we respond to by serving assets hosted on S3.

chrisroos:
I'm going to wait until the govuk-puppet change has been deployed to integration so that I can test this behaviour before closing the issue.

chrisroos:
I've just tested this in integration and am confident that it's working as expected.

I manually modified the value in `/etc/govuk/asset-manager/env.d/PROXY_PERCENTAGE_OF_ASSET_REQUESTS_TO_S3_VIA_NGINX` and used the `deploy:with_hard_restart` Jenkins job to force a reload of asset-manager app. I uploaded a file to asset-manager and used `curl` to request that file from the Rails app directly (i.e. bypassing nginx).

With `PROXY_PERCENTAGE_OF_ASSET_REQUESTS_TO_S3_VIA_NGINX` set to 100, I saw that all responses contained the `X-Accel-Redirect` header instructing nginx to proxy the request to S3. 

With `PROXY_PERCENTAGE_OF_ASSET_REQUESTS_TO_S3_VIA_NGINX` set to 50, I saw that about 50% of the response included the `X-Accel-Redirect` header.

Issue #92, opened 2017-07-28
Use rubocop-rspec


floehopper:
Probably [via govuk-lint](https://github.com/alphagov/govuk-lint/issues/80), but possibly directly.

This would be a better fix for the `Metrics/BlockLength` rubocop problem than [this commit](https://github.com/alphagov/asset-manager/commit/9cb3ba39e9613ac3a0302a8a760275985b99bae2).

floehopper:
I've added `rubocop-rspec` as a dependency to `govuk-lint` in alphagov/govuk-lint#83. However, note that doing so doesn't fix the `Metrics/BlockLength` problem as I had hoped. However, @h-lame has opened alphagov/govuk-lint#82 with an alternative fix for this which we should adopt.

As part of this investigation, I've done some work locally on auto-correcting violations in the `rubocop-rspec` cops. I'll open those changes as a separate PR once a new version of `govuk-lint` is available.

chrisroos:
The `rubocop-rspec` gem was added in PR #192. @floehopper: Can this issue be closed?

floehopper:
@chrisroos:

> Can this issue be closed?

Yes, I think so.

PR #152, opened 2017-08-23
Allow a percentage of traffic to be proxied to S3 via Nginx

floehopper:
The motivation behind this commit is to allow us to incrementally roll-out the use of the proxying to S3 via Nginx mechanism. Although we've done a bunch of performance testing and analysis of the production logs, it's difficult to be certain that the production stack will be able to handle all its usual traffic when this mechanism is enabled. By only enabling it gradually we hope to observe any problems before they become catastrophic.

The `PROXY_ALL_ASSET_REQUESTS_TO_S3_VIA_NGINX` environment variable has been renamed to `PROXY_PERCENTAGE_OF_ASSET_REQUESTS_TO_S3_VIA_NGINX` and this should be set to an integer value between 0 and 100. We will need to set appropriate values for each environment via `govuk-puppet`.

This addresses the Asset Manager aspect of #151, but we will need to set values for this new environment variable via `govuk-puppet`.

I ended up doing quite a bit of tidying of the code prior to making the actual change; the latter is the final commit. Other changes include:

* Making other environment variable names consistent (I don't think we've used them anywhere yet, so this should be safe)
* Making code & specs more consistent.
* Improving test coverage and reducing combinatorial explosion in specs.
* Improve documentation about feature flags.

I think there's a case to be made to extract the response strategy selection logic and the response strategies themselves into their own classes - this would make the `MediaController#download` action easier to understand and easier to test. I did make a start on doing this, but decided it was a step too far for this PR. It may also be moot if we remove some of the response strategies in the near future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,13 +56,14 @@ The following environment variables are only needed if you want to enable this f
 
 * `AWS_S3_BUCKET_NAME` - name of bucket where assets are to be stored
 * `PROXY_ALL_ASSET_REQUESTS_TO_S3_VIA_RAILS` - causes *all* asset requests to be proxied to S3 via the Rails app
+* `PROXY_ALL_ASSET_REQUESTS_TO_S3_VIA_NGINX` - causes *all* asset requests to be proxied to S3 via Nginx
 * `REDIRECT_ALL_ASSET_REQUESTS_TO_S3` - causes *all* asset requests to be redirected to S3
 * `AWS_S3_USE_VIRTUAL_HOST` - generate URLs for virtual host (requires CNAME setup for bucket)
 
 #### Request parameters
 
 * Asset requests can be proxied to S3 via the Rails app even if `PROXY_ALL_ASSET_REQUESTS_TO_S3_VIA_RAILS` is not set by adding `proxy_to_s3_via_rails=true` as a request parameter key-value pair to the query string.
-* Asset requests can be proxied to S3 via Nginx by adding `proxy_to_s3_via_nginx=true` as a request parameter key-value pair to the query string.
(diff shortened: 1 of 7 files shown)


floehopper:
@chrisroos: If you're going to do the `govuk-puppet` work related to this, it might make sense for you to review it...

chrisroos:
The body of the commit titled "Improve consistency & naming in MediaController spec" contains "* Avoid use of "should" in example names". Is there a styleguide or similar that explains we shouldn't be using `should` in test names?

chrisroos:
The first sentence of the commit titled, "Make order of asset response strategies more logical", reads "I've changed the order of the conditions in the `if/elsif/else` statement to a more logical.". Is there a word missing?

floehopper:
@chrisroos:

Thanks for reviewing.

> The body of the commit titled "Improve consistency & naming in MediaController spec" contains "* Avoid use of "should" in example names". Is there a styleguide or similar that explains we shouldn't be using should in test names?

There is in the [`RSpec/ExampleWording` cop][1] in the `rubocop-rspec` gem which I think we should be using (see #92). The cop documentation refers to [this page][2].

> The first sentence of the commit titled, "Make order of asset response strategies more logical", reads "I've changed the order of the conditions in the `if/elsif/else` statement to a more logical.". Is there a word missing?

Yes. I'll fix it. It should say something more like: "I've changed the conditions in the `if/elsif/else` statement to be in a more logical order."

[1]: https://github.com/backus/rubocop-rspec/blob/79818d3c184552e039f6b66da10c2e641eb00cfd/lib/rubocop/cop/rspec/example_wording.rb#L6-L22
[2]: http://www.betterspecs.org/#should

floehopper:
Rebased against `master`, fixed commit wording, and force-pushed in preparation for merging.

Issue #148, opened 2017-08-23
Ensure Last-Modified from nginx+s3proxy matches Last-Modified from nginx+Sendfile


chrisroos:
This is similar to issue #144.

We're imaging an incremental rollout of the nginx+s3proxy approach to serving assets. The same asset served from the two approaches currently have different values for the Last-Modified response header. I presume nginx+sendfile uses the modification date of the file on disk while nginx+s3proxy uses the modification date of the object on S3.

The same asset having different Last-Modified values (e.g. last-modified-1 and last-modified-2) may result in some additional load on the asset-manager servers because a conditional GET using last-modified-1 might end up with a 200 response and last-modified-2, instead of a 304 response for last-modified-1.

floehopper:
I think @chrisroos effectively tested this in his testing of #164. For example, you can see that the `Last-Modified` header is the same whether Nginx serves the file from NFS or proxies it to S3 in the `curl` output in [this comment](https://github.com/alphagov/asset-manager/pull/164#issuecomment-325951294).

Closing.

Issue #144, opened 2017-08-23
Ensure Etag from nginx+s3proxy matches Etag from nginx+Sendfile


chrisroos:
We're imaging an incremental rollout of the nginx+s3proxy approach to serving assets. The same asset served from the two approaches currently has a different Etag. The Etag generated by nginx appears to be ([according to this Serverfault answer][serverfault-nginx-etag]) a combination of last modified date and content length. The Etag generated by S3 [may or may not be an md5 digest of the content depending on how the object was created and how it's encrypted][s3-etag].

The same asset having different Etags (e.g. etag-1 and etag-2) will result in some additional load on the asset-manager servers because a conditional GET using etag-1 might end up with a 200 response and etag-2, instead of a 304 response for etag-1.

I can think of two options (there may be others):

1. Generate the same Etags that nginx currently generates and store them against the objects in S3

    * This assumes that we can set an Etag on S3 and have it sent in the response to the client.
    * The advantage is that Etags remain the same and so caches can be informed that they can continue using a previously cached version of the file
    * The problem might be in generating the Etag value the same as nginx

2. Change the Etags for the current Sendfile mechanism and then use those same Etags when serving files from S3

    * This assumes we can customise the Etag sent from nginx (maybe by setting it in the Rails app) and that we can customise the Etag sent from S3
    * The disadvantage is that all cached content will/might have to be re-populated because the Etags have changed. If we're going to do this then we should roll it out in nginx and give it at least 24 hours to ensure all caches should have picked up the new Etags.
    * The advantage is that we're not tied to the Etag implementation of nginx.

[s3-etag]: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html
[serverfault-nginx-etag]: https://serverfault.com/questions/690341/algorithm-behind-nginx-etag-generation

chrisroos:
There is a similar issue for making sure the Last-Modified values are the same in #148.

floehopper:
As far as we can see in the [AWS SDK documentation](http://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Object.html) there's no way to *set* the value of the ETag or Last-Modified response headers for an S3 Object. We think this rules out option 1.

floehopper:
It might be possible for the Rails app to generate an ETag and somehow instruct Nginx to replace the ETag response header which comes back from S3, but it's not at all obvious how we would do that.

chrisroos:
[...]

PR #157, opened 2017-08-25
Send ETag & Last-Modified response headers when proxying to S3 via Nginx

floehopper:
We're planning to gradually rollout the implementation whereby the Rails app instructs Nginx to proxy asset requests to S3. The current default implementation is for the Rails app to instruct Nginx to serve the asset files from NFS. In this implementation Nginx is automatically adding an ETag and a Last-Modified header.

When we switch to proxying the asset requests to S3, we want to ensure these response headers are the same for a given asset file in order to avoid unnecessary cache invalidation and therefore extra load which will make it hard to discriminate between changes in performance due to the change in implementation versus changes in performance due to a reduction in caching.

This commit changes the Rails app so that it calculates the ETag and Last-Modified values based on the underlying file, in the same way that Nginx would, and sends them in the response if we are proxying to S3. We plan to make changes to the Nginx configuration to store these headers and add them to the proxied response from S3. Without those changes, these response headers are ignored by Nginx and so I think it's safe to make this change in isolation.

The ETag value is calculated based on the string format in the [`ngx_http_set_etag` function][1], `"%xT-%xO"` in combination with the file's last modified time and its size. It turns out that the `"T"` and the `"O"` in that string format are Nginx-specific and are defined in the [`ngx_vslprintf` function][2]. As far as I can see they are just type modifiers [for `time_t`][3] and [for `off_t`][4] respectively and I don't think they are relevant to the Ruby string format, `"%x-%x"`.

According to RFC2616, the [ETag value should be quoted][5] and the [Last-Modified value should be formatted as an "HTTP date"][6].

This change means that when we're proxying to S3, the Rails app will have to access the underlying file on the NFS mount which might slow things down. I've tried to minimize this by memo-izing the call to `File.stat` which is used to obtain the data needed to calculate the two header values. This means that the app should only make a single call to `File.stat` per request.

Note that this solution relies on the fact that the Rails app has access to the NFS mount. Since the purpose of the change in overall implementation is to stop using NFS, we may want to store these header values in the database, but that can be a separate change.

See [this Gist][7] for a demonstration of how this can work in conjunction with changes in the Nginx configuration.

Addresses the Asset Manager aspect of #144 and #148.

[1]: http://lxr.nginx.org/source/src/http/ngx_http_core_module.c?v=nginx-1.4.6#1801
[2]: http://lxr.nginx.org/source/src/core/ngx_string.c?v=nginx-1.4.6#0147
[3]: http://lxr.nginx.org/source/src/core/ngx_string.c?v=1.4.6#0279
[4]: http://lxr.nginx.org/source/src/core/ngx_string.c?v=1.4.6#0269
[5]: https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.19
[6]: https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.29
[7]: https://gist.github.com/floehopper/4b52c39cff18e108962c793c579dee1a

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -21,6 +21,8 @@ class MediaController < ApplicationController
         elsif proxy_to_s3_via_nginx?
           url = Services.cloud_storage.presigned_url_for(asset)
           headers['X-Accel-Redirect'] = "/cloud-storage-proxy/#{url}"
+          headers['ETag'] = %{"#{asset.etag}"}
+          headers['Last-Modified'] = asset.last_modified.httpdate
           render nothing: true
         elsif proxy_to_s3_via_rails?
           body = Services.cloud_storage.load(asset)
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -62,6 +62,14 @@ class Asset
     end
   end
 
+  def etag
+    '%x-%x' % [last_modified, file_stat.size]
+  end
+
+  def last_modified
+    file_stat.mtime
+  end
+
   def scan_for_viruses
     scanner = VirusScanner.new(self.file.current_path)
     if scanner.clean?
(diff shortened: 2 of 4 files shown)


floehopper:

I'll update the commit message as you suggest and get the changes merged.

Issue #149, opened 2017-08-23
Avoid duplicate Cache-Control headers when serving from nginx+s3proxy


chrisroos:
Requests for files hosted on S3 currently respond with a duplicate `Cache-Control: max-age=86400, public` header. We think one of these headers will be coming from Rails and the other from the object on S3. Although I don't think this'll cause any problems (because they're identical) we should only send a single `Cache-Control` header to avoid any potential problem in future.

floehopper:
We should probably test the combination of #164 and alphagov/govuk-puppet#6362 in integration before closing this.

chrisroos:
I've just tested this in integration. I uploaded a file and then used `curl` to request it from the filesystem and from S3. I've confirmed that I only see a single `Cache-Control` header in both cases.

```
# Requesting from filesystem
$ curl "https://assets-origin.integration.publishing.service.gov.uk/media/59a6eaa2e5274a62e055d262/1MB.dat" -v > /dev/null
< HTTP/1.1 200 OK
< Accept-Ranges: bytes
< Access-Control-Allow-Headers: origin, authorization
< Access-Control-Allow-Methods: GET, OPTIONS
< Access-Control-Allow-Origin: *
< Cache-Control: max-age=86400, public
< Connection: close
< Content-Disposition: inline; filename="1MB.dat"
< Content-Length: 1048576
< Content-Type: application/octet-stream
< Date: Wed, 30 Aug 2017 16:41:47 GMT
< ETag: "59a6eaa2-100000"
< Last-Modified: Wed, 30 Aug 2017 16:41:06 GMT
< Server: nginx
< Strict-Transport-Security: max-age=31536000
< X-Frame-Options: DENY

# Requesting from S3
$ curl "https://assets-origin.integration.publishing.service.gov.uk/media/59a6eaa2e5274a62e055d262/1MB.dat?proxy_to_s3_via_nginx=true" -v > /dev/null
< HTTP/1.1 200 OK
< Accept-Ranges: bytes
< Access-Control-Allow-Headers: origin, authorization
< Access-Control-Allow-Methods: GET, OPTIONS
< Access-Control-Allow-Origin: *
< Cache-Control: max-age=86400, public
< Connection: close
< Content-Disposition: inline; filename="1MB.dat"
< Content-Length: 1048576
< Content-Type: */*; charset=utf-8
< Date: Wed, 30 Aug 2017 16:42:03 GMT
< ETag: "59a6eaa2-100000"
[...]

PR #158, opened 2017-08-29
Avoid duplicate Cache-Control headers when proxying to S3 via Nginx

floehopper:
We are setting a `Cache-Control` response header on the S3 object and that's the one we want the client to receive. However prior to this commit, the Rails app was also sending a (duplicate) `Cache-Control` response header which Nginx was including in the response. Thus the user was receiving two duplicate `Cache-Control` headers. Although the values would almost certainly have been the same, we want to avoid confusion and/or unexpected behaviour.

This commit aims to achieve that by stopping the Rails app from sending a `Cache-Control` response header. Unfortunately that turns out to be harder than one might expect and necessitates the following changes:

* Avoid explicitly calling `ActionController::ConditionalGet::ClassMethods#expires_in` via `ApplicationController#set_expiry`. I think there's an argument for not calling this in some of the other cases, but I'm ignoring that for now.

* Avoid setting `ETag` or `Last-Modified` response headers. Setting either of these means that [`ActionDispatch::Http::Cache#handle_conditional_get!`][1] sets a default `Cache-Control` header value of ["max-age=0, private, must-revalidate"][2]. We avoid setting `ETag` or `Last-Modified` headers by renaming them to `X-Accel-ETag` and `X-Accel-Last-Modified`. This makes sense to me because the response from the Rails app is just an instruction to Nginx to proxy the request to S3; Nginx can then make use of these headers as it sees fit. Note that this will require a corresponding change to the Nginx configuration.

* Remove default `Cache-Control` response header set by `Rack::ETag`. [Rails explicitly sets a default `Cache-Control` header value of "no-cache" on this middleware][3] which [is added if no `ETag` header is set][4]. By inserting a new custom middleware, we can remove this `Cache-Control` header if the response is instructing Nginx to proxy the request to S3. We can detect this scenario via the presence of the `X-Accel-ETag` header in the response.

I considered the following alternative solutions:

1. Change the Nginx configuration to ignore the `Cache-Control` header from Rails.
2. Change the Nginx configuration to ignore the `Cache-Control` header from S3.
3. Remove the `Cache-Control` header from the S3 object.

While all of these would've avoided the client receiving duplicate `Cache-Control` headers, I couldn't find a way to implement option 1 and options 2 & 3 feel like a step backwards on the journey to redirecting asset requests to S3. Even if there was a way to implement option 1, it feels better to me that the Rails app doesn't send the `Cache-Control` header in the first place rather than having Nginx ignore it.

Addresses the Asset Manager aspect of #149. As noted above, we will need to make some corresponding changes to `govuk-puppet`.

[1]:
https://github.com/rails/rails/blob/v4.2.7.1/actionpack/lib/action_dispatch/http/cache.rb#L129-L133
[2]:
https://github.com/rails/rails/blob/v4.2.7.1/actionpack/lib/action_dispatch/http/cache.rb#L135
[3]:
https://github.com/rails/rails/blob/v4.2.7.1/railties/lib/rails/application/default_middleware_stack.rb#L58
[4]: https://github.com/rack/rack/blob/1.6.4/lib/rack/etag.rb#L39

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -15,19 +15,21 @@ class MediaController < ApplicationController
 
     respond_to do |format|
       format.any do
-        set_expiry(AssetManager.cache_control.max_age)
         if redirect_to_s3?
+          set_expiry(AssetManager.cache_control.max_age)
           redirect_to Services.cloud_storage.public_url_for(asset)
         elsif proxy_to_s3_via_nginx?
           url = Services.cloud_storage.presigned_url_for(asset)
           headers['X-Accel-Redirect'] = "/cloud-storage-proxy/#{url}"
-          headers['ETag'] = %{"#{asset.etag}"}
-          headers['Last-Modified'] = asset.last_modified.httpdate
+          headers['X-Accel-ETag'] = %{"#{asset.etag}"}
+          headers['X-Accel-Last-Modified'] = asset.last_modified.httpdate
           render nothing: true
         elsif proxy_to_s3_via_rails?
+          set_expiry(AssetManager.cache_control.max_age)
           body = Services.cloud_storage.load(asset)
           send_data(body.read, **AssetManager.content_disposition.options_for(asset))
         else
+          set_expiry(AssetManager.cache_control.max_age)
           send_file(asset.file.path, disposition: AssetManager.content_disposition.type)
         end
       end
--- a/config/application.rb
+++ b/config/application.rb
@@ -32,6 +32,9 @@ module AssetManager
     config.action_dispatch.rack_cache = nil
 
     config.action_dispatch.x_sendfile_header = "X-Accel-Redirect"
+
+    require "cache_control_middleware"
+    config.middleware.insert_before Rack::ETag, CacheControlMiddleware
   end
 
   mattr_accessor :aws_s3_bucket_name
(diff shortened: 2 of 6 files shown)


chrisroos:
@floehopper: I find it slightly confusing that the Rails app is responsible for sending the `Etag` and `Last-Modified` headers to nginx and that S3 is responsible for sending the `Cache-Control` header. What do you think about consistently sending these headers from either the Rails app or S3? Either by writing all headers to the S3 objects, or by ignoring the `Cache-Control` header from S3 (your option 2).

floehopper:
@chrisroos:

We can't specify the `ETag` or `Last-Modified` that S3 returns, so that's not an option unless we're happy to use the **new** values returned by S3 which I think we've already agreed we don't want to at this stage.

As I mentioned in my comment about option 2, ignoring headers like `Cache-Control` from S3 seems like a backward step, because we're going to want to use the S3 responses if/when we start redirecting to S3. If we were going to go down this route, I'd probably want to stop writing any headers to S3.

Although I admit it's a bit confusing, I think we should be using as many of the headers from S3 as we can. It's only the `ETag` and `Last-Modified` that we want to override **temporarily** until we're happy to use the values returned by S3.

floehopper:
Superseded by #164. Closing.

PR #159, opened 2017-08-29
Fail fast if S3 bucket name is not configured

### LINKED ISSUE #154: Fail fast if AWS_S3_BUCKET_NAME isn't set
I think we should prevent the Asset Manager app from starting if the `AWS_S3_BUCKET_NAME` environment variable is missing. We're committed to uploading assets to S3 and so I'd like to be alerted as early as possible if something is going to prevent that from happening.

floehopper:
The app will now fail fast with a KeyError is the `AWS_S3_BUCKET_NAME` is not set in development or production environments.

In the test environment, `AssetManager.aws_s3_bucket_name` is set to an easily identifiable value.

I've updated the README to reflect these changes and indicate which environment variables are required.

Fixes #154.


chrisroos:
@floehopper: I think the change is good but I think we should also update the README to make it clear that you'll need an AWS account and an S3 bucket in order to use the Asset Manager in development after this change.

floehopper:
Superseded by #167 - I decided to only make the change in production. Closing.

Issue #94, opened 2017-07-31
Asset Manager should generate IDs of objects stored on S3


chrisroos:
The objects we write to S3 are currently named using the BSON ID of the corresponding record in the Asset Manager database. We want to decouple the storage of the assets from the storage of their metadata in Mongo. We should generate our own IDs and use those to write the files to S3.

floehopper:
We're now actively investigating [an alternative approach](https://github.com/alphagov/asset-manager/milestone/4). Closing.

chrisroos:
We think this is going to be useful irrespective of whether we're redirecting to the assets, or serving them via an nginx proxy. Reopening.

PR #163, opened 2017-08-29
Asset Manager generates IDs of objects stored on S3

chrislo:
Before this PR we were using Asset#id to generate the `key` used to store the asset in the S3 bucket. We feel that this couples the storage of the assets a little too closely to Mongo and its mechanism for generating ids and may make it difficult to change database technology in the future. In addition it overloads further the `_id` field in the asset document which is already used to generate URLs in the API response and filenames for NFS storage.

Instead we'll use a UUID as the object key, generated using Ruby's RFC 4122[1] compliant `SecureRandom#uuid`. This means that the key generation is in the application code and decoupled from the database.

At this stage we are not considering changing how assets are stored on the NFS filesystem, or the object ID created for them in the JSON response.

I'd like a bit of feedback on: 

- the test for the `Asset` `uuid` field, at the moment I'm just expecting it to be a 36-character string, but I could probably use a regex or something and introduce a uniqueness constraint. 
- the best way to migrate existing data. I think I should iterate through the existing records in batches and generate UUIDs but this may require downtime. 

Fixes: #94

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -10,6 +10,8 @@ class Asset
   field :filename_history, type: Array, default: -> { [] }
   protected :filename_history=
 
+  field :uuid, type: String, default: -> { SecureRandom.uuid }
+
   field :access_limited, type: Boolean, default: false
   field :organisation_slug, type: String
 
--- a/lib/data_migration.rb
+++ b/lib/data_migration.rb
@@ -1,8 +1,7 @@
 class DataMigration
   def self.add_uuid_to_assets
     Asset.where(uuid: nil).each do |asset|
-      asset.uuid = SecureRandom.uuid
-      asset.save
+      asset.update_attribute(:uuid, SecureRandom.uuid)
     end
   end
 end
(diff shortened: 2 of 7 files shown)


chrisroos:
> the test for the Asset uuid field, at the moment I'm just expecting it to be a 36-character string, but I could probably use a regex or something and introduce a uniqueness constraint.

@chrislo: Did you consider stubbing the return value of `SecureRandom.uuid` and asserting that the value of `Asset#uuid` matches the stubbed value of `SecureRandom.uuid`. Would this be any better?

chrisroos:
> the best way to migrate existing data. I think I should iterate through the existing records in batches and generate UUIDs but this may require downtime.

Have you tried running this migration locally against a production-like copy of the asset-manager database? It doesn't feel as though it should take very long (and therefore shouldn't require any downtime) but I might be missing something.

chrisroos:
@chrislo: Are you planning to add the migration to update existing assets as part of this branch? If not, is there any danger of deploying this branch before all the assets have a uuid?

chrislo:
Thanks @chrisroos! I've addressed your feedback, I'll squash my fixes and open a new PR.

PR #164, opened 2017-08-29
Send more response headers when proxying to S3 via Nginx

floehopper:
And stop setting them on the S3 object.

Prior to this commit, the Rails app was both setting a `Cache-Control` response header on the S3 object *and* sending a `Cache-Control` response header in the response to Nginx instructing it to proxy the request to S3. Thus the user was receiving two duplicate `Cache-Control` headers. Although the values would have been the same, we want to avoid confusion and/or unexpected behaviour.

After this commit we no longer set the `Cache-Control`, `Content-Disposition`, or `Content-Type` headers on the S3 object. Instead the Rails app sends these headers to Nginx. We plan to update the Nginx configuration in a separate pull request against govuk-puppet in order to ensure the headers from Rails are added to the proxied response.

In #158 we considered preventing the Rails app from sending the `Cache-Control` header, but as @chrisroos pointed out, it was confusing that some headers were being set by Rails and some by S3. Given that we want to override the `ETag` and `Last-Modified` response headers to match those sent by Nginx when using the default Sendfile functionality (see Rails app to be the canonical source of *all* response headers when proxying to S3 via Nginx.

Note that the `Cache-Control` header is already being set via the call to `ApplicationController#set_expiry` and so we don't need to set it more verbosely like we have for the other headers.

Addresses the Asset Manager aspect of #149. As noted above, we will need to make some corresponding changes to govuk-puppet.

Supersedes #158.

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -23,6 +23,8 @@ class MediaController < ApplicationController
           headers['X-Accel-Redirect'] = "/cloud-storage-proxy/#{url}"
           headers['ETag'] = %{"#{asset.etag}"}
           headers['Last-Modified'] = asset.last_modified.httpdate
+          headers['Content-Disposition'] = AssetManager.content_disposition.header_for(asset)
+          headers['Content-Type'] = asset.content_type
           render nothing: true
         elsif proxy_to_s3_via_rails?
           body = Services.cloud_storage.load(asset)
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -90,21 +90,13 @@ class Asset
   end
 
   def save_to_cloud_storage
-    Services.cloud_storage.save(self, cloud_storage_options)
+    Services.cloud_storage.save(self)
   rescue => e
     Airbrake.notify_or_ignore(e, params: { id: self.id, filename: self.filename })
     raise
   end
 
 protected
-
-  def cloud_storage_options
-    {
-      cache_control: AssetManager.cache_control.header,
-      content_disposition: AssetManager.content_disposition.header_for(self),
-      content_type: content_type
-    }
-  end
 
   def valid_filenames
     filename_history + [filename]
--- a/spec/controllers/media_controller_spec.rb
+++ b/spec/controllers/media_controller_spec.rb
@@ -239,6 +239,7 @@ RSpec.describe MediaController, type: :controller do
         let(:cloud_storage) { double(:cloud_storage) }
         let(:presigned_url) { 'https://s3-host.test/presigned-url' }
         let(:last_modified) { Time.zone.parse("2017-01-01 00:00") }
+        let(:content_disposition) { instance_double(ContentDispositionConfiguration) }
 
         before do
           allow(controller).to receive(:proxy_to_s3_via_nginx?).and_return(true)
@@ -247,6 +248,9 @@ RSpec.describe MediaController, type: :controller do
           allow(controller).to receive(:asset).and_return(asset)
           allow(asset).to receive(:etag).and_return("599ffda8-e169")
           allow(asset).to receive(:last_modified).and_return(last_modified)
+          allow(asset).to receive(:content_type).and_return('content-type')
+          allow(AssetManager).to receive(:content_disposition).and_return(content_disposition)
+          allow(content_disposition).to receive(:header_for).with(asset).and_return("content-disposition")
         end
 
         it "responds with 200 OK" do
@@ -262,6 +266,16 @@ RSpec.describe MediaController, type: :controller do
         it "sends Last-Modified response header in HTTP time format" do
           do_get
           expect(response.headers["Last-Modified"]).to eq("Sun, 01 Jan 2017 00:00:00 GMT")
+        end
+
+        it "sends Content-Disposition response header based on asset filename" do
+          do_get
+          expect(response.headers["Content-Disposition"]).to eq("content-disposition")
+        end
+
+        it "sends Content-Type response header based on asset file extension" do
+          do_get
+          expect(response.headers["Content-Type"]).to eq("content-type")
         end
 
         it "instructs nginx to proxy the request to S3" do
--- a/spec/models/asset_spec.rb
+++ b/spec/models/asset_spec.rb
@@ -1,5 +1,4 @@
 require "rails_helper"
-require "content_disposition_configuration"
 
 RSpec.describe Asset, type: :model do
   include DelayedJobHelpers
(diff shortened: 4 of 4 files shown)


chrisroos:
I've tested this branch plus the associated changes in https://github.com/alphagov/govuk-puppet/pull/6362 in my Development VM. I've used a simple text file to confirm that we're seeing the  behaviour we expect:

NOTE. I ran asset-manager in production environment so that the Cache-Control header is set to something realistic.

NOTE. I modified the nginx config to set a `Debug-Served-By` header (to either 'SendFile' or 'S3Proxy') so that I can tell how nginx handled the request.

NOTE. I ordered the response headers in the examples below so that they were easier to visually compare.

```
# Create text file
$ echo `date` > tmp.txt

# Upload to Asset Manager
$ curl http://asset-manager.dev.gov.uk/assets/ --form "asset[file]=@tmp.txt"

# Request from file system
$ curl http://asset-manager.dev.gov.uk/media/59a5a42e759b7420dd0f877c/tmp.txt -v
< HTTP/1.1 200 OK
* Server nginx is not blacklisted
< Accept-Ranges: bytes
< Cache-Control: max-age=86400, public
< Connection: close
< Content-Disposition: inline; filename="tmp.txt"
< Content-Length: 29
< Content-Type: text/plain
< Date: Tue, 29 Aug 2017 17:44:29 GMT
< Debug-Served-By: SendFile
< ETag: "59a5a42e-1d"
< Last-Modified: Tue, 29 Aug 2017 17:28:14 GMT
< Server: nginx
< Vary: Accept-Encoding

# Request from S3
$ curl http://asset-manager.dev.gov.uk/media/59a5a42e759b7420dd0f877c/tmp.txt?proxy_to_s3_via_nginx=true -v
< HTTP/1.1 200 OK
* Server nginx is not blacklisted
< Accept-Ranges: bytes
< Cache-Control: max-age=86400, public
< Connection: close
< Content-Disposition: inline; filename="tmp.txt"
< Content-Length: 29
< Content-Type: text/plain
< Date: Tue, 29 Aug 2017 17:45:13 GMT
< Debug-Served-By: S3Proxy
< ETag: "59a5a42e-1d"
< Last-Modified: Tue, 29 Aug 2017 17:28:14 GMT
< Server: nginx
```

chrisroos:
I want to double check this with fresh eyes in the morning before approving the PR.

floehopper:
@chrisroos:

Thanks for checking this.

I think the extra `Vary` header might be [being added by the `gzip_vary on` statement in the overall `nginx.conf`][1].

I think it might also be worth comparing the headers returned from the `assets-origin.dev.gov.uk` host. I've been seeing an extra `X-Frame-Options: DENY` which I think is being added by [this `add_header` statement](https://github.com/alphagov/govuk-puppet/blob/e9aaed83505a26f316ed0f38c370425bbe356d86/modules/nginx/templates/proxy-vhost.conf#L64-L66).

[1]: https://github.com/alphagov/govuk-puppet/blob/e9aaed83505a26f316ed0f38c370425bbe356d86/modules/nginx/templates/etc/nginx/nginx.conf.erb#L80

floehopper:
@chrisroos:

>  I've been seeing an extra `X-Frame-Options: DENY` which I think is being added by [this `add_header` statement](https://github.com/alphagov/govuk-puppet/blob/e9aaed83505a26f316ed0f38c370425bbe356d86/modules/nginx/templates/proxy-vhost.conf#L64-L66).

I wonder whether we either need to duplicate this `add_header` statement in the "proxying to S3" location block or somehow copy the header set by in the outer block...?

chrisroos:
I've just tested this combination again, but this time with a PDF file:

```
# Upload PDF to Asset Manager
$ curl http://asset-manager.dev.gov.uk/assets/ --form "asset[file]=@hello-world.pdf"

# Request from file system
$ curl http://assets-origin.dev.gov.uk/media/59a69295759b743ff350a7c0/hello-world.pdf -v > /dev/null
< HTTP/1.1 200 OK
* Server nginx is not blacklisted
< Accept-Ranges: bytes
< Access-Control-Allow-Headers: origin, authorization
< Access-Control-Allow-Methods: GET, OPTIONS
< Access-Control-Allow-Origin: *
< Cache-Control: max-age=86400, public
< Connection: close
< Content-Disposition: inline; filename="hello-world.pdf"
< Content-Length: 7781
< Content-Type: application/pdf
< Date: Wed, 30 Aug 2017 10:27:48 GMT
< ETag: "59a69295-1e65"
< Last-Modified: Wed, 30 Aug 2017 10:25:25 GMT
< Server: nginx
< Strict-Transport-Security: max-age=31536000
< X-Frame-Options: DENY

# Request from S3
$ curl http://assets-origin.dev.gov.uk/media/59a69295759b743ff350a7c0/hello-world.pdf?proxy_to_s3_via_nginx=true -v
< HTTP/1.1 200 OK
* Server nginx is not blacklisted
< Accept-Ranges: bytes
< Access-Control-Allow-Headers: origin, authorization
< Access-Control-Allow-Methods: GET, OPTIONS
< Access-Control-Allow-Origin: *
< Cache-Control: max-age=86400, public
< Connection: close
< Content-Disposition: inline; filename="hello-world.pdf"
< Content-Length: 7781
< Content-Type: application/pdf
< Date: Wed, 30 Aug 2017 10:28:08 GMT
< ETag: "59a69295-1e65"
< Last-Modified: Wed, 30 Aug 2017 10:25:25 GMT
< Server: nginx
< Strict-Transport-Security: max-age=31536000
```

chrisroos:
> I think the extra `Vary` header might be being added by the `gzip_vary on` statement in the overall `nginx.conf`.

I'm not too worried about the presence of the `Vary` header. Nginx is only configured to enable compression for text/css, text/xml, application/x-javascript, application/atom+xml and text/plain and they make up a [tiny percentage of the overall assets we have uploaded](https://github.com/alphagov/asset-manager/blob/master/docs/existing_assets.md#file-extensions).

chrisroos:
> I wonder whether we either need to duplicate this add_header statement in the "proxying to S3" location block or somehow copy the header set by in the outer block...?

Possibly. But I think we can investigate that separately to this PR. Can you open an issue to remind us to look into it?

floehopper:
@chrisroos:

>> I wonder whether we either need to duplicate this add_header statement in the "proxying to S3" location block or somehow copy the header set by in the outer block...?
>
>Possibly. But I think we can investigate that separately to this PR. Can you open an issue to remind us to look into it?

I've created this [new issue](https://github.com/alphagov/asset-manager/issues/166).

PR #173, opened 2017-08-30
Use a generated UUID for storing Assets on S3

chrislo:
Before this PR we were using Asset#id to generate the key used to store the asset in the S3 bucket. We feel that this couples the storage of the assets a little too closely to Mongo and its mechanism for generating ids and may make it difficult to change database technology in the future. In addition it overloads further the _id field in the asset document which is already used to generate URLs in the API response and filenames for NFS storage.

Instead we'll use a UUID as the object key, generated using Ruby's RFC 4122[1] compliant SecureRandom#uuid. This means that the key generation is in the application code and decoupled from the database.

At this stage we are not considering changing how assets are stored on the NFS filesystem, or the object ID created for them in the JSON response.

I ran the data migration in this PR locally against a copy of the production database from today (with 58,541 asset records):

```
$ time be rake migrate:add_uuid_to_assets

real	1m22.772s
user	1m8.585s
sys	0m4.795s
```

I'm fairly confident that we should be able to run that against the production environments without taking any special precautions.

### DIFF
--- a/spec/lib/data_migration_spec.rb
+++ b/spec/lib/data_migration_spec.rb
@@ -1,0 +1,23 @@
+require 'rails_helper'
+require 'data_migration'
+
+RSpec.describe DataMigration do
+  let!(:migrated_asset) { FactoryGirl.create(:asset, uuid: 'some-uuid') }
+  let!(:unmigrated_asset) { FactoryGirl.create(:asset, uuid: nil) }
+
+  describe 'add_uuid_to_assets' do
+    it 'generates a new uuid for assets with a blank uuid' do
+      DataMigration.add_uuid_to_assets
+
+      unmigrated_asset.reload
+      expect(unmigrated_asset.uuid).to_not be_blank
+    end
+
+    it 'does not generate a new uuid for an asset when one already exists' do
+      DataMigration.add_uuid_to_assets
+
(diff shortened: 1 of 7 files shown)


spec/lib/data_migration_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
+require 'data_migration'
+
+RSpec.describe DataMigration do
+  let!(:unmigrated_asset) { FactoryGirl.create(:asset, uuid: 'some-uuid') }
+  let!(:migrated_asset) { FactoryGirl.create(:asset) }
```
chrisroos: Won't this asset have a `#uuid` given that there's a default block in the `field :uuid` definition? Even if that's not true, I think it might be clearer to explicitly set the `uuid` to `nil` in this set-up.

chrisroos: Should these names be reversed (i.e. unmigrated has a blank uuid and migrated has a real uuid)?

chrislo: That would be clearer - addressed in 7dfc266

chrislo:
Thanks @chrisroos! I'll rebase and force push before merging.

Issue #175, opened 2017-08-30
Ensure Content-Type is the same irrespective of whether asset is served from NFS or S3


chrisroos:
While testing issue #149 I noticed that the value of the `Content-Type` header was different when the file was served from NFS (`application/octet-stream`) vs when it was served from S3 (`*/*; charset=utf-8`).

We should ensure the value of the `Content-Type` header for the [commonly uploaded asset filetypes][asset-filetypes] is the identical when served from NFS or S3. It might be that the problem I saw in issue #149 was due to me using an unusual file extension (`.dat`) and that for everything else it might work just fine.

[asset-filetypes]: https://github.com/alphagov/asset-manager/blob/master/docs/existing_assets.md#file-extensions

chrisroos:
I've investigated the `Content-Type` being reported for each of [these 20 file extensions](https://github.com/alphagov/asset-manager/blob/master/docs/existing_assets.md#file-extensions).

In each case I created an empty file with the relevant file extension using `touch file.<extension>`. I uploaded the file to my local asset-manager and then requested it from the filesystem and from S3. I did this testing in the Rails development environment but also tested a couple of them in the Rails production environment to ensure we were seeing the same results.

| File extension | Filesystem               | S3                 |
| -------------- | ------------------------ | ------------------ |
| pdf            | application/pdf          | application/pdf    |
| jpg            | image/jpeg               | image/jpeg         |
| doc            | application/octet-stream | \*/*; charset=utf-8 |
| png            | image/png                | image/png          |
| gif            | image/gif                | image/gif          |
| docx           | application/octet-stream | \*/*; charset=utf-8 |
| xlsx           | application/octet-stream | \*/*; charset=utf-8 |
| xls            | application/octet-stream | \*/*; charset=utf-8 |
| odt            | application/octet-stream | \*/*; charset=utf-8 |
| ppt            | application/octet-stream | \*/*; charset=utf-8 |
| zip            | application/zip          | application/zip    |
| ods            | application/octet-stream | \*/*; charset=utf-8 |
| pptx           | application/octet-stream | \*/*; charset=utf-8 |
| xlsm           | application/octet-stream | \*/*; charset=utf-8 |
| url            | application/octet-stream | \*/*; charset=utf-8 |
| csv            | text/csv                 | text/csv           |
| jpeg           | image/jpeg               | image/jpeg         |
| pd             | application/octet-stream | \*/*; charset=utf-8 |
| txt            | text/plain               | text/plain         |
[...]

PR #177, opened 2017-08-31
Avoid sending empty Content-Type header when proxying to S3

chrisroos:
Prior to this commit, `Asset#content_type` would return an empty string
if the file extension wasn't found by `Mime::Type.lookup`. This meant
that the `Content-Type` response header was also being set to an empty
string. Something else in the Rails stack (presumably some middleware)
was then replacing this empty string with "\*/\*; charset=utf-8". The
effect of this was that the `Content-Type` response header for files
with unrecognised mime types would be 'application/octet-stream' when
serving the file from the disk and '\*/\*; charset=utf-8' when serving the
file from S3.

This commit ensures we send the default content type
('application/octet-stream') for files with unrecognised mime types when
proxying requests to S3.

Note that `Mime::Type.lookup_by_extension(nil)` returns the same as
`Mime::Type.lookup_by_extension('unrecognised-extension')` which allows
me to get rid of the `if extension.present?` condition.

Fixes issue #175.

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -57,11 +57,8 @@ class Asset
   end
 
   def content_type
-    if extension.present?
-      Mime::Type.lookup_by_extension(extension).to_s
-    else
-      AssetManager.default_content_type
-    end
+    mime_type = Mime::Type.lookup_by_extension(extension)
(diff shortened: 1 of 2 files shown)


chrisroos:
Thanks @floehopper! Merging.

PR #199, opened 2017-09-18
Enable Rails asset pipeline

floehopper:
We're going to want to use this for the `thumbnail-placeholder.png` image which we're going to move across from Whitehall.

We've set the assets prefix in order to: (a) avoid a conflict with the `/assets` route defined within Asset Manager for its API endpoint; and (b) conform with the approach taken by other apps like Static to allow routing [via the Nginx configuration for the assets-origin host][1].

Note that we don't need any of `sass-rails`, `coffee-rails` and `uglifier` gems, because we only need to serve images.

[1]:
https://github.com/alphagov/govuk-puppet/blob/994d5424c7e4e28e903039f9f603e8d345b4f01d/hieradata/common.yaml#L1545-L1566

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -7,7 +7,7 @@ require File.expand_path('../boot', __FILE__)
 require "action_controller/railtie"
 require "action_mailer/railtie"
 # require "action_view/railtie"
-# require "sprockets/railtie"
+require "sprockets/railtie"
 # require "rails/test_unit/railtie"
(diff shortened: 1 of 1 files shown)


floehopper:
Note that we will probably need to add some code to the `asset-manager` bit of the `govuk-app-deployment` repo to actually do the asset pre-compilation on deployment.

Issue #196, opened 2017-09-18
Make Asset#uuid read-only


floehopper:
Possibly using [`Mongoid::Attributes::Readonly::ClassMethods#attr_readonly`][1].

[1]: http://www.rubydoc.info/github/mongoid/mongoid/Mongoid/Attributes/Readonly/ClassMethods#attr_readonly-instance_method

chrisroos:
I've opened PR #209 to make `Asset#uuid` read-only.

chrisroos:
PR #209 has been merged so this issue can be closed.

PR #209, opened 2017-09-20
Add validation to Asset#uuid

chrisroos:
This branch started out addressing issue #196 (make `Asset#uuid` read-only) but I've extended it to add validation to check that `Asset#uuid`s are present, unique and conform to the UUID format defined in RFC4122.

I've checked all the Assets in production and confirmed that the existing UUIDs all conform to the validation added in this branch. I did this by downloading the asset-manager database and running the following command in my local Rails console:

```
invalid_uuid_assets = Asset.all.select { |asset| asset.valid?; asset.errors[:uuid].present? }; nil
p invalid_uuid_assets.count
=> 0
```


chrisroos:
Thanks @chrislo! I'll get this merged.

chrisroos:
> with the change to only allow uuid to be set at instantiation time the format and uniqueness validations are probably not strictly necessary but I think the additional validation means we have peace of mind that we won't accidentally change how they are generated.

I agree that they feel a little unnecessary but given that it was previously possible to set the `#uuid` to _anything_ before the `Asset` was saved I think they do add some benefit. A slightly nicer approach might have been to prevent the `#uuid` from ever being set by anything other than the `Asset` object but I don't think that's possible unfortunately.

PR #210, opened 2017-09-20
Serve public Whitehall assets from asset manager

floehopper:
We've introduced a new endpoint for the Asset Manager API to allow us to add Whitehall assets. This new endpoint is necessary, because, for now, we need Whitehall to tell Asset Manager the URL path at which the asset should be served. We haven't yet added this endpoint to `gds-api-adapters` nor is Whitehall using it.

We've also introduced a new Asset Manager route and controller to serve these publicly-accessible Whitehall assets based on the functionality in Whitehall's `PublicUploadsController`. No such assets exist at the moment and in any case no requests for such assets should be made at the moment.

The image placeholder used when an asset is being virus scanned is now served by Asset Manager as a Rails pipeline asset (see https://github.com/alphagov/govuk-puppet/pull/6502 & https://github.com/alphagov/asset-manager/pull/199); whereas the HTML placeholder is still served by Whitehall. There are a couple of small differences in the behaviour when an asset does not exist or is being virus scanned - see commit notes for details.


floehopper:
@chrisroos & @chrislo: I'm not 100% sure I've captured everything I need to in this PR, but I think there's enough here to ask you to have a look and see what you think.

floehopper:

I'll get the branch rebased against master, resolve the conflicts and force push in preparation for merging.

Issue #179, opened 2017-09-05
Use Sidekiq instead of Delayed Job


chrisroos:
This was suggested when we asked whether there were any supported ways of viewing the status of Delayed Job queues. The majority of apps use Sidekiq and there's already [support for monitoring these queues](https://docs.publishing.service.gov.uk/manual/monitor-sidekiq-workers.html).

TODO: 
- [x] Configure sidekiq in asset manager
- [x] Setup redis configuration in puppet
- [x] Setup worker process in puppet
- [x] Make sure worker is restarted on deployment in govuk-app-deployment
- [x] Configure [monitoring of sidekiq](https://docs.publishing.service.gov.uk/manual/setting-up-new-sidekiq-monitoring-app.html)
- [x] Migrate virus worker job to sidekiq
- [x] Migrate s3 uploaded job to sidekiq
- [ ] Remove delayed job

tijmenb:
FYI: https://github.com/alphagov/govuk_sidekiq contains configuration and explains how to set up Sidekiq on GOV.UK.

chrislo:
The following branches are relevant: 

- [asset-manager/migrate-to-sidekiq-from-delayed-job](https://github.com/alphagov/asset-manager/tree/migrate-to-sidekiq-from-delayed-job)
- [govuk-puppet/configure-sidekiq-for-asset-manager](https://github.com/alphagov/govuk-puppet/tree/configure-sidekiq-for-asset-manager)
- [govuk-app-deployment/restart-asset-manager-sidekiq-on-deployment](https://github.com/alphagov/govuk-app-deployment/tree/restart-asset-manager-sidekiq-on-deployment)

chrislo:
I've [opened a PR](https://github.com/alphagov/govuk-puppet/pull/6483) for the `govuk-puppet` changes, and [one for](https://github.com/alphagov/govuk-app-deployment/pull/178) the `govuk-app-deployment` changes.

chrislo:
With the above two PRs merged monitoring of the queues now seems to work on integration. You can see the sidekiq dashboard by forwarding a port on integration (`ssh backend-1.backend.integration -CNL 9000:127.0.0.1:80`) and then visiting http://localhost:9000/asset-manager. 

[...]

PR #221, opened 2017-09-20
Migrate S3 upload worker to sidekiq

chrislo:
This PR moves the uploading of assets to S3 from a delayed_job worker to a sidekiq worker. 

There is a reference to `asset-manager/lib/tasks/s3.rake` to `save_to_cloud_storage`. Looking at this rake task, I think it should be removed as it has already been run (and if we need it again it'll be in a different form) so I haven't converted it. I've [opened a new PR](https://github.com/alphagov/asset-manager/pull/222) to do that. 

Contributes to: #179

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -36,7 +36,7 @@ class Asset
     end
 
     after_transition to: :clean do |asset, _|
-      asset.delay.save_to_cloud_storage
+      SaveToCloudStorageWorker.perform_async(asset)
     end
 
     event :scanned_infected do
@@ -113,10 +113,7 @@ class Asset
   end
 
   def save_to_cloud_storage
-    Services.cloud_storage.save(self)
-  rescue => e
-    Airbrake.notify_or_ignore(e, params: { id: self.id, filename: self.filename })
-    raise
+    SaveToCloudStorageWorker.new.perform(self)
   end
 
 protected
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -36,7 +36,7 @@ class Asset
     end
 
     after_transition to: :clean do |asset, _|
-      SaveToCloudStorageWorker.perform_async(asset)
+      SaveToCloudStorageWorker.perform_async(asset.id)
     end
 
     event :scanned_infected do
@@ -113,7 +113,7 @@ class Asset
   end
 
   def save_to_cloud_storage
-    SaveToCloudStorageWorker.new.perform(self)
+    SaveToCloudStorageWorker.new.perform(self.id)
   end
 
 protected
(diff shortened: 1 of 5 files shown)


app/models/asset.rb
```diff
@@ -48,7 +48,7 @@ class Asset
     end
 
     after_transition to: :clean do |asset, _|
-      asset.delay.save_to_cloud_storage
+      SaveToCloudStorageWorker.perform_async(asset)
```
floehopper: Did you discover how Sidekiq stores the parameters passed to `perform_async`? Is there any downside of passing the model instance vs its ID?

spec/models/asset_spec.rb
```diff
@@ -256,69 +256,27 @@
...
-        it 'reports the exception to Errbit via Airbrake' do
-          expect(Airbrake).to receive(:notify_or_ignore)
-            .with(exception, params: { id: asset.id, filename: asset.filename })
-
-          asset.save_to_cloud_storage rescue exception_class
-        end
-
-        it 're-raises the exception so Delayed::Job will re-try it' do
-          allow(Airbrake).to receive(:notify_or_ignore)
+    let(:asset) { FactoryGirl.create(:asset) }
```
floehopper: Minor: Does this asset need to be saved in the database?

spec/workers/save_to_cloud_storage_worker_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+        let(:exception_class) { Class.new(StandardError) }
+        let(:exception) { exception_class.new }
+
+        before do
+          allow(cloud_storage).to receive(:save).and_raise(exception)
+        end
+
+        it 'reports the exception to Errbit via Airbrake' do
+          expect(Airbrake).to receive(:notify_or_ignore)
+          .with(exception, params: { id: asset.id, filename: asset.filename })
```
floehopper: Very minor: I think I'd expect this line to be indented two more spaces. Does Rubocop not suggest that?

app/models/asset.rb
floehopper: Ignore my previous comment - I've just seen the next commit where you change this to pass the asset ID.

spec/models/asset_spec.rb
floehopper: I suspect that maybe it didn't need to be saved in the database in this commit, but does need to be in the next commit when the asset ID is used. So I suspect you can ignore my previous comment.

chrislo: Good catch - I don't think I'll edit the previous commit to fix this, but I think you're right that I don't need a persisted asset in that commit.

spec/workers/save_to_cloud_storage_worker_spec.rb
chrislo: No - it seems to pass a full check locally and on CI.

chrislo:
Rebasing and force-pushing.

Issue #145, opened 2017-08-23
Regularly copy assets from Production to Staging and Integration


chrisroos:
This is on the assumption that we currently regularly copy production assets to staging and integration. We should validate this assumption before doing any work on this.

floehopper:
We think we only need this when stop saving assets to NFS.

chrisroos:
I wonder whether we might need to do this before we start serving a percentage of asset requests by proxying to S3 in integration or staging... If we don't copy the assets between the S3 environments (i.e. production to staging and integration) then the database in integration/staging won't be in sync with the assets on S3 after the nightly database copy. If the database and S3 are out of sync then any requests served by proxying to S3 will fail.

I think we can also avoid this problem by ensuring that we don't serve any asset requests by proxying to S3 in integration or staging.

@chrislo / @floehopper: What do you think we should do about this? Should we prioritise copying the assets between S3 buckets? Or just keep the option disabled in integration/staging in the short term?

chrislo:
I *think* that the production databases for the publishing applications also get copied over nightly - so in the case of manuals publisher on integration, it will have asset urls that point to the production assets for any assets added yesterday - I think that will continue to work when we switch to proxying to S3. Any newly uploaded assets in integration will be added to the integration bucket and will work as normal. Is that what you mean @chrisroos?

chrisroos:
Aha, you're quite right @chrislo. I'd forgotten that we mostly/always(?) store asset URLs in content rather than an identifier that needs to be resolved by the asset-manager API. I think I'm happy that this isn't as much of a problem as I'd imagined.

floehopper:
The other day we worked out that this is in fact an issue for *Whitehall* assets, because (at least some of) the URLs of existing assets are generated in the Rails app (rather than being embedded in content in the Content Store) and are therefore potentially environment-specific. This implies that we do need to setup syncing between production and staging/integration before we can remove the files on NFS corresponding to Whitehall assets within Asset Manager.

floehopper:
I've opened the following PRs to implement this:

1. https://github.com/alphagov/govuk-terraform-provisioning/pull/143
2. https://github.com/alphagov/govuk-puppet/pull/6763
3. https://github.com/alphagov/env-sync-and-backup/pull/34

Note that 1 & 2 need to be merged and applied/deployed before 3 can be merged and deployed.

floehopper:
I've just merged https://github.com/alphagov/govuk-puppet/pull/6763.

[...]

PR #222, opened 2017-09-20
Remove upload_all_clean_assets rake task

chrislo:
This task has already been run and doesn't need to be run again in
this form. If we need to do something like this later, we can fetch it
out of the history.

This task was [run on integration](https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/2908/) at 05-Sep-2017 11:51:49. 

It was run on production at 06-Sep-2017 09:42:00. 

It has not been run on staging - we [have an issue](https://github.com/alphagov/asset-manager/issues/145) to work out how to replicate between environments.


chrislo:
Rebasing before merging.

PR #227, opened 2017-09-21
Migrate virus check job to sidekiq

chrislo:
Contributes to: #179 

This PR migrates the virus scan delayed job worker to sidekiq. It leaves `Asset#scan_for_viruses` in place in case there are still jobs in delayed_job that expect that method to be present. This can be removed in future when we are sure there are no more jobs being added to the delay job queue.


.rubocop_todo.yml
```diff
@@ -68,5 +69,6 @@ RSpec/VerifiedDoubles:
     - 'spec/controllers/media_controller_spec.rb'
     - 'spec/lib/virus_scanner_spec.rb'
     - 'spec/models/asset_spec.rb'
+    - 'spec/workers/virus_scan_worker_spec.rb'
```
floehopper: Very minor: Do we definitely still need the exceptions for `spec/models/asset_spec.rb` in these two lists? i.e. are there still Rubocop violations in the `Asset` spec?

spec/workers/virus_scan_worker_spec.rb
```diff
@@ -0,0 +1,68 @@
+require 'rails_helper'
+
+RSpec.describe VirusScanWorker, type: :worker do
```
floehopper: Very minor: What does `type: :worker` do and do we definitely need it?

spec/workers/virus_scan_worker_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  let(:worker) { described_class.new }
```
floehopper: Very minor: I might be wrong, but I think you might be able to achieve this more succinctly with `subject(:worker)`.

spec/workers/virus_scan_worker_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  let(:asset) { FactoryGirl.create(:asset) }
+
+  it "calls out to the VirusScanner to scan the file" do
+    scanner = double("VirusScanner")
+    expect(VirusScanner).to receive(:new).with(asset.file.path).and_return(scanner)
```
floehopper: Minor: I would move the first of these lines into a `let(:scanner)` statement and use an `instance_double` of `VirusScanner`. And I'd move the second line into a `before` block and convert the `expect` into an `allow`.

floehopper: I've just realised that you've moved these examples from the `Asset` spec, so you can ignore my previous comment and I will refrain from making more suggestions for improvement in this commit.

app/models/asset.rb
```diff
@@ -135,7 +135,7 @@ def reset_state
   end
 
   def schedule_virus_scan
-    self.delay.scan_for_viruses if self.unscanned?
+    VirusScanWorker.perform_async(self.id) if self.unscanned?
```
floehopper: Minor: I know you mentioned it in a previous commit note, but I think it would be nice to mention in the note for this commit that you're leaving `scan_for_viruses` executing the virus scanning synchronously to support any remaining Delayed Jobs in the queue.

spec/models/asset_spec.rb
```diff
@@ -226,43 +226,38 @@
...
     it "does not schedule a scan after update if the file is unchanged" do
       a = FactoryGirl.create(:clean_asset)
       a.created_at = 5.days.ago
-      expect {
-        a.save!
-      }.not_to change(Delayed::Job, :count)
+
+      expect(VirusScanWorker).not_to receive(:perform_async).with(a)
+
+      a.save!
```
floehopper: Minor: I wonder whether this means we can stop including `DelayedJobHelpers` in this spec?

spec/requests/virus_scanning_spec.rb
```diff
@@ -56,7 +56,7 @@ def initialize
     get "/media/#{asset.id}/eicar.com"
     expect(response).to have_http_status(:not_found)
 
-    run_all_delayed_jobs
+    VirusScanWorker.drain
```
floehopper: Minor: I wonder whether this means we can stop automatically including the `DelayedJobHelpers` module and maybe even delete it? Maybe we can do that as a separate PR in which we remove all DJ related code...?

spec/models/asset_spec.rb
```diff
@@ -226,43 +226,38 @@
...
     end
 
     it "does not schedule a scan after update if the file is unchanged" do
       a = FactoryGirl.create(:clean_asset)
       a.created_at = 5.days.ago
-      expect {
-        a.save!
-      }.not_to change(Delayed::Job, :count)
+
+      expect(VirusScanWorker).not_to receive(:perform_async).with(a)
```
floehopper: I think the `with` should be passed `a.id` vs `a`. Having said that, I think you could drop the `with` altogether here.

spec/requests/virus_scanning_spec.rb
chrislo: I was thinking of removing all of the delayed job code in one PR - but yes, it does leave things a bit messy until then.

.rubocop_todo.yml
chrislo: Yes, there's still unrelated violations for these exceptions even with the virus related blocks moved across. I'll see if I can address those in a separate PR.

spec/workers/virus_scan_worker_spec.rb
chrislo: No - it was added by the sidekiq rspec generator, I'll remove it.

spec/workers/virus_scan_worker_spec.rb
chrislo: Rspec supports [implicit subject](https://relishapp.com/rspec/rspec-core/v/2-0/docs/subject/implicit-subject) but using them to make this more succinct violates [RuboCop::Cop::RSpec::NamedSubject](http://www.rubydoc.info/gems/rubocop-rspec/1.6.0/RuboCop/Cop/RSpec/NamedSubject). There's not much difference between `let(:worker) { described_class.new }` and `subject(:worker) { described_class.new }` so I'll leave it.

app/models/asset.rb
chrislo: Thanks! I'll add a note to the commit message.

spec/models/asset_spec.rb
chrislo: Good catch - I'll fix that. Thanks!

chrislo:
Thanks for the review @floehopper. I've addressed your feedback by rebase and am force-pushing before merging.

PR #228, opened 2017-09-21
Require legacy_url_path when creating Whitehall asset

floehopper:
This introduces a `WhitehallAsset` subclass of `Asset` and pulls down the `legacy_url_path` attribute from `Asset` to `WhitehallAsset` along with related behaviour. The validation for `WhitehallAsset` is then modified to require the presence of the `legacy_url_path` attribute. The effect of this will be to respond with an error if no `legacy_url_path` is supplied to `WhitehallAssetsController#create`.


chrisroos:
I'm going to review this now.

Issue #234, opened 2017-09-26
Fix ETag & Last-Modified headers for Whitehall assets served from Asset Manager


floehopper:
Ref: https://github.com/alphagov/asset-manager/issues/220

As explained in [this asset-manager PR][7], Nginx generates the ETag from a combination of the underlying file's `mtime` (last modified) and its size. Since in the tests above we've uploaded different incarnations of the files, I'd expect the `mtime`s to be different, which explains why the first part of the `ETag` value is different in the two scenarios.

We'll need to work out whether we want to keep this unchanged in order to avoid invalidating all cached assets and, if we do, how do we want to do it.

[7]: https://github.com/alphagov/asset-manager/pull/157

floehopper:
I've just opened #242 to allow `etag` and `last_modified` to be specified when creating a Whitehall asset via the Asset Manager API.

floehopper:
I've added notes to the descriptions for #215 & #216 to remind us to set `legacy_etag` and `legacy_last_modified` when migrating existing Whitehall assets to Asset Manager in order to avoid wholesale cache invalidation. On that basis I'm happy to close this issue.

PR #240, opened 2017-09-27
Store asset attributes generated from file metadata in database

### LINKED ISSUE #182: Store file metadata in database so we don't have to obtain in from NFS
The following Asset attributes are derived from file metadata:

* `Asset#etag` derived from `File.stat`
* `Asset#last_modified` derived from `File.stat`
* `Asset#md5_hexdigest` derived from `File#read`

If we want to switch off NFS, but keep proxying to S3 via Nginx, we'll need to store these values in the database.

The first two are only relevant if we want to continue to avoid unnecessary cache invalidation.

The last one is only relevant if we want to continue to avoid uploading identical asset files to S3 multiple times as per ff2ed6163e80327d5b51f0cabfe16b8013601a84.

floehopper:
Previously the `Asset` attribute methods `#etag`, `#last_modified` and `#md5_hexdigest` generated a value from the underlying file metadata. However, we want to remove dependencies on the "local" (NFS) filesystem so that it can be removed in preference to the S3 file store. To this end, I've made the following changes:

* For new assets, the attribute values generated from the underlying file metadata are stored in the database.
* For existing assets, I've created the `assets:store_values_generated_from_file_metadata` rake task so we can use it to store the values generated from the underlying file metadata.
* For all assets, the attribute methods now return the value in database in preference to generating a value from the underlying metadata. However, for the time being they will fallback to generating the attribute value from the underlying file metadata if no value is stored in the database. Once we've run the rake task (see above) in production, we'll be able to remove these fallbacks.

My motivation for working on this was to make it easier to allow Whitehall to specify the `etag` and `last_modified` values when it creates an asset in Asset Manager. See #234 for more details.

Closes #182.

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -17,6 +17,7 @@ class Asset
   field :organisation_slug, type: String
 
   field :etag, type: String
+  field :last_modified, type: Time
 
   validates :file, presence: true
   validates :organisation_slug, presence: true, if: :access_limited?
@@ -93,6 +94,10 @@ class Asset
   end
 
   def last_modified
+    self[:last_modified] || last_modified_from_file
+  end
+
+  def last_modified_from_file
     file_stat.mtime
   end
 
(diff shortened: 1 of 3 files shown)


floehopper:

PR #241, opened 2017-09-27
Add default REDIS_HOST env var to dockerfile

kevindew:
This project is now dependent on Redis since Sidekiq was introduced,
this adds in a default value for the hostname.


floehopper:
@kevindew: Thanks for making this change. Is there a build failing somewhere that highlights this problem?

thomasleese:
@floehopper It was the publishing-e2e-tests that were failing,  see: https://github.com/alphagov/publishing-e2e-tests/pull/51

https://ci.integration.publishing.service.gov.uk/job/publishing-e2e-tests/job/test-against/414/

floehopper:
@thomasleese: Ah. Thanks for explaining! 👍

Issue #216, opened 2017-09-20
Upload all existing Whitehall non-attachment assets not modelled in the database to Asset Manager


chrisroos:
This includes everything in the `/mnt/uploads/whitehall/clean/` directory on the asset hosts that isn't:

* a Whitehall image (these are being handled in #215)
* an HMRC or Number 10 upload (these are being handled in #217)
* assets stored in government/uploads/system/uploads (these are being handled in #420)
* a Whitehall attachment (we're not migrating these at the moment)

These assets are ones that previously had a carrierwave uploader in Whitehall but that has subsequently been removed. The files themselves are still publically accessible, so need to be migrated. The list of paths and the number of assets is as follows:

```
| path                                                    | count | total size |
|---------------------------------------------------------+-------+------------|
| `system/uploads/attachment/file`                        |  6882 | 1.8 GiB    |
| `system/uploads/consultation_response_form/file`        |    19 | 4.3 MiB    |
| `system/uploads/edition_organisation_image_data/file`   | 11402 | 1.4 GiB    |
| `system/uploads/edition_world_location_image_data/file` |  5243 | 507.4 MiB  |
| `system/uploads/news_article/featuring_image`           |    10 | 4.1 MiB    |
| `system/uploads/news_article/image`                     |   112 | 31.5 MiB   |
| `system/uploads/topical_event/logo`                     |   490 | 44.7 MiB   |
```

There are a list of example assets with URLs in https://gist.github.com/chrisroos/ae70ae2bbc8a5e0e06e98b7aea08d233#example-whitehall-image-assets-on-integration.

We'll need to do this before we can remove the `PublicUploadsController` from Whitehall (see issue #213).

## Tasks

- [x] run `asset_manager:migrate_assets[system/uploads/attachment/file]` on whitehall in production
- [x] run `asset_manager:migrate_assets[system/uploads/consultation_response_form/file]` on whitehall in production
- [x] run `asset_manager:migrate_assets[system/uploads/edition_organisation_image_data/file]` on whitehall in production
- [x] run `asset_manager:migrate_assets[system/uploads/edition_world_location_image_data/file]` on whitehall in production
- [x] run `asset_manager:migrate_assets[system/uploads/news_article/featuring_image]` on whitehall in production
- [x] run `asset_manager:migrate_assets[system/uploads/news_article/image]` on whitehall in production
- [x] run `asset_manager:migrate_assets[system/uploads/topical_event/logo]` on whitehall in production
[...]

Issue #215, opened 2017-09-20
Migrate all Whitehall ImageUploader-backed assets to Asset Manager


chrisroos:
From #341 these are the remaining assets that need to be migrated as part of this task.

| Model                                  | Uploader                         | path                                                       | count   | total size |
|----------------------------------------|----------------------------------|------------------------------------------------------------|---------|------------|
| ClassificationFeaturingImageData\#file | ImageUploader                    | `system/uploads/classification_featuring_image_data/file`  | 15498   | 2.0 GiB    |
| DefaultNewsOrganisationImageData\#file | ImageUploader                    | `system/uploads/default_news_organisation_image_data/file` | 1862    | 264.7 MiB  |
| Feature\#image                         | ImageUploader                    | `system/uploads/feature/image`                             | 473781  | 55.1 GiB   |
| ImageData\#file                        | ImageUploader                    | `system/uploads/image_data/file`                           | 494970  | 57.5 GiB   |
| Person\#image                          | ImageUploader                    | `system/uploads/person/image`                              | 23927   | 2.0 GiB    |
| PromotionalFeatureItem\#image          | ImageUploader                    | `system/uploads/promotional_feature_item/image`            | 371     | 12.8 MiB   |
| TakePartPage\#image                    | ImageUploader                    | `system/uploads/take_part_page/image`                      | 147     | 18.2 MiB   |

Note that `Classification#logo` also [mounts the ImageUploader](https://github.com/alphagov/whitehall/blob/e42d4bc787828a3da77de89891a8c44e04c6cfd5/app/models/classification.rb#L61) but there are no associated files stored on NFS.

- [x] Run `asset_manager:migrate_assets[system/uploads/promotional_feature_item/image]` on production
- [x] Run `asset_manager:migrate_assets[system/uploads/take_part_page/image]` on production
- [x] Run `asset_manager:migrate_assets[system/uploads/default_news_organisation_image_data/file]` on production
- [x] Run `asset_manager:migrate_assets[system/uploads/classification_featuring_image_data/file]` on production
- [x] Run `asset_manager:migrate_assets[system/uploads/person/image]` on production
- [x] Run `asset_manager:migrate_assets[system/uploads/feature/image]` on production

floehopper:
This work feels quite dependent on #212, so it's probably worth holding off until we've made some progress on that.

floehopper:
[...]

PR #242, opened 2017-09-28
Allow etag and last_modified to be specified when creating Whitehall assets via API

floehopper:
Add optional `legacy_etag` & `legacy_last_modified` parameters for the API endpoint which allows creation of Whitehall assets. We're going to need these when we move existing Whitehall assets to Asset Manager in order to avoid wholesale cache invalidation. See #234 for more details.

I've chosen to use `legacy_xxx` vs just `xxx` as the parameter names, because once we've moved the existing Whitehall assets to Asset Manager they will become redundant. Hopefully the names will also discourage anyone from using the parameters in any of the other publishing apps.

I've aliased `WhitehallAsset#legacy_xxx=` to `Asset#xxx=` partly in recognition that we'll only ever need to set `Asset#xxx` via the API when we move existing Whitehall assets to Asset Manager and partly to keep the code as simple as possible. I think it's correct that the canonical reader method is still just `Asset#xxx` vs `Asset#legacy_xxx`, because that code is used more generally within Asset Manager.

I think there is a case for preventing calls to `Asset#etag=` or `Asset#last_modified=` except from within `Asset#store_metadata` which is called after creation of an Asset, but I'm not sure it's that important so I haven't done anything about it for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ See the `AssetPresenter` class for the return format for the above API calls. Al
 
 `GET /media/:id/:filename` serves the file to the user if it is marked as clean.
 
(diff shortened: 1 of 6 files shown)


floehopper:
I've opened alphagov/gds-api-adapters#760 to document these new optional attributes.

floehopper:
I've added notes to the descriptions for #215 & #216 to remind us to set `legacy_etag` and `legacy_last_modified` when migrating existing Whitehall assets to Asset Manager in order to avoid wholesale cache invalidation.

PR #243, opened 2017-09-29
Ensure S3Storage::Null object implements all public methods of S3Storage

floehopper:
The `S3Storage::Null` class had got out of sync with `S3Storage`. I've fixed this and added an example to ensure it doesn't happen again. I've also extracted the exception message into a constant to avoid duplication.


floehopper:
Rebased against master and force-pushed in preparation for merging.

Issue #245, opened 2017-10-02
Change Whitehall so that it can replace non-attachment assets in Asset Manager


chrislo:
Arises from: #212 

Add the ability to replace images in Asset Manager when they're replaced on the local file system. 

We've investigated replacing assets in asset manager in #229. There may be some changes needed in asset manager to support this - but we'll need to explore what Whitehall actually needs to do before changing asset manager.

floehopper:
When we work on this, we should consider what should happen to any values stored in `Whitehall#legacy_etag` and `#legacy_last_modified` in Asset Manager when an asset is updated. See [this comment](https://github.com/alphagov/asset-manager/pull/257#issuecomment-335421556) for more details.

floehopper:
We discussed this just now and agreed that this work is somewhat dependent on #244, so I'm going to mark it as blocked.

chrisroos:
I've removed the blocked label now that #244 has been closed.

chrislo:
I think this issue can be addressed by making a `POST` to `/whitehall_assets` with an existing `legacy_url_path` replace the existing asset. #336 makes that change.

chrislo:
https://github.com/alphagov/whitehall/pull/3602 uses the modified `/whitehall_assets` `POST` endpoint to soft-delete existing assets when they are replaced by ones with the same filename (in the case of Person images in Whitehall for example). Replacing an image with one with a different filename is also addressed in https://github.com/alphagov/whitehall/pull/3602. 

I think this issue can be closed.

PR #257, opened 2017-10-09
Fix asset attributes generated from file metadata

floehopper:
This was prompted by the concern I mentioned in [this comment][1]:

> If the attributes on an `Asset` which are derived from the file metadata are set in the database and then the `Asset` is updated with a new file, the attributes will not be updated as you might expect.

In doing this work, I also realised that although the attribute values derived from file metadata were being set, they were not being persisted! This was because I had used an `after_create` callback vs `before_create` and it is fixed in the first commit (with title "Save asset attribute values generated from file metadata").

In some ways, the fact that no values were being persisted makes things simpler for us in deciding whether to merge this PR, because we know that no assets will have `etag`, `last_modified` or `md5_hexdigest` values set in integration, staging or production.

In the second commit (with title "Store WhitehallAsset#legacy_etag & #legacy_last_modified in database"), I've promoted `WhitehallAsset#legacy_etag` and `WhitehallAsset#legacy_last_modified` to be proper fields on `WhitehallAsset`. This simplifies the logic in `Asset` and separates the concerns of (a) caching the file metadata-derived values in the database; and (b) overriding the attribute values when migrating existing Whitehall assets to Asset Manager.

The third commit (with title "Regenerate attributes from file metadata on asset update") I fixed the problem which triggered this work (see above).

And in the final commit (with title "Add spec for AssetFileMetadataWorker used by Rake task"), I've added some test coverage for the worker used by the Rake task to populate the attributes for existing Asset Manager assets.

[1]: https://github.com/alphagov/asset-manager/issues/182#issuecomment-334786966

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -17,8 +17,13 @@ class Asset
   field :organisation_slug, type: String
 
   field :etag, type: String
+  protected :etag=
+
   field :last_modified, type: Time
+  protected :last_modified=
+
   field :md5_hexdigest, type: String
+  protected :md5_hexdigest=
 
   validates :file, presence: true
   validates :organisation_slug, presence: true, if: :access_limited?
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -119,9 +119,9 @@ class Asset
 protected
 
   def store_metadata
-    self[:etag] = etag_from_file
-    self[:last_modified] = last_modified_from_file
-    self[:md5_hexdigest] = md5_hexdigest_from_file
+    self.etag = etag_from_file
+    self.last_modified = last_modified_from_file
+    self.md5_hexdigest = md5_hexdigest_from_file
   end
 
   def valid_filenames
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -32,7 +32,7 @@ class Asset
 
   mount_uploader :file, AssetUploader
 
-  after_create :store_metadata
+  before_create :store_metadata
   after_save :schedule_virus_scan
 
   state_machine :state, initial: :unscanned do
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -32,7 +32,7 @@ class Asset
 
   mount_uploader :file, AssetUploader
 
-  before_create :store_metadata
+  before_save :store_metadata
   after_save :schedule_virus_scan
 
   state_machine :state, initial: :unscanned do
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -119,9 +119,9 @@ class Asset
 protected
 
   def store_metadata
-    self[:etag] ||= etag_from_file
-    self[:last_modified] ||= last_modified_from_file
-    self[:md5_hexdigest] ||= md5_hexdigest_from_file
+    self[:etag] = etag_from_file
+    self[:last_modified] = last_modified_from_file
+    self[:md5_hexdigest] = md5_hexdigest_from_file
   end
 
   def valid_filenames
--- a/spec/models/asset_spec.rb
+++ b/spec/models/asset_spec.rb
@@ -330,15 +330,15 @@ RSpec.describe Asset, type: :model do
       allow(asset).to receive(:etag_from_file).and_return('etag-from-file')
     end
 
-    context "when etag attribute is explicitly set" do
-      let(:etag) { 'etag-attribute-value' }
+    context "when etag attribute is set" do
+      let(:etag) { 'etag-value' }
 
-      it "returns attribute value" do
-        expect(asset.etag).to eq('etag-attribute-value')
+      it "returns etag attribute value" do
+        expect(asset.etag).to eq('etag-value')
       end
     end
 
-    context "when etag attribute is not explicitly set" do
+    context "when etag attribute is not set" do
       let(:etag) { nil }
 
       it "returns value generated from file metadata" do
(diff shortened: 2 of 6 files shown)


floehopper:
I've just realised the `legacy_etag` and `legacy_last_modified` should probably be ignored if the asset is updated. Closing in order to address this issue.

floehopper:
It turns out it would be quite a bit of work to detect whether the file has _changed_ on an update in order to reset the `legacy_etag` and `legacy_last_modified` attributes in a sensible way and it would add quite a bit of complexity. Since nobody should be creating, let alone updating, Whitehall assets in production at the moment, I think it's OK to leave the functionality as it stands, i.e. the `legacy_etag` and `legacy_last_modified` will always take priority if they are set, even if an asset is updated with a new file.

We will naturally revisit this when we work on [Change Whitehall so that it can replace non-attachment assets in Asset Manager](https://github.com/alphagov/asset-manager/issues/245) at which point it should become clearer how we want this to work.

floehopper:
I've just confirmed that there are no WhitehallAssets in integration nor are there any Assets which have any of `etag`, `last_modified` or `md5_hexdigest` set to anything other than `nil`:

```
> Asset.count
=> 61119
> WhitehallAsset.count
=> 0
> Asset.where(:etag.ne => nil).count
=> 0
> Asset.where(:last_modified.ne => nil).count
=> 0
> Asset.where(:md5_hexdigest.ne => nil).count
=> 0
```

I _think_ this means that, at least at the point where the integration database was last copied from production, my assertion in the PR description was correct.

spec/models/asset_spec.rb
```diff
@@ -330,28 +330,43 @@
...
+
+    context "when asset is created" do
+      let(:etag) { nil }
+
+      before do
+        asset.save!
+      end
+
+      it "stores the value generated from the file in the database" do
+        expect(asset.reload[:etag]).to eq('etag-from-file')
```
chrisroos: I found it slightly confusing that this is using `#[]` and passing `:etag` instead of calling `#etag` directly (and similarly when calling `#[]` with `:last_modified` and `:md5_hexdigest` elsewhere). Can we call the methods instead?

app/models/asset.rb
```diff
@@ -119,9 +119,9 @@ def accessible_by?(user)
 protected
 
   def store_metadata
-    self[:etag] ||= etag_from_file
-    self[:last_modified] ||= last_modified_from_file
-    self[:md5_hexdigest] ||= md5_hexdigest_from_file
+    self[:etag] = etag_from_file
+    self[:last_modified] = last_modified_from_file
+    self[:md5_hexdigest] = md5_hexdigest_from_file
```
chrisroos: Is it important to use, for example, `self#[:md5_hexdigest]` instead of `self.md5_hexdigest=` in here?

chrisroos:
The last two commits (titled "Regenerate attributes from file metadata on asset update" and "Add spec for AssetFileMetadataWorker used by Rake task") don't contain any additional detail. I'd find a bit of an explanation useful in each case.

chrisroos:
I realise they weren't added in this commit but did you consider making it harder to set `Asset#etag`, `Asset#last_modified` and `Asset#md5_hexdigest` from outside the model?

spec/models/asset_spec.rb
floehopper: I've addressed this via an interactive rebase in a couple of commits where calls to the `Asset#[]` method were introduced. I plan to force push once I've addressed the other issues.

app/models/asset.rb
floehopper: No, I don't think so. I've addressed this in a new commit, because this code is already in `master`.

floehopper:
> The last two commits (titled "Regenerate attributes from file metadata on asset update" and "Add spec for AssetFileMetadataWorker used by Rake task") don't contain any additional detail in the commit note. I'd find a bit of an explanation useful in each case.

I've expanded the commit notes to give a better explanation of these changes.

floehopper:
I've just force-pushed my changes based on the feedback from @chrisroos.

floehopper:
These changes have been deployed to integration and I've just [run the `govuk_assets:store_values_generated_from_file_metadata` Rake task on integration][1].

I monitored the Sidekiq queues while this was running observed the number of jobs increasing quite quickly up to ~50K (while the jobs were being queued) and then decreasing at a slower rate back down to 0 (as the jobs were processed). The whole process took about 9 mins.

2017-10-10 asset manager - rake task - grafana

I used the Rails console to verify that the attributes had all been set for all the assets:

```
> Asset.count
=> 61119
> Asset.where(:etag => nil).count
=> 0
> Asset.where(etag: nil).count
=> 0
> Asset.where(last_modified: nil).count
=> 0
> Asset.where(md5_hexdigest: nil).count
=> 0
```

I plan to get these changes deployed to staging & production tomorrow and have someone run the Rake task in both environments too.

[1]: https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/3206/console

floehopper:
I got someone from the 2nd Line team to deploy this to staging and production this morning and to run the Rake task.

In each environment, I added an "attachment" to a draft Manual section and used the preview functionality to obtain a URL for the asset. I then used `curl` to check the values of the `ETag` & `Last-Modified` before and after the change to ensure that they remained the same.

## Staging

All 61231 assets were updated.

Test asset: https://assets-origin.staging.publishing.service.gov.uk/media/59ddea4940f0b67f446838dc/blue-960640-1.gif

### Before change

```
Last-Modified: Wed, 11 Oct 2017 09:54:17 GMT
ETag: "59ddea49-52b"
```

### After change

```
Last-Modified: Wed, 11 Oct 2017 09:54:17 GMT
ETag: "59ddea49-52b"
```

## Production

All 61244 assets were updated.

Test asset: https://assets-origin.publishing.service.gov.uk/media/59ddeb2ced915d6aaafc2e7e/blue-960x640-1.gif

### Before change

```
Last-Modified: Wed, 11 Oct 2017 09:58:04 GMT
ETag: "59ddeb2c-52b"
```

### After change

```
Last-Modified: Wed, 11 Oct 2017 09:58:04 GMT
ETag: "59ddeb2c-52b"
```

PR #261, opened 2017-10-13
Allow Whitehall assets to be proxied to S3

floehopper:
Conditionally proxy requests for Whitehall assets to S3 using the same logic we use for deciding whether to proxy requests for Mainstream assets to S3 or serve them from NFS.

In order to implement this without introducing duplication, I've introduced a common base class for the two media controllers. Usually I would prefer delegation vs inheritance, but in this case, since the longer term plan is to bring the serving of Mainstream & Whitehall assets into line with each other (and ideally we'd end up with a single media controller), I think inheritance is an acceptable approach.

Note that at the time of making this change the `PROXY_PERCENTAGE_OF_ASSET_REQUESTS_TO_S3_VIA_NGINX` environment variable is set to `100` in production, but `0` in staging and integration. If that continues to be the case, it means that *all* Whitehall asset requests will be proxied to S3 in production as soon as these changes are merged & deployed.

I can imagine it *might* be useful to be able to control the percentage for Whitehall assets separately from that for Mainstream assets. @chrisroos & @chrislo: what do you think?


chrislo:
> I can imagine it might be useful to be able to control the percentage for Whitehall assets separately from that for Mainstream assets. @chrisroos & @chrislo: what do you think?

We discussed this in person this morning and agreed that having the ability to control the percentage for Whitehall assets independently would be useful at this stage. We recognised that these options will go away when we remove the dependence on NFS (milestone/5)

app/controllers/media_controller.rb
```diff
@@ -1,5 +1,4 @@
 class MediaController < BaseMediaController
-  skip_before_filter :require_signin_permission!
```
chrislo: Can the corresponding specs in `media_controller_spec.rb:265` be removed here too?

floehopper: Although superficially, [the example you mention](https://github.com/alphagov/asset-manager/blob/0c43e01220f272f2887c93d3880c989bb007c189/spec/controllers/media_controller_spec.rb#L265-L269) *looks* as if it is related to the `skip_before_filter`, it is in fact testing [the behaviour provided by `MediaController#authenticate_if_private](https://github.com/alphagov/asset-manager/blob/0c43e01220f272f2887c93d3880c989bb007c189/app/controllers/media_controller.rb#L58-L60) which is added by a `before_filter` in `MediaController`.

As far as I could see, although there are examples which rely on the skipping of the before filter, there are none that directly test for its existence. Thus I don't think there are any examples which can be removed.

PR #262, opened 2017-10-13
Fix brittle AssetFileMetadataWorker spec

floehopper:
On Jenkins CI builds (but not local MacOS or development VM builds), the example for `Asser#last_modified` was slightly brittle for a couple of reasons:

* The memo-izing of `Asset#file_stat` seemed to cause odd results.
* The rounding to an integer occasionally went the wrong way.

I'm not 100% I understand what was going on, but I'm pretty confident that this fixes the problem!

### DIFF
--- a/spec/workers/asset_file_metadata_worker_spec.rb
+++ b/spec/workers/asset_file_metadata_worker_spec.rb
@@ -18,7 +18,7 @@ RSpec.describe AssetFileMetadataWorker, type: :worker do
 
     worker.perform(asset.id.to_s)
 
-    expect(asset.reload.last_modified.to_i).to eq(asset.last_modified_from_file.to_i)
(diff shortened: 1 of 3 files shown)


spec/workers/asset_file_metadata_worker_spec.rb
```diff
@@ -18,7 +18,7 @@
 
     worker.perform(asset.id.to_s)
 
-    expect(asset.reload.last_modified.to_i).to eq(asset.last_modified_from_file.to_i)
+    expect(asset.reload.last_modified).to be_within(0.001).of(asset.last_modified_from_file)
```
chrislo: I might be tempted to add a custom `be_within_a_millisecond` matcher here to make it easier to understand the `0.001` magic number.

PR #264, opened 2017-10-17
Add Whitehall asset endpoint

chrisroos:
This endpoint allows us to find a Whitehall asset using its `legacy_url_path`. We're going to use this in Whitehall to update/delete assets stored in Asset Manager.

### DIFF
--- a/app/models/whitehall_asset.rb
+++ b/app/models/whitehall_asset.rb
@@ -7,7 +7,9 @@ class WhitehallAsset < Asset
 
   validates :legacy_url_path,
     presence: true,
-    uniqueness: true,
+    uniqueness: {
+      conditions: -> { where(deleted_at: nil) }
+    },
     format: {
       with: %r{\A/government/uploads},
       message: 'must start with /government/uploads'
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,6 +6,7 @@ Rails.application.routes.draw do
   end
 
   resources :whitehall_assets, only: %i(create)
+  get '/whitehall_assets/*path' => 'whitehall_assets#show'
 
   get "/media/:id/:filename" => "media#download", :constraints => { filename: /.*/ }
   get "/government/uploads/*path" => "whitehall_media#download"
--- a/spec/controllers/whitehall_assets_controller_spec.rb
+++ b/spec/controllers/whitehall_assets_controller_spec.rb
@@ -95,4 +95,46 @@ RSpec.describe WhitehallAssetsController, type: :controller do
       end
     end
   end
+
+  describe 'GET show' do
+    let(:asset) { FactoryGirl.create(:whitehall_asset, legacy_url_path: '/government/uploads/image.png') }
+    let(:presenter) { instance_double(AssetPresenter) }
+
+    before do
+      allow(AssetPresenter).to receive(:new).with(asset, anything).and_return(presenter)
+      allow(presenter).to receive(:as_json).and_return('asset-as-json')
+    end
+
+    it 'returns a 200 response' do
+      get :show, path: 'government/uploads/image', format: 'png'
+
+      expect(response).to have_http_status(:ok)
+    end
+
+    it 'returns a 404 response if the asset cannot be found' do
+      get :show, path: 'government/uploads/non-existent-image', format: 'png'
+
+      expect(response).to have_http_status(:not_found)
+    end
+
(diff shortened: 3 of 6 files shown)


spec/models/whitehall_asset_spec.rb
```diff
@@ -44,6 +44,17 @@
           expect(asset).not_to be_valid
           expect(asset.errors[:legacy_url_path]).to include('is already taken')
         end
+
+        context 'but the existing asset has been deleted' do
```
floehopper: Minor: I think it might be clearer to say "marked as deleted" in the example name, rather than just "deleted".

spec/models/whitehall_asset_spec.rb
```diff
@@ -44,6 +44,17 @@
...
+          before do
+            asset.legacy_url_path = existing_asset.legacy_url_path
+            existing_asset.delete
+          end
+
+          it 'is valid' do
```
floehopper: Minor: It might be worth using the example name to explain things further, e.g. `it 'is valid because legacy_url_path is unique within the assets not marked as deleted' do`.

app/models/whitehall_asset.rb
```diff
@@ -7,7 +7,9 @@ class WhitehallAsset < Asset
 
   validates :legacy_url_path,
     presence: true,
-    uniqueness: true,
+    uniqueness: {
+      conditions: -> { where(deleted_at: nil) }
+    },
```
floehopper: Very minor: Why does this need to be split onto multiple lines?

spec/controllers/whitehall_assets_controller_spec.rb
```diff
@@ -95,4 +95,46 @@
...
+    end
+
+    it 'sets the cache expiry to 30 minutes if the asset is scanned' do
+      asset.scanned_clean!
+
+      get :show, path: 'government/uploads/image', format: 'png'
+
+      expect(response.headers['Cache-Control']).to match("max-age=#{30.minutes}")
+    end
+  end
```
floehopper: It looks as if you've duplicated behaviour from `AssetsController#show`, e.g. the `Cache-Control` header logic. If my assumption is correct, it would be useful to explain that you've done this in the commit note.

floehopper: Minor: Unlike the `AssetsController` spec, it looks as if you've chosen to stub the `AssetPresenter` behaviour. I think it might be worth noting this in the commit note.

config/routes.rb
```diff
@@ -6,6 +6,7 @@
   end
 
   resources :whitehall_assets, only: %i(create)
+  get '/whitehall_assets/*path' => 'whitehall_assets#show'
```
floehopper: Did you consider using the standard resource-based route for `show`? If you did, it would be useful to explain why you didn't use it in the commit note. I have to admit it seems a bit odd to me, but I'm probably missing something!

app/models/whitehall_asset.rb
chrisroos: We've used multiple lines for the `format` options so I did the same here.

config/routes.rb
chrisroos: I've added the following to the commit message to explain this:

> Requests look something like '/whitehall_assets/government/uploads/asset.png'. I can't use the standard Rails restful route for the show action because that doesn't appear to work for `:id` parameters that include slashes.

spec/controllers/whitehall_assets_controller_spec.rb
chrisroos: I've added the following the commit note to address these two points:

> The implementation of `WhitehallAssets#show` is identical to `Assets#show` in all but the way we find the asset.
>  
> I've chosen to assert that we're using the `AssetPresenter` to render the asset (versus testing the actual rendered output) as the behaviour of `AssetPresenter` is being unit-tested elsewhere.

spec/models/whitehall_asset_spec.rb
chrisroos: I've updated this and force pushed.

spec/models/whitehall_asset_spec.rb
chrisroos: I've updated this and force-pushed.

chrisroos:
Thanks for reviewing, @floehopper. I've addressed your feedback, rebased and force pushed in preparation for merging.

PR #265, opened 2017-10-18
Simplify Asset attributes obtained from file metadata

floehopper:
Now that the `govuk_assets:store_values_generated_from_file_metadata` Rake task has been run in all environments, we can rely on the presence of the value in the database which is set in a `before_save` callback by `Asset#store_metadata`.

I did consider adding validation for the presence of these attributes, but it turned out to be very awkward and in any case, I soon realised that it's a bit odd to validate attributes which are set automatically by a `before_save` callback, especially given that the attribute writer methods are protected. So I decided that adding such validation was unnecessary.

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -91,24 +91,12 @@ class Asset
     %w(jpg jpeg png gif).include?(extension)
   end
 
-  def etag
-    self[:etag] || etag_from_file
-  end
-
   def etag_from_file
     '%x-%x' % [last_modified_from_file, file_stat.size]
   end
 
-  def last_modified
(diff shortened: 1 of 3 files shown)


floehopper:
@chrislo:

> In the commit message for 060e1ab you say "I'm about to remove these fallbacks" but it's the last commit in the PR - was that intentional? I guess it refers to fd53f58? Github playing games with the ordering of rebased commits?

Your assumption is correct.

floehopper:

floehopper:
Rebasing again and force-pushing in preparation for merging.

PR #266, opened 2017-10-18
Improvements to Asset spec

floehopper:
This makes some of the examples in Asset spec a bit more idiomatic from an RSpec and FactoryGirl point of view. My motivation for making the changes was that I was planning to make some changes to Asset validation. I didn't end up making those changes, but I think these improvements are still worth making.


floehopper:

PR #267, opened 2017-10-19
Fake S3 in development environments

### LINKED ISSUE #155: Work out how asset-manager should work in development when uploading to S3
We'll need to consider how asset-manager will work in development once we're uploading to S3 by default. I think we'll need to do this before we do PR #154.

floehopper:
This introduces a new `S3Storage::Fake` class and configures the app to use an instance of this class in development environments if S3 is not configured.

We want to remove support for serving asset requests from NFS and proxy all of them to S3. However, having to set up the necessary S3 resources in a development environment seems rather onerous. The changes in this PR mean that by default development environments will use a *fake* version of `S3Storage` which stores the files in the local filesystem and serves them via a `Rack::File` instance mounted in the Rails app.

I did consider a couple of other options:

* The [fake-s3][3] gem allows you to run a separate server process which acts like S3 but is backed by the local file system. I chose not to use it for two reasons: (a) although it's open-source and free, it requires you to obtain a commercial-looking license which I thought might be problematic for GDS; and (b) running a separate process is a lot more hassle to setup and there are more things to go wrong.

* Retaining the `send_file` behaviour that we currently use to serve assets from NFS. The problem with this would've been that the code path in development would've been really quite different to that used in all other environments. This would've meant an extra maintenance overhead and an increased likelihood you'd need to setup S3 in order to diagnose problems with serving assets.

I'm happy with the solution I've chosen, because requests are genuinely proxied to a fake S3 which just happens to be served in-process by the Rails app backed by the local file system.

Note that if the `X-Accel-Mapping` header is set in a request to the `Rack::File` endpoint, this will trigger the `Rack::Sendfile` middleware in the default Rails stack to send a response with the `X-Accel-Redirect` header instead of the body of the file. This is currently a problem, because [the Nginx config for Asset Manager sets the `X-Accel-Mapping` request header whether or not the request is going to be served from NFS or proxied to S3][1]. Since the `X-Accel-Mapping` value is not set with the fake S3 paths in mind, this results in an attempt to serve a non-existent file and a 404 Not Found.

We will need to address this issue before merging this PR into master. However, since we are planning to move towards proxying all asset requests to S3, it should be possible to remove the setting of `X-Sendfile-Type` & `X-Accel-Mapping` headers shortly. I have confirmed that doing this solves the problem on my development VM.

Alternatively, if we want to merge this sooner, we could consider trying to remove these headers from the proxying request sent to S3 or fake S3, i.e. within [the "cloud-storage-proxy" location in the Nginx configuration][2]. However, I think we might've had problems trying to remove headers in this way in the past, so I'm reluctant to attempt it for now.

Closes #155.

[1]: https://github.com/alphagov/govuk-puppet/blob/1b99dbc24b0ac477fadb75ef8d1356dbfd58bf2e/modules/govuk/manifests/apps/asset_manager.pp#L91-L92
[2]: https://github.com/alphagov/govuk-puppet/blob/1b99dbc24b0ac477fadb75ef8d1356dbfd58bf2e/modules/govuk/manifests/apps/asset_manager.pp#L116-L173
[3]: https://github.com/jubos/fake-s3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@
 /uploads
 
 /spec/examples.txt
+
+/fake-s3
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -82,3 +82,5 @@ RSpec/VerifiedDoubles:
     - 'spec/support/authentication.rb'
     - 'spec/support/cloud_storage.rb'
     - 'spec/workers/save_to_cloud_storage_worker_spec.rb'
+    - 'spec/lib/s3_storage_spec.rb'
+    - 'spec/routing/fake_s3_route_spec.rb'
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -6,6 +6,10 @@
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
+Style/ConditionalAssignment:
+  Exclude:
+    - 'config/initializers/carrier_wave.rb'
+
 # Offense count: 11
 RSpec/AnyInstance:
   Exclude:
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -69,6 +69,7 @@ RSpec/NestedGroups:
 RSpec/PredicateMatcher:
   Exclude:
     - 'spec/models/asset_spec.rb'
+    - 'spec/lib/s3_storage/fake_spec.rb'
 
 # Offense count: 12
 # Configuration parameters: IgnoreSymbolicNames.
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,4 +1,4 @@
-require 'cloud_storage'
+require 's3_storage'
 
 class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
@@ -10,7 +10,7 @@ class ApplicationController < ActionController::Base
   before_filter :require_signin_permission!
 
   rescue_from Mongoid::Errors::DocumentNotFound, with: :error_404
-  rescue_from CloudStorage::NotConfiguredError, with: :error_500
+  rescue_from S3Storage::NotConfiguredError, with: :error_500
 
 private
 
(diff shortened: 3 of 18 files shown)


chrisroos:
Should the commit titled "Move S3Storage examples into their own spec & file", be "Move S3Storage::__Null__ examples into their own spec & file"?

chrisroos:
I _think_ we can avoid the problem that's blocking this being merged (i.e. the setting of the `X-Sendfile-Type` and `X-Accel-Mapping`) by setting them in a new `location` block in the nginx config. At the moment these headers are being added to every request made to asset-manager but I think they're only actually required for requests to the `MediaController`, i.e. requests matching "/media/". Adding something like this to our nginx config seems to do the trick, in that we can continue to serve files from NFS and we can serve files from the Fake S3 endpoint added in this branch:

```
location /media/ {
  proxy_set_header X-Sendfile-Type X-Accel-Redirect;
  proxy_set_header X-Accel-Mapping /var/apps/asset-manager/uploads/assets/=/raw/;
  proxy_pass http://asset-manager.dev.gov.uk-proxy;
}
```

I wonder whether this would be preferable to waiting until we're proxying all asset requests to S3.

floehopper:
@chrisroos:

Thanks for reviewing.

> Should the commit titled "Move S3Storage examples into their own spec & file", be "Move S3Storage::Null examples into their own spec & file"?

Yes. I'll fix that up now. Well spotted.

> I wonder whether this would be preferable to waiting until we're proxying all asset requests to S3.

Yes, that sounds great. I'll try to put together a PR for `govuk-puppet` to do that.

floehopper:
I've fixed up the commit title mentioned by @chrisroos and am about to force-push.

floehopper:
We should be able to merge this once alphagov/govuk-puppet#6683 is merged and deployed.

Note that there are a few small issues with the response headers returned from asset requests proxied to fake S3 URLs, e.g. the Rails middleware stack adds a `Cache-Control: no-cache` to such responses, because no `ETag` header is set. However, there isn't a straightforward way to fix this and the most important things is that the behaviour works in development (locally and on the development VM), which it does.

floehopper:
This was waiting on the fix in https://github.com/alphagov/govuk-puppet/pull/6689. However, that has now been deployed and applied in staging & production.

I'm still planning to hold back on merging this until we've deployed the pending changes in the master of this repo to staging and production, because there are quiet a few changes in there already.

floehopper:
Asset Manager has just been deployed to staging & production and so I'm happy that it's OK to merge this now.

PR #286, opened 2017-11-02
Trigger replication for all non-replicated assets on S3

floehopper:
S3 objects for some assets were added to the production S3 bucket *before* replication was enabled on the bucket. Enabling replication does not automatically replicate these S3 objects, so we need to do it manually.

The simplest way to achieve this seems to be to modify the metadata on the relevant S3 objects. This feels better than manually syncing/copying the relevant S3 objects to the backup bucket, because it uses the same mechanism (replication) to sync to the target bucket using all the same users, roles & permissions.

One possible downside of the approach is that the `Last-Modified` timestamp is updated when the metadata on an S3 object is modified. However, at the moment, we're not making use of this metadata; the Rails app sets a `Last-Modified` response header using a value stored in the database.

Another possible downside is that the new metadata entry will be sent as a response header when the S3 object is requested. However, the worker immediately removes the new metadata entry, so it will only ever be a very transient issue and in any case, the extra response header shouldn't cause any problems.

This PR introduces a new Rake task, `govuk_assets:trigger_replication_for_non_replicated_assets`, which iterates through all the assets and (for all those that have a corresponding S3 object which has never been replicated) updates the metadata on the corresponding S3 object in order to trigger replication for that object. Since doing this involves a number of requests to S3 for each asset, the Rake task queues up a Sidekiq job for each asset on a low priority queue to be processed asynchronously.

Note that at the moment, cross-region replication is only setup on the production S3 bucket and so this Rake task only needs to be run in production, not integration or staging. Having said that, running it in the other environments shouldn't cause any problems.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -69,6 +69,7 @@ RSpec/NestedGroups:
 RSpec/PredicateMatcher:
   Exclude:
     - 'spec/models/asset_spec.rb'
+    - 'spec/lib/s3_storage_spec.rb'
     - 'spec/lib/s3_storage/fake_spec.rb'
 
 # Offense count: 12
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -85,3 +85,7 @@ RSpec/VerifiedDoubles:
     - 'spec/workers/save_to_cloud_storage_worker_spec.rb'
     - 'spec/lib/s3_storage_spec.rb'
     - 'spec/routing/fake_s3_route_spec.rb'
+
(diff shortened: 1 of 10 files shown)


floehopper:
In doing some final testing I noticed that although updating the metadata on an S3 object does trigger a corresponding S3 object to be created in the replicated bucket, I’ve just noticed that the object in the replicated bucket is 0 bytes in size. I'm guessing this is because it's just "replaying" the API calls against the replicated bucket. I'm going to close this PR and come up with a better plan.

PR #287, opened 2017-11-02
Improve consistency of methods in the S3Storage class

floehopper:
These are some tidying-up commits from #286 which I still think are generally useful. Also I'm going to need the `S3Storage#exists?` method which this PR introduces when I implement the new version of #286.


floehopper:
@chrisroos These commits are unchanged since you reviewed them in #286.

PR #298, opened 2017-11-14
Upgrade to Rails 5.1.4

danielburnley:
Main changes required revolved around the changes to how params are
passed within tests when calling get. Rather than passing key/values
when calling get, you now need to pass them within a params hash.

In addition to this, the use of head to render pages using only headers
rather than render nothing: true, and before_filter being changed to
before_action

The only logic change was required in base_media_controller as if the
response had a body, e.g. if it was a text file, it was expected to
render that file. Otherwise it was expected to return that file

Testing around ensuring that mongo attributes were readonly were changed
as trying to change these attributes now throws an exception

Finally, we upgraded anything that was required to be upgraded to be
compatible with Rails 5


Gemfile
```diff
@@ -2,18 +2,19 @@ source 'https://rubygems.org'
...
+gem 'mongoid', '~> 6.2.0'
+gem 'mongoid_paranoia', '0.3.0'
 gem 'nokogiri', '1.6.6.4'
 gem 'plek', '~> 2.0'
-gem 'rack_strip_client_ip', '0.0.1'
-gem 'rails', '4.2.10'
-gem 'state_machines-mongoid', '~> 0.1.1'
+gem 'rack_strip_client_ip', '0.0.2'
+gem 'rails', '5.1.4'
+gem 'rails-controller-testing'
```
chrislo: Why has this gem been added? Could it be explained in the commit message, or ideally, introduced as a separate commit?

Gemfile
```diff
@@ -2,18 +2,19 @@ source 'https://rubygems.org'
 
 gem 'aws-sdk'
 gem 'carrierwave', '~> 0.10.0'
-gem 'carrierwave-mongoid', '~> 0.8.1', require: 'carrierwave/mongoid'
-gem 'gds-sso', '~> 11.2'
+gem 'carrierwave-mongoid', '~> 0.10.0', require: 'carrierwave/mongoid'
+gem 'gds-sso', '~> 13.2'
```
chrislo: Could this gem upgrade be made as a separate commit before the upgrade to Rails?

app/controllers/application_controller.rb
```diff
@@ -7,7 +7,7 @@ class ApplicationController < ActionController::Base
 
   include GDS::SSO::ControllerMethods
 
-  before_filter :require_signin_permission!
+  before_action :require_signin_permission!
```
chrislo: I think `before_action` is available in Rails 4 (and deprecated in 5.1). If possible, could this change be made in a separate commit before the Rails upgrade?

app/controllers/base_media_controller.rb
```diff
@@ -1,5 +1,5 @@
 class BaseMediaController < ApplicationController
-  skip_before_filter :require_signin_permission!
+  skip_before_action :require_signin_permission!
```
chrislo: I think this change could happen before the upgrade to 5.

spec/models/asset_spec.rb
```diff
@@ -61,9 +61,7 @@
     it 'cannot be changed after creation' do
       uuid = '11111111-1111-1111-1111-11111111111111'
       asset = FactoryGirl.create(:asset, uuid: uuid)
-      asset.uuid = '22222222-2222-2222-2222-222222222222'
-      asset.save!
-      expect(asset.reload.uuid).to eq(uuid)
+      expect { asset.uuid = '22222222-2222-2222-2222-222222222222' }.to raise_error(Mongoid::Errors::ReadonlyAttribute)
```
chrislo: This looks like a change of behaviour - are you able to explain it in the commit message?

spec/models/whitehall_asset_spec.rb
```diff
@@ -75,9 +75,7 @@
 
     context 'when updating asset' do
       it 'cannot be set' do
-        asset.legacy_url_path = '/government/uploads/another-asset.png'
-        asset.save!
-        expect(asset.reload.legacy_url_path).to eq('/government/uploads/asset.png')
+        expect { asset.legacy_url_path = '/government/uploads/another-asset.png' }.to raise_error(Mongoid::Errors::ReadonlyAttribute)
```
chrislo: This also looks like a change in behaviour. Is it related to upgrading Mongo?

danielburnley:
I'll work on splitting these out into separate PRs where possible and explain the changes in those commits 👍

spec/models/whitehall_asset_spec.rb
danielburnley: Yeah - I've added an explanation to the commit message as to why these tests have changed. Mongoid now throws exceptions when trying to update a readonly attribute so the tests needed to check that the exception was thrown

app/controllers/media_controller.rb
```diff
@@ -1,5 +1,5 @@
 class MediaController < BaseMediaController
-  before_filter :authenticate_if_private
+  before_action :authenticate_if_private
```
floehopper: I think it would be worth explaining in the commit note that [`before_action` is just a renaming of `before_filter`](https://github.com/rails/rails/commit/9d62e04838f01f5589fa50b0baa480d60c815e2c), i.e. there is no inherent change in behaviour.

Gemfile
```diff
@@ -3,7 +3,7 @@ source 'https://rubygems.org'
 gem 'aws-sdk'
 gem 'carrierwave', '~> 0.10.0'
 gem 'carrierwave-mongoid', '~> 0.8.1', require: 'carrierwave/mongoid'
-gem 'gds-sso', '~> 11.2'
+gem 'gds-sso', '~> 13.2'
```
floehopper: The [`CHANGELOG` for `gds-sso` v13.0.0](https://github.com/alphagov/gds-sso/blob/9478045b27fd2aa9b2832c7d3df266eea43dab47/CHANGELOG.md#1300) mentions two breaking changes. Can you confirm that these do not affect this app?

spec/controllers/base_media_controller_spec.rb
```diff
@@ -3,7 +3,7 @@
 RSpec.describe BaseMediaController, type: :controller do
   controller do
     def anything
-      render nothing: true
+      head :ok
```
floehopper: I think this is related to the previous commit, isn't it?

danielburnley: You're correct! I've put it in the commit before. Missed it whilst splitting out

danielburnley:
Mongoid depends on Rails 5 and Rails 5 depends on mongoid yeah - couldn't separate them any further

floehopper:
I believe all of the comments in @chrislo's review have been addressed and so I'm going to dismiss the review as he's currently away.

PR #299, opened 2017-11-14
Respond with 204 if no content when downloading

danielburnley:
In Rails 5 - When the response has no body an error is thrown, this
change makes it so that the error will not be thrown, along with keeping
existing behaviour for files that have data inside the response body
(e.g. a text file)

done in preperation for #298

### DIFF
--- a/app/controllers/base_media_controller.rb
+++ b/app/controllers/base_media_controller.rb
@@ -25,6 +25,7 @@ protected
   end
 
   def serve_from_nfs_via_nginx(asset)
-    send_file(asset.file.path, disposition: AssetManager.content_disposition.type)
+    send_file asset.file.path, disposition: AssetManager.content_disposition.type
+    head :no_content if response.body.empty?
   end
 end


chrisroos:
Hi @danielburnley. Have you tried this change in the dev-vm, or at least with Asset Manager running behind a web server? The `send_file` method doesn't actually send the content of the file when it's running behind our nginx configuration - it'll send the location of the file in the `X-Sendfile` header instead. My concern is that in this case, `response.body.empty?` will be true and that we'll now be sending something different.

Can you also add some tests for this change, please?

chrisroos:
In case it's useful, the following commands illustrate the difference between requesting the file directly from Asset Manager vs requesting via nginx. You can see that requesting via nginx results in a `Content-Length` of 0 and an empty response body. As explained above, my concern is that the code change in this PR might not play well with the requests being made by nginx.

```
# Create a file to upload
$ echo `date` > tmp.txt

# Upload the file
$ curl http://localhost:3000/assets --form "asset[file]=@tmp.txt"

# Request the file
$ curl http://localhost:3000/media/5a0b0278edf8770dfb465870/tmp.txt
Tue 14 Nov 2017 14:49:07 GMT

# Simulate a request of the file from nginx
# Observe that the response body is empty
$ curl http://localhost:3000/media/5a0b0278edf8770dfb465870/tmp.txt \
-H"X-Accel-Mapping: `pwd`/uploads/assets/=/raw/" \
-H"X-Sendfile-Type: X-Accel-Redirect" \
-v

> GET /media/5a0b0278edf8770dfb465870/tmp.txt HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/7.43.0
> Accept: */*
> X-Accel-Mapping: /Users/chrisroos/Code/alphagov/asset-manager/uploads/assets/=/raw/
> X-Sendfile-Type: X-Accel-Redirect
>
< HTTP/1.1 200 OK
< X-Frame-Options: DENY
< X-Xss-Protection: 1; mode=block
< X-Content-Type-Options: nosniff
< Content-Disposition: inline; filename="tmp.txt"
< Content-Transfer-Encoding: binary
< Content-Type: text/plain
< Cache-Control: private
< X-Request-Id: 23c813d0-46bb-4fd5-a9b5-4f384aed7aba
< X-Runtime: 0.063183
< Content-Length: 0
< X-Accel-Redirect: /raw/0b/02/5a0b0278edf8770dfb465870/tmp.txt
< Server: WEBrick/1.3.1 (Ruby/2.3.1/2016-04-26)
< Date: Tue, 14 Nov 2017 15:10:24 GMT
< Connection: Keep-Alive
<
* Connection #0 to host localhost left intact
```

danielburnley:
Yeah you're correct - it does cause the response to return nothing. Looking into why these tests fail because it seems like it might be a problem with them, rather than the actual code

chrisroos:
Hey @danielburnley. We're hoping to get rid of the functionality for serving files from NFS in the near future (possibly in the next couple of days) so it's probably not worth spending too long investigating this particular problem.

danielburnley:
Hey @chrisroos  - I solved the issue anyway, it was a problem with the tests causing `send_file`not to return anything so it wasn't a problem with the actual code, just the tests

PR #313, opened 2017-11-22
Update Ruby to 2.4.2 and bump Gemfile dependancies

RoryMacDonald:
Update Ruby and Gemfile dependancies

Ahead of enabling Dependabot on the repository, we want to be using the latest
dependancies, to minimise the onslaught of incremental PRs.


.ruby-version
```diff
@@ -1 +1 @@
-2.3.1
+2.4.2
```
chrisroos: Can you add the newline back to the end of this file please, @RoryMacDonald?

RoryMacDonald:
@chrisroos - Thanks for your comments. Agree that bisecting will be easier if the bumps were incremental, though unsure how pragmatic this approach is across ~35 repos.

We discussed both routes with @tijmenb last week and felt the simplest path for teams across GOV.UK would be to batch update the gems to the latest available versions and then get dependabot running against a repo, which will open dependency specific PRs that the teams can deal with.

floehopper:
I don't suppose Dependabot has an option to batch multiple incremental upgrade commits into a single PR...? If it doesn't, it should have!

RoryMacDonald:
Not that i've seen, though we're hoping to be able to push some changes (like this) back into https://github.com/dependabot/dependabot-core, if we have some time.

PR #315, opened 2017-11-23
Add editorconfig

RoryMacDonald:
Introducing an editorconfig file to help with consistency across 
different editors and IDEs. See: https://github.com/alphagov/asset-manager/pull/313#discussion_r152592465

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+tab_width = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true


floehopper:
This feels like a cross-project concern - are you planning to add the same file to every GOV.UK/GDS repo? If so, we'd end up with duplication and the danger that the config for different projects would end up diverging.

RoryMacDonald:
I've seen a few PR comments about trailing white space / missing final lines, so think some standardisation might help. Take your point that having these in every repository could create divergence.

Another route that we could consider is to include a single `.default-editorconfig` within the vagrant setup and then symbolic link from each of the project repositories back to that. Though, I suspect the gradual adoption of Docker in projects may scupper this route. 

Might be worth getting thoughts from @danielroseman @tijmenb on this too

chrislo:
I think I'd rather enforce the whitespace rules in `rubocop` and leave the editor config out of the individual application repositories.

floehopper:
> I think I'd rather enforce the whitespace rules in `rubocop` and leave the editor config out of the individual application repositories. 

That might make more sense, because then the rules could go in the shared `govuk-lint` gem thus avoiding duplication and potential divergence.

kevindew:
FYI there's also this PR: https://github.com/alphagov/gds-way/pull/107 which seems to imply the copy/pasting approach and might be a good place to raise concerns regarding that.

I wonder if there's a way to tell editorconfig to just use the contents of a url that all points to master equivalent of https://github.com/alphagov/gds-way/blob/ac3cc2f8051e54b5bcef3cfa3596c7e7e385e5e4/source/manuals/programming-languages/editorconfig

floehopper:
I agree with [this comment](https://github.com/alphagov/gds-way/pull/107#issuecomment-347267143):

> My stance is that we should be using one [.editorconfig file] for all projects. Ideally only a single file in the root of a project to simplify maintenance. 

Does anyone know whether there's been any decision about this? This PR has been open for over two weeks and unless I hear otherwise I propose closing it until we have some clarity about the approach.

floehopper:
@RoryMacDonald Thanks 👍

PR #318, opened 2017-11-24
Remove redundant Rake tasks

floehopper:
This removes a bunch of one-off Rake tasks which have been run in all the relevant environments and are no longer needed.


floehopper:

Issue #171, opened 2017-08-30
Change Jenkins CI build to run non-incremental version of Rubocop


floehopper:
The Jenkins CI build did not pick up Rubocop offences fixed in #170.

I wonder whether the Jenkins CI build is still running the incremental version of `govuk-lint-ruby`...?

floehopper:
I have confirmed that the [shared `govuk_jenkinslib.groovy` script][1] uses the `--diff` and `--cached` options by default and there is currently no way to override them.

I believe this may also a problem for Manuals Publisher and Smart Answers apps - we have previously fixed all Rubocop violations in these apps.

[1]: https://github.com/alphagov/govuk-puppet/blob/ed07581c5054c8ef9183e4101571f6cd2fe5f60e/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy#L420-L422

floehopper:
Related to #307.

floehopper:
A bunch of Rubocop warnings, deprecation warnings & offences have recently been introduced - see https://github.com/alphagov/asset-manager/issues/307#issuecomment-349616006 for details.

floehopper:
This was incorrectly closed automatically by GitHub. Re-opening.

PR #332, opened 2017-12-07
Fix govuk-lint-ruby warnings and offences

floehopper:
* Fix violation of `Style/CommentedKeyword` cop
* Update `govuk-lint` from v3.3.1 to v3.4.0 to [fix deprecation warnings](https://github.com/alphagov/govuk-lint/pull/890)
* Regenerate `.rubocop_todo.yml` file

Although I know we really need to fix https://github.com/alphagov/asset-manager/issues/171 to prevent us getting into this situation again, I still think doing this is worthwhile.

Note that GitHub is displaying the two commits re-generating the `.rubocop_todo.yml` file in the wrong order - they should appear either side of the `govuk-lint` upgrade.


floehopper:
I've rebased against master, resolved the conflicts, and force-pushed.

floehopper:
@chrislo, @chrisroos: I'd appreciate a review of this when you have a mo - it's been hanging around for a while.

chrisroos:
I'm going to review this now.

PR #334, opened 2017-12-11
Allow override of fake S3 host via environment variable

floehopper:
Following on from #333, I realised I still couldn't do what I need to do in the `publishing-e2e-tests` project. So I've re-jigged things a bit and now it's possible to override the host used by the fake S3 "service" using the `FAKE_S3_HOST` environment variable.

I want to be able to do this in the `publishing-e2e-tests` project, because I want to set the fake S3 host to http://localhost and that isn't possible without this change.


floehopper:
@chrisroos Thanks. I had already mentioned my motivation in the final commit note, but I've now added an explanatory sentence to the PR description.

Issue #229, opened 2017-09-21
Change Asset Manager API to allow replacement of Whitehall assets


chrisroos:
This stems from my investigation in issue #225.

It turns out that it's possible to replace Whitehall images in a couple of instances (e.g. the images for people). I think this means we'll need to add this same functionality to the Whitehall endpoint in the Asset Manager unless we can think of an alternative approach (possibly by changing Whitehall)?

You can test the current behaviour of replacing an image in Whitehall by:

* Creating a person and uploading an image.
* Making a change to the image locally while ensuring to keep the filename the same.
* Updating the person with the new image.
* Observing that the image URL hasn't changed (i.e. it has the same `<id>` and `<filename>` in the path).

floehopper:
As far as I can see, Whitehall should be able to use the existing `PUT /assets/:id` API endpoint to update a Whitehall asset as long as it knows the Whitehall asset's ID which is returned in the response from the `POST /whitehall_assets` API endpoint. The implication here is that Whitehall will need to store the asset's ID in its database.

Note that since Whitehall specifies the public URL path at which to serve the asset, any change in filename does not affect this public URL path. I think this behaviour makes sense and is what we need in Whitehall.

@chrisroos: Does that sound correct to you? Or do you think we need to be able to update the filename in the public URL path?

The following shows how I was able to use the existing API endpoint to update a Whitehall asset:

## Create test file to upload

    $ echo `date` > tmp.txt
    $ cat tmp.txt
    Thu 21 Sep 2017 14:17:56 BST

## Create Whitehall asset to be served at /government/uploads/path/to/tmp.txt

    $ curl http://asset-manager.dev.gov.uk/whitehall_assets --form "asset[file]=@tmp.txt" --form "asset[legacy_url_path]=/government/uploads/path/to/tmp.txt"
    {"_response_info":{"status":"created"},"id":"http://asset-manager.dev.gov.uk/assets/59c3bc3c759b743503c0c66d","name":"tmp.txt","content_type":"text/plain","file_url":"http://assets-origin.dev.gov.uk/government/uploads/path/to/tmp.txt","state":"unscanned"}

## Get status of Whitehall asset (after successful virus scanning)

    $ curl http://asset-manager.dev.gov.uk/assets/59c3bc3c759b743503c0c66d
[...]

PR #336, opened 2017-12-11
POST /whitehall_assets replaces existing asset

chrislo:
Fixes: #229 

Currently in Whitehall images stored using the `ImageUploader` are
stored on NFS only. We want to change Whitehall to store these images
in Asset Manager.

In some circumstances an image using this Uploader (such as the image
associated with a Person[1]) can be replaced. For example

1. A new person is created with an image `foo.jpg`
2. The person is saved
3. The person is edited to have a different image with the same
filename `foo.jpg`
4. The person is saved again

In this sequence of events the new image will be available at the same
URL as the image added in step 1. The image added in step 1. will be
deleted from the file system.

In order to allow this behaviour in Asset Manager this commit modifies
`POST /whitehall_assets` to check whether an asset with the same
`legacy_url_path` already exists and soft-deletes it before creating
the new asset.

This will allow us to maintain the desirable existing behaviour from
Whitehall (a new image served at an existing URL) and also allows the
replaced image to be recovered if required[2].

[1] /government/admin/people/ in Whitehall Admin

### DIFF
--- a/app/controllers/whitehall_assets_controller.rb
+++ b/app/controllers/whitehall_assets_controller.rb
@@ -1,5 +1,9 @@
 class WhitehallAssetsController < ApplicationController
   def create
+    if existing_asset_with_this_legacy_url_path.exists?
+      existing_asset_with_this_legacy_url_path.destroy
+    end
+
     @asset = WhitehallAsset.new(asset_params)
 
     if @asset.save
@@ -24,5 +28,9 @@ private
     params
       .require(:asset)
       .permit(:file, :legacy_url_path, :legacy_etag, :legacy_last_modified)
+  end
+
+  def existing_asset_with_this_legacy_url_path
+    WhitehallAsset.where(legacy_url_path: asset_params[:legacy_url_path])
   end
 end
--- a/spec/controllers/whitehall_assets_controller_spec.rb
+++ b/spec/controllers/whitehall_assets_controller_spec.rb
@@ -8,9 +8,9 @@ RSpec.describe WhitehallAssetsController, type: :controller do
   end
 
   describe "POST create" do
-    context "a valid asset" do
-      let(:attributes) { FactoryBot.attributes_for(:whitehall_asset, :with_legacy_metadata) }
+    let(:attributes) { FactoryBot.attributes_for(:whitehall_asset, :with_legacy_metadata) }
 
+    context "a valid asset" do
       it "is persisted" do
         post :create, params: { asset: attributes }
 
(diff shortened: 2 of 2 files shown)


app/controllers/whitehall_assets_controller.rb
```diff
@@ -25,4 +29,8 @@ def asset_params
       .require(:asset)
       .permit(:file, :legacy_url_path, :legacy_etag, :legacy_last_modified)
   end
+
+  def existing_asset_with_this_legacy_url_path
```
floehopper: Minor: I'd be inclined to extract this method and the call to `exists?` in the `create` action into a class method on `WhitehallAsset`, e.g. `WhitehallAsset.exists_with_legacy_url_path?(path)`, and then call `if WhitehallAsset.exists_with_legacy_url_path?(asset_params[:legacy_url_path])` from the action.

spec/controllers/whitehall_assets_controller_spec.rb
```diff
@@ -94,6 +94,17 @@
         expect(response).to have_http_status(:unprocessable_entity)
       end
     end
+
+    context "an asset with the same legacy_url_path as an existing asset" do
+      let(:attributes) { FactoryBot.attributes_for(:whitehall_asset, :with_legacy_metadata) }
```
floehopper: Minor: Could you move the `let(:attributes)` statement into the outer context to avoid duplicating it here?

spec/controllers/whitehall_assets_controller_spec.rb
```diff
@@ -94,6 +94,17 @@
...
+      let!(:existing_asset) { FactoryBot.create(:whitehall_asset, legacy_url_path: attributes[:legacy_url_path]) }
+
+      it "marks the existing asset as deleted" do
+        post :create, params: { asset: attributes }
+
+        expect(existing_asset.reload).to be_deleted
```
floehopper: Observation: I'm slightly surprised that reloading a deleted asset works - I'd have thought it might result in a `RecordNotFound` exception.

chrislo: Yes, it's slightly odd. This pattern is used in the [mongoid_paranoia](https://github.com/simi/mongoid_paranoia/blob/e0c6608daa61c8ecf1165e278565b1b6da1f2671/spec/mongoid/nested_attributes_spec.rb#L98) specs too.

spec/controllers/whitehall_assets_controller_spec.rb
chrislo: I'll make that change.

app/controllers/whitehall_assets_controller.rb
chrislo: I can't think of a way of doing this without adding two methods, one to find and one to check the existence. I'll leave this change for now and see if I can come up with a refactoring in another PR.

chrislo:
Rebasing and force-pushing before merging.

PR #338, opened 2017-12-12
Add env vars to help end-to-end tests run in rails production mode

adrianclay:
Hey :wave:,

To help us enable rails production mode when running alphagov/publishing-e2e-tests we've added a new ENV var for using the fake S3 mode while still being in rails prod mode.

Similarly the latest version of gds-sso introduces a similar variable for using mock SSO and also being in rails production mode.

### DIFF
--- a/lib/s3_configuration.rb
+++ b/lib/s3_configuration.rb
@@ -10,7 +10,7 @@ class S3Configuration
   end
 
   def bucket_name
-    if Rails.env.production?
+    if Rails.env.production? && !ENV['USE_FAKE_S3']
       @env.fetch('AWS_S3_BUCKET_NAME')
     else
       @env['AWS_S3_BUCKET_NAME']


floehopper:
@adrianclay: Can we hold off on this change until we get https://github.com/alphagov/publishing-e2e-tests/pull/93 merged and running OK? Also I'm slightly wary about introducing (yet) another configuration option. I'll try to give this some more thought as soon as I can.

config/routes.rb
```diff
@@ -11,7 +11,7 @@
   get "/media/:id/:filename" => "media#download", :constraints => { filename: /.*/ }
   get "/government/uploads/*path" => "whitehall_media#download"
 
-  if Rails.env.development?
+  if Rails.env.development? || ENV["USE_FAKE_S3"]
```
floehopper: I think it would be better to extract the concept of using the fake S3 mechanism into a single place to reduce duplication and make the code easier to understand.

adrianclay:
> I'm not convinced that using the fake S3 mechanism in a production-like environment in the e2e tests is a sensible approach.

We're keen to reduce the difference between production and the end-to-end testing env as much as possible.  We wanted to enable production ENV on all the apps, but have had to make some slices in places to deliver sooner.  I've added a [card](https://trello.com/c/2ExdoGwV) to our roadmap to come back to this.

> A better alternative might be to run a Minio server in a Docker container (using this image) to simulate S3 more accurately. That way we could reduce the number of configuration options in Asset Manager and provide a more realistic environment for the e2e tests.

Nice link, we were looking at https://github.com/localstack/localstack as a solution.  Minio looks very similar.

config/routes.rb
floehopper: @adrianclay FYI I'm about to extract this concept for other reasons as a separate PR.

floehopper: I've opened https://github.com/alphagov/asset-manager/pull/349 to address this concern.

RyanMacG:
@floehopper We've updated this based off of your configuration refactor and removed the SSO bump as there is now a Dependabot PR #353 

We'd like to get this merged to progress the rails production work we've done. When Kevin is back from his AL we can discuss prioritising the S3 work.

floehopper:
@RyanMacG:

Thanks for updating the PR against my changes.

I think the logic you're adding now belongs in [`S3Configuration#fake?`][1] rather than `S3Configuration#bucket_name`. Also I would expect to see an additional example or two explaining the intention behind the change in the `S3Configuration` spec.

I think it might also be better to use the presence/absence of the `FAKE_S3_HOST` environment variable which is [already set in the e2e tests project][2], as opposed to creating a new environment variable, `USE_FAKE_S3 `. What do you think?

> We'd like to get this merged to progress the rails production work we've done. When Kevin is back from his AL we can discuss prioritising the S3 work.

We still feel uneasy that this change means the fake S3 functionality could accidentally be switched on in production. We were wary about adding the behaviour to the Rails app in the first place even just for development. Have you tried the Minio/Localstack approach? Have you run into problems with that approach? When is Kevin back?

/cc @chrisroos 

[1]: https://github.com/alphagov/asset-manager/blob/master/lib/s3_configuration.rb#L26-L28
[2]: https://github.com/alphagov/publishing-e2e-tests/blob/1e82539d6c294019bd8a318d36ad7d270b6c9031/docker-compose.yml#L583

RyanMacG:
Hey @floehopper.

Thanks for taking another look at this.  We've just tried putting the check inside of the ` fake? ` method but had trouble.  The method ` S3Configuration.build ` gets called from ` config/initializers/aws.rb ` which blows up with this error:

```
KeyError: key not found: "AWS_S3_BUCKET_NAME"                       
/app/lib/s3_configuration.rb:14:in `fetch'                          
/app/lib/s3_configuration.rb:14:in `bucket_name'                    
/app/lib/s3_configuration.rb:21:in `configured?'                    
/app/lib/s3_configuration.rb:4:in `build'                           
/app/config/initializers/aws.rb:3:in `<top (required)>'             
```

adrianclay:
>We still feel uneasy that this change means the fake S3 functionality could accidentally be switched on in production. We were wary about adding the behaviour to the Rails app in the first place even just for development. Have you tried the Minio/Localstack approach? Have you run into problems with that approach? When is Kevin back?

I can appreciate your concern about adding code purely for the sake of the E2E test suite.  In my opinion the ENV var is named such that it would be unlikely for it to happen ` USE_FAKE_S3 ` but would defer my opinion in favour of those involved with the deployment process.

We looked into the Localstack approach, but were coming up against time constraints and so agreed to drop it as it didn't feel like a priority.  I'd estimate with another half day of investigation we'd get it hooked up and working 🤞.

Kevin is back on the 28th, @thomasleese is filling in for him this week.

floehopper:
@RyanMacG: OK. I'll have a quick think about this and get back to you shortly.

@adrianclay: Thanks for answering my questions. My worry is that by removing this obstacle, there will *never* be pressure to adopt the Minio/Localstack approach. If you can assure me it's on the roadmap, then I think I can live with it. I think I might suggest an even more obvious name for the environment variable, but I'll include that with my response to @RyanMacG. I hope that's OK.

floehopper:
@RyanMacG, @adrianclay: I've opened [a new pull request](https://github.com/alphagov/asset-manager/pull/361) which supersedes this one. I hope it makes sense. In the meantime, I'm going to close this one. I hope that's OK.

PR #340, opened 2017-12-14
Fix Whitehall asset lookup when URL path has no format/file extension

floehopper:
[We were re-constructing the `legacy_url_path` to search for by assuming there was a format][1]. Thus if an asset had no file extension, e.g. `/government/uploads/foo`, we were attempting to lookup the asset using `/government/uploads/foo.` (note the trailing full-stop). This did not find the asset and the app responded with a `404 Not Found`.

[1]:
https://github.com/alphagov/asset-manager/blob/c14c324b4f41467ab6356599bc95ea92b34b99f6/app/controllers/whitehall_media_controller.rb#L3-L4

### DIFF
--- a/app/controllers/whitehall_media_controller.rb
+++ b/app/controllers/whitehall_media_controller.rb
@@ -1,6 +1,7 @@
 class WhitehallMediaController < BaseMediaController
   def download
-    path = "/government/uploads/#{params[:path]}.#{params[:format]}"
+    path = "/government/uploads/#{params[:path]}"
+    path += ".#{params[:format]}" if params[:format].present?
     asset = WhitehallAsset.find_by(legacy_url_path: path)
 
     if asset.unscanned?
--- a/spec/controllers/whitehall_media_controller_spec.rb
+++ b/spec/controllers/whitehall_media_controller_spec.rb
@@ -18,6 +18,16 @@ RSpec.describe WhitehallMediaController, type: :controller do
 
         get :download, params: { path: path, format: format }
       end
+
+      context 'and legacy_url_path has no format' do
+        let(:legacy_url_path) { "/government/uploads/#{path}" }
+
+        it "proxies asset to S3 via Nginx" do
+          expect(controller).to receive(:proxy_to_s3_via_nginx).with(asset)
+
+          get :download, params: { path: path, format: nil }
+        end
+      end
     end
 
     context 'when asset is unscanned image' do


app/controllers/whitehall_media_controller.rb
```diff
@@ -1,6 +1,7 @@
 class WhitehallMediaController < BaseMediaController
   def download
-    path = "/government/uploads/#{params[:path]}.#{params[:format]}"
+    path = "/government/uploads/#{params[:path]}"
+    path += ".#{params[:format]}" if params[:format].present?
```
chrislo: Minor: I might be tempted to use [Pathname#sub_ext](https://ruby-doc.org/stdlib-2.4.0/libdoc/pathname/rdoc/Pathname.html#method-i-sub_ext) to perform this kind of path manipulation, but I don't think it makes much difference.

floehopper:
@chrislo:

Thanks for reviewing:

> Was there an issue in production that caused you to find this bug? 

No. I came across it while doing some manual testing on my development VM. I'm not sure how real a problem it is in production - I suspect most/all assets have a file extension.

app/controllers/whitehall_media_controller.rb
floehopper: Oh. That's interesting. I haven't come across that method before. Thanks!

floehopper:
This was merged and deployed some time ago, so I've just checked that the fix is working in integration:

```
$ export BEARER_TOKEN=`cat /etc/govuk/manuals-publisher/env.d/ASSET_MANAGER_BEARER_TOKEN`

$ curl -H"Authorization: Bearer $BEARER_TOKEN" -H"Accept: application/json" https://asset-manager.integration.publishing.service.gov.uk/whitehall_assets --form "asset[file]=@foo" --form "asset[legacy_url_path]=/government/uploads/another/path/to/foo"
{"_response_info":{"status":"created"},"id":"https://asset-manager.integration.publishing.service.gov.uk/assets/5a3cd71bed915d07938ab2c5","name":"foo","content_type":"application/octet-stream","file_url":"https://assets-origin.integration.publishing.service.gov.uk/government/uploads/another/path/to/foo","state":"unscanned"}

$ curl -H"Authorization: Bearer $BEARER_TOKEN" -H"Accept: application/json" https://asset-manager.integration.publishing.service.gov.uk/assets/5a3cd71bed915d07938ab2c5
{"_response_info":{"status":"ok"},"id":"https://asset-manager.integration.publishing.service.gov.uk/assets/5a3cd71bed915d07938ab2c5","name":"foo","content_type":"application/octet-stream","file_url":"https://assets-origin.integration.publishing.service.gov.uk/government/uploads/another/path/to/foo","state":"uploaded"}

$ curl -s -v https://asset-manager.integration.publishing.service.gov.uk/government/uploads/another/path/to/foo
* Hostname was NOT found in DNS cache
*   Trying 10.1.3.254...
* Connected to asset-manager.integration.publishing.service.gov.uk (10.1.3.254) port 443 (#0)
* successfully set certificate verify locations:
*   CAfile: none
  CApath: /etc/ssl/certs
* SSLv3, TLS handshake, Client hello (1):
* SSLv3, TLS handshake, Server hello (2):
* SSLv3, TLS handshake, CERT (11):
* SSLv3, TLS handshake, Server key exchange (12):
* SSLv3, TLS handshake, Server finished (14):
* SSLv3, TLS handshake, Client key exchange (16):
* SSLv3, TLS change cipher, Client hello (1):
* SSLv3, TLS handshake, Finished (20):
* SSLv3, TLS change cipher, Client hello (1):
* SSLv3, TLS handshake, Finished (20):
* SSL connection using ECDHE-RSA-AES256-GCM-SHA384
* Server certificate:
* 	 subject: OU=Domain Control Validated; OU=Gandi Standard Wildcard SSL; CN=*.integration.publishing.service.gov.uk
* 	 start date: 2017-10-26 00:00:00 GMT
* 	 expire date: 2019-12-19 23:59:59 GMT
* 	 subjectAltName: asset-manager.integration.publishing.service.gov.uk matched
* 	 issuer: C=FR; ST=Paris; L=Paris; O=Gandi; CN=Gandi Standard SSL CA 2
* 	 SSL certificate verify ok.
> GET /government/uploads/another/path/to/foo HTTP/1.1
> User-Agent: curl/7.35.0
> Host: asset-manager.integration.publishing.service.gov.uk
> Accept: */*
> 
< HTTP/1.1 200 OK
* Server nginx is not blacklisted
< Server: nginx
< Date: Fri, 22 Dec 2017 09:58:44 GMT
< Content-Type: application/octet-stream
< Content-Length: 29
< Connection: close
< Content-Disposition: inline; filename="foo"
< Cache-Control: max-age=14400, public
< Accept-Ranges: bytes
< ETag: "5a3cd71b-1d"
< Last-Modified: Fri, 22 Dec 2017 09:57:47 GMT
< X-Frame-Options: SAMEORIGIN
< Strict-Transport-Security: max-age=31536000
< 
Fri Dec 22 09:54:41 UTC 2017
* Closing connection 0
* SSLv3, TLS alert, Client hello (1):
```

The fact that we can create an asset at `/government/uploads/another/path/to/foo` (i.e. without a format) and request it successfully means that the bug is fixed.

Issue #238, opened 2017-09-27
Binary file content being displaying in Internet Explorer


chrisroos:
We've had some reports in Zendesk of users seeing binary file content being displayed in Internet Explorer when clicking on links to assets.

Examples of problem URLs:

* [Shared_Ownership_Flat_Lease.doc](https://assets.publishing.service.gov.uk/media/58075723e5274a67eb000010/Shared_Ownership_Flat_Lease.doc), linked from [Help to Buy: Shared Ownership](https://www.gov.uk/guidance/capital-funding-guide/1-help-to-buy-shared-ownership).

* [C1 Communicationsplan form](https://assets.publishing.service.gov.uk/media/59a7ead440f0b6616a1cbd33/C1-communications-plan-form.docx), linked from [UK Aid Match](https://www.gov.uk/international-development-funding/uk-aid-match)

chrisroos:
I've been able to replicate the problem in Internet Explorer 11 on a Windows Server 2016 EC2 instance:

This is a request for https://assets.publishing.service.gov.uk/media/59a7ead440f0b6616a1cbd33/C1-communications-plan-form.docx.

![screenshot 2017-09-27 12 46 27](https://user-images.githubusercontent.com/6556/30911795-eac94250-a381-11e7-9ae5-78267bdf263a.png)

chrisroos:
The Zendesk reports suggest that the problem is intermittent and I've had that experience in IE too.

chrisroos:
I'm fairly confident that this problem is triggered in IE when no `Content-Type` header is set. I tested this by uploading an MS Word document to [Digital Ocean Spaces](https://www.digitalocean.com/products/object-storage/) and toggling the `Content-Type` between `binary/octet-stream` and an empty string:

* With a `Content-Type` of `binary/octet-stream`, IE would prompt me to download the file.
* With a `Content-Type` of an empty string, IE would display the file inline in the browser.

I initially tried to use AWS S3 to perform this test but I don't seem to be able to remove the `Content-Type` header from objects in S3.

chrisroos:
I've used curl to confirm that some responses for assets __don't__ have a `Content-Type` header set:

`curl -v https://assets.publishing.service.gov.uk/media/59a7ead440f0b6616a1cbd33/C1-communications-plan-form.docx > /dev/null`

* `X-Served-By: cache-lhr6338-LHR` - No `Content-Type`
[...]

PR #343, opened 2017-12-14
Prevent nginx unnecessarily downloading assets from S3

chrisroos:
This should avoid the problem captured in issue #238 ("Binary file content being displaying in Internet Explorer").

We believe that the problem was being triggered when Fastly made a conditional GET request for an asset that it had stored in its cache. Rails (or more specifically, [`Rack::ConditionalGet`][1]) would respond with a 304 Not Modified and an empty `Content-Type` and empty `Content-Length` header. This would've been fine if it didn't also send the `X-Accel-Redirect` header. The presence of this header was telling Nginx to request the file from S3 again. Nginx would then send a 200 response containing the asset but with an empty `Content-Type` header!

By not sending the `X-Accel-Redirect` header along with a 304 response, we ensure that Nginx passes the 304 back to Fastly as expected.

Note that I test-drove a hand-rolled implementation of `request.fresh?(response)` to give myself confidence that it was behaving as I expected, before then switching to the Rails implementation.

[1]: https://github.com/rack/rack/blob/master/lib/rack/conditional_get.rb

### DIFF
--- a/app/controllers/base_media_controller.rb
+++ b/app/controllers/base_media_controller.rb
@@ -4,23 +4,25 @@ class BaseMediaController < ApplicationController
 protected
 
   def proxy_to_s3_via_nginx(asset)
+    headers['ETag'] = %{"#{asset.etag}"}
+    headers['Last-Modified'] = asset.last_modified.httpdate
+    headers['Content-Disposition'] = AssetManager.content_disposition.header_for(asset)
+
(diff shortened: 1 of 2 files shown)


chrisroos:
Thanks for reviewing, @floehopper. I'm going to get this merged.

chrislo:
Lovely work @chrisroos well done!

PR #349, opened 2017-12-15
DRY-up logic for enabling fake S3 functionality

floehopper:
I have a number of motivations for doing this:

* There is currently some implicit duplication in logic for this - in `config/routes.rb` and in `S3Storage.build`
* I want to be able to enable the fake S3 functionality for an integration test and this might help
* It addresses the concern I raised in [this PR review comment](https://github.com/alphagov/asset-manager/pull/338#discussion_r156611880).

I've ended up doing a bit of related tidying up along the way:

* Extract `S3Configuration` class c.f. `FakeS3Configuration` and add specs for it
* Remove `S3Storage::Null` which is no longer needed

Note that this PR is based on the work in https://github.com/alphagov/asset-manager/pull/344, so it should ideally be rebased against that branch before merging.


floehopper:
Now that #344 has been merged, I've rebased this against `master` and am force-pushing.

floehopper:
@chrisroos Thanks for reviewing. I'll get it merged.

PR #328, opened 2017-12-06
Remove options to serve assets from NFS via Nginx

floehopper:
Currently both feature flags are [set to 100% in all environments][1], so *all* asset requests are being proxied to S3 via Nginx. We're happy that this mechanism is working OK and have no plans to rollback to proxying less than 100% of asset requests to S3 via Nginx, so there's no longer
any need for either the environment variable feature flags nor the request parameter overrides. This pull request hard-codes these configuration options and simplifies the application code accordingly.

My actual motivation for doing this is that I want to [start deleting asset files from NFS when they've been uploaded to S3](https://github.com/alphagov/asset-manager/issues/323). These changes should make that easier, because it removes one key method (`BaseMediaController#serve_from_nfs_via_nginx`) which was relying on the asset's file existing on NFS.

Note that once this has been deployed to all environments, we should remove the environment variables which will be made redundant by this PR from `govuk-puppet` to avoid future confusion.

[1]:
https://github.com/alphagov/govuk-puppet/blob/476d9e764b42348deb8a2253be78d460d44c33c0/hieradata/common.yaml#L523-L524

PR #358, opened 2017-12-18
Avoid "X-Accel-Mapping header missing" warning

floehopper:
Since alphagov/asset-manager#328 the Rails app only ever proxies asset requests to S3; it no longer instructs Nginx to serve asset files from NFS. And so this setting is no longer needed. It seems as if it is this setting which is causing `Rack::Sendfile` to emit the warning mentioned above.

We're planning to remove the redundant `Sendfile`-related config from the asset-manager Nginx virtual host in alphagov/govuk-puppet#6971, but this change doesn't need to wait until that is deployed/applied.

Note that The x_sendfile_header configuration option was originally set to `X-Sendfile` in [this commit][1] and then to `X-Accel-Redirect` in [this commit][2].

[1]: https://github.com/alphagov/asset-manager/commit/9be379ed8a8f5622a67151884325d21d52c8f3a3
[2]: https://github.com/alphagov/asset-manager/commit/9be379ed8a8f5622a67151884325d21d52c8f3a3

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -32,8 +32,6 @@ module AssetManager
     # Disable Rack::Cache
     config.action_dispatch.rack_cache = nil
 
-    config.action_dispatch.x_sendfile_header = "X-Accel-Redirect"
-
     config.assets.prefix = '/asset-manager'
   end
 


chrisroos:
@floehopper: The "this commit" links in the description are identical. It looks as though the first one is correct and that the second one should be updated.

floehopper:
@chrisroos:

Thanks for reviewing.

> The "this commit" links in the description are identical. It looks as though the first one is correct and that the second one should be updated.

Well spotted. I'll fix that, force-push and get this merged.

PR #360, opened 2017-12-19
Upgrade aws-sdk from v2.10.88 to v2.10.105

floehopper:
This is the latest v2 release.

I ran the following command to achieve this:

    bundle update --conservative aws-sdk

This is an alternative to #352 by Dependabot which upgrades to v3 and which includes breaking changes to the API.


floehopper:
@chrislo:

Thanks for reviewing.

> Is it worth putting a `~> 2` in the Gemfile so that dependabot doesn't try to upgrade to v3 again?

The `Gemfile` already has this constraint: `'~> 2.10.10'`, but Dependabot still seems to want to upgrade to v3. Not sure why. It would be nice to make it less aggressive.

You can [tell Dependabot to only try again when there is a new minor or major version](https://github.com/alphagov/asset-manager/pull/352#issuecomment-352690768), but I'm not sure that's what we want.

PR #361, opened 2017-12-19
Allow use of fake S3 in non-development environments

floehopper:
The motivation behind this is two-fold:

1. I want to be able to use the fake S3 functionality in the Rails test environment in order to write some integration testing which involve uploading to S3.
2. The team working on the publishing-e2e-tests project want to be able to use the fake S3 functionality in the tests when running the app in the Rails production environment.

The idea is to make the methods on `S3Configuration` as self-explanatory as possible and to make it hard to enable the fake S3 functionality in production by accident.

I've intentionally not added the new environment variable, `ALLOW_FAKE_S3_IN_PRODUCTION_FOR_PUBLISHING_E2E_TESTS` to the documentation, because I don't want to encourage its use and in any case, its name makes its purpose pretty clear.

This supersedes #338.

### DIFF
--- a/lib/s3_configuration.rb
+++ b/lib/s3_configuration.rb
@@ -1,6 +1,6 @@
 class S3Configuration
-  def self.build
-    config = new
+  def self.build(env = ENV)
+    config = new(env)
     config.check!
     config
   end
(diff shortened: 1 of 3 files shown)


lib/s3_configuration.rb
```diff
@@ -24,6 +24,6 @@ def check!
   end
 
   def fake?
-    !configured? && Rails.env.development?
+    !configured? && !Rails.env.production?
```
chrislo: I find the double negative here a little hard to read. Would something like 

```
def real?
  configured? && Rails.env.production?
end
```

work in the context of the other changes you've made?

floehopper: I take your point, but that method ends up like this in the end:

```ruby
  def fake?
    !configured? && allow_fake?
  end
```

I think this is more readable and I think I prefer this to having to use `if !AssetManager.s3.real?` or `unless AssetManager.s3.real?` elsewhere in the codebase, e.g. `config/routes.rb`.

Issue #323, opened 2017-12-05
Remove new Asset Manager asset files from NFS once they have been uploaded to S3


floehopper:
All Asset Manager assets are now stored and served from S3, so there is no longer any need to store the asset files on NFS as well as S3. The asset files will still need to be saved to NFS on upload so that the Sidekiq workers (which might be on a different machine) can access the file for virus scanning and S3 upload. However, once the file has been uploaded to S3, we should be able to delete the file from NFS.

Note that we'll need to be careful that methods on the `Asset` class which rely on an underlying file being present continue to work in a sensible fashion when the underlying file has been removed.

Note also that we'll also need to [remove asset files for existing assets](https://github.com/alphagov/asset-manager/issues/296) after we've done this.

floehopper:
[This PR](https://github.com/alphagov/asset-manager/pull/328) is somewhat relevant to this issue.

floehopper:
https://github.com/alphagov/asset-manager/pull/328 has now been merged.

floehopper:
As a precursor to this I'm currently looking at extending the examples in `spec/requests/virus_scanning_spec.rb` to include the upload to S3. This will give me more confidence that nothing is broken when we start deleting files from NFS after the S3 upload.

As a precursor to the above, I'm looking at making it possible to use the fake S3 functionality within the specs. And in order to do this I'm first going to address a problem with the configuration of this which I highlighted in [this comment](https://github.com/alphagov/asset-manager/pull/338#discussion_r156611880).

floehopper:
I'm also planning to address this bug as part of the above work: https://github.com/alphagov/asset-manager/issues/281

floehopper:
I've opened https://github.com/alphagov/asset-manager/pull/373 to do this.

floehopper:
https://github.com/alphagov/asset-manager/pull/373 has now been deployed to integration, staging & production and so we are now deleting files from NFS for new assets once their files have been uploaded to S3. This PR also included a Rake task which will make it possible to implement https://github.com/alphagov/asset-manager/issues/296. Closing.

Issue #281, opened 2017-10-31
404 Not Found when new asset is virus scanned but not uploaded to S3


floehopper:
I haven't actually checked this, but I'm pretty sure an asset's state is updated to clean if the virus scan job completes successfully and the upload to S3 is only triggered at this point. If the upload to S3 takes some time or fails for some reason, I think the app will attempt to proxy a request for the asset to S3 which will presumably respond with a 404 Not Found.

Pragmatically, the S3 upload usually runs immediately after the virus scan and doesn't usually take very long to complete, so the time window in which this could happen is very short. However, it feels as if we're modelling the asset state incorrectly and this feels more significant now we're serving all assets from S3 in production.

floehopper:
I've just realised that Asset Manager responds with a `404 Not Found` when a Mainstream asset that hasn't been virus scanned is requested and so this isn't so much of an issue for Mainstream assets. However, for Whitehall assets this bug will result in a placeholder image/page redirect while the asset is being virus scanned, then a 404 Not Found (from S3) until the asset is uploaded to S3, and then finally the asset will be served correctly. The behaviour should be to continue to do the placeholder redirect until the asset is uploaded to S3.

floehopper:
I've opened #369 to address this.

floehopper:
https://github.com/alphagov/asset-manager/pull/369 has been deployed to integration, staging & production. I couldn't come up with an easy way to test this on any of the production environments, but I'm happy that the specs test this behaviour pretty thoroughly - particularly in [this example](https://github.com/alphagov/asset-manager/blob/8c8b87d158e47c0a165162c35fe426c7ba2834c6/spec/requests/virus_scanning_spec.rb#L8-L34). Closing.

PR #363, opened 2017-12-20
Introduce uploaded states for assets

floehopper:
This introduces two new asset states: "uploaded" and "not_uploaded". New assets are still scanned for viruses and if no viruses are found they are marked as "clean". However, now they will be marked as "uploaded" or "not_uploaded" depending on whether the upload to cloud storage is successful.

For the moment, assets in the "clean", "uploaded" and "not_uploaded" states are treated the same. I've made use of shared examples in the two media controller specs to check this without introducing duplication. However, once the `govuk:set_uploaded_state_for_all_clean_assets` Rake task included in this PR has been run, we'll be in a position to treat these states differently and at that point we should be able to do away with the shared examples.

The motivation behind this PR is two-fold:

1. Now that all asset requests are proxied to S3, we shouldn't attempt to make an asset available via the media controllers until the asset has been successfully uploaded to S3. See #281 for details.

2. We want to remove files from NFS once they have been uploaded to S3. Having these new separate asset states should make this easier to reason about. See #323 for details.

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -52,6 +52,14 @@ class Asset
     event :scanned_infected do
       transition any => :infected
     end
+
+    event :upload_success do
+      transition %i(clean uploaded not_uploaded) => :uploaded
+    end
+
+    event :upload_failure do
+      transition %i(clean uploaded not_uploaded) => :not_uploaded
+    end
   end
 
   def public_url_path
--- a/lib/tasks/govuk_assets.rake
+++ b/lib/tasks/govuk_assets.rake
@@ -1,0 +1,12 @@
+require 'services'
+require 'asset_processor'
+
+namespace :govuk_assets do
+  desc 'Set uploaded state for all clean assets (uploaded/not_uploaded)'
+  task set_uploaded_state_for_all_clean_assets: :environment do
+    processor = AssetProcessor.new(scope: Asset.where(state: 'clean'))
+    processor.process_all_assets_with do |asset_id|
+      AssetUploadStateWorker.perform_async(asset_id)
+    end
+  end
+end
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -29,7 +29,7 @@ protected
 
   def asset_servable?
     asset.filename_valid?(params[:filename]) &&
-      asset.clean? &&
+      (asset.clean? || asset.uploaded? || asset.not_uploaded?) &&
       asset.accessible_by?(current_user) &&
       asset.mainstream?
   end
--- a/app/workers/asset_upload_state_worker.rb
+++ b/app/workers/asset_upload_state_worker.rb
@@ -1,0 +1,15 @@
+require 'services'
+
+class AssetUploadStateWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: 'low_priority'
+
+  def perform(asset_id)
+    asset = Asset.find(asset_id)
+    if Services.cloud_storage.exists?(asset)
+      asset.upload_success!
+    else
+      asset.upload_failure!
+    end
+  end
+end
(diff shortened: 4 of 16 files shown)


lib/tasks/govuk_assets.rake
```diff
@@ -0,0 +1,12 @@
+require 'services'
+require 'asset_processor'
+
+namespace :govuk_assets do
+  desc 'Set uploaded state for all clean assets (uploaded/not_uploaded)'
+  task set_uploaded_state_for_all_clean_assets: :environment do
+    processor = AssetProcessor.new(scope: Asset.where(state: 'clean'))
```
chrisroos: Should we use `Asset.unscoped.where(state: 'clean')` so that we also update the state of deleted assets?

app/models/asset.rb
```diff
@@ -52,6 +52,14 @@ class Asset
     event :scanned_infected do
       transition any => :infected
     end
+
+    event :upload_success do
+      transition %i(clean uploaded not_uploaded) => :uploaded
```
chrisroos: Do we definitely need to be able to transition from `uploaded` to `uploaded`? Presumably if we were to upload a new file for the `Asset` then we'd also reset the state back to `unscanned` which would mean it transitions through `clean` to `uploaded`.

app/models/asset.rb
```diff
@@ -52,6 +52,14 @@ class Asset
...
+    end
+
+    event :upload_failure do
+      transition %i(clean uploaded not_uploaded) => :not_uploaded
```
chrisroos: Similar to the question above, do we definitely need to be able to transition from `uploaded` to `not_uploaded` and from `not_uploaded` to `not_uploaded`?

chrisroos: Did you consider adding the `uploaded` state only, and using `clean` to mean that it's been virus scanned but not uploaded? Is there any benefit to knowing that the upload has been tried and failed over knowing that the upload has not yet succeeded?

app/models/asset.rb
floehopper: In general, I was leaning towards not being too restrictive on the state transitions, because I didn't want an unforeseen transition to cause the app to start failing. Although I wanted to be more restrictive than the completely unconstrained transitions into `clean` and `infected` states.

In this case, I was thinking that in the past we've run Rake tasks to re-upload the same files, i.e. re-do the upload with exactly the same file as before.

What are the benefits of preventing the transition from `uploaded` to `uploaded`?

app/models/asset.rb
floehopper: `uploaded` -> `not_uploaded`: If we had a Rake task that was re-uploading the existing files and the new upload failed, we might need this.

`not_uploaded` -> `not_uploaded`: If we had a Rake task that was re-uploading the existing files for assets which had previously failed to upload (and none of the Sidekiq retrying had worked) and the new upload failed, we might need this.

What are the benefits of preventing the transitions you mention?

lib/tasks/govuk_assets.rake
floehopper: Yes. Well spotted! I'd completely forgotten about the deleted assets (again!).

app/models/asset.rb
floehopper: No, I hadn't considered that. I think it's a good idea.

floehopper:
@chrisroos Thanks for your feedback. I’m going to change the PR to only add an `uploaded` state (no `not_uploaded`) as you suggest - I think it’s a good idea. I’ll also add the `unscoped` scope to the Rake task. Can you have a read of my responses to your two questions about the state transitions and see what you think. I guess the 2nd one will become irrelevant in the new version of the PR without the `not_uploaded` state, but the I think the 1st one is still relevant.

app/models/asset.rb
chrisroos: > In this case, I was thinking that in the past we've run Rake tasks to re-upload the same files, i.e. re-do the upload with exactly the same file as before.

Good point. Although this isn't going to be possible as soon as we remove the files from NFS.

> What are the benefits of preventing the transition from `uploaded` to `uploaded`?

I guess I think it's a bit easier to understand if we can only transition from `clean` to `uploaded`.

floehopper: OK. I'll go with that approach. Thanks.

floehopper:
Closed in favour of #364.

PR #365, opened 2017-12-21
Upgrade govuk_sidekiq to v3

floehopper:
The motivation behind this is to fix the Grafana dashboard for queue size for the Asset Manager app.

I found that I also needed to upgrade `govuk_app_config` and hence remove `unicorn` & `logstasher` from the `Gemfile` altogether.


tijmenb:
You can also remove logstasher.rb: https://github.com/alphagov/asset-manager/blob/d7390ff2628a83a26e88a8a5cd9e9f928e6649eb/config/initializers/logstasher.rb

floehopper:
@tijmenb:

> You can also remove `logstasher.rb`

Ah, great - thanks!

floehopper:
@chrisroos: What do you think about merging this? I'm slightly wary of doing so just before we head off for the Xmas break!

floehopper:
Rebasing against `master`, resolving conflicts and force-pushing.

floehopper:
Given that the Xmas break has passed, I'm now happy to merge this.

floehopper:
This has been deployed to integration. However, when I manually added a new asset I didn't see the jobs appear in the queue in either the old or the new Grafana view, even though the asset has made it into the "uploaded" state. I've asked in Slack to see whether anyone knows what's going on.

floehopper:
I'm now happy that the app is working OK in integration - I tested using `curl` commands from the README.

Also @cbaines & @kevindew helped me [find the Sidekiq jobs on the queue in Grafana][1]. This was complicated by the low volume of jobs being processed - hence the "with points" aspect of the query.

Anyway, I'm now happy to proceed to staging & production.

[1]: https://grafana.integration.publishing.service.gov.uk/dashboard/db/sidekiq-with-points?refresh=1m&orgId=1&from=now-15m&to=now

floehopper:
This has now been deployed to staging and production and the app seems to be working fine.

@cbaines has opened https://github.com/alphagov/govuk-puppet/pull/7006 and plans to fix the Sidekiq (Older versions ...) dashboard manually.

Issue #326, opened 2017-12-06
Check that create_asset & update_asset are up-to-date/working and add specs for them


floehopper:
I don't think these two scripts in the `bin` directory have any specs around them and so it's possible they are now broken. I believe they are still referenced in some GOV.UK documentation and so we should probably make sure they are working.

floehopper:
I've opened https://github.com/alphagov/asset-manager/pull/375 to address this.

floehopper:
I didn't want to spend too much time on this, because I think the effort would be better spent building a simple web UI for Asset Manager.

floehopper:
https://github.com/alphagov/asset-manager/pull/375 has been deployed to integration, staging & production. Closing.

PR #375, opened 2018-01-02
Add test coverage for scripts

floehopper:
I've extracted the two custom scripts (`bin/create_asset` & `bin/update_asset`) into methods on a new class (`CLI`) and added a spec for that class.

In doing this I identified a slight flaw in the argument parsing of `bin/update_asset` which I have fixed. Other than that I've attempted to keep the code as unchanged as possible, albeit with some changes to make testing easier. This is not because I think the code is especially good, but because I wanted to minimise the chance of introducing bugs.

I did contemplate turning them into Rake tasks as we did in Manuals Publisher, but I realised that would mean updating [the corresponding documentation](https://github.com/alphagov/govuk-developer-docs/blob/8757e2bcc61469b8d94f3d09a552c1608b17a90c/source/manual/howto-upload-an-asset-to-asset-manager.html.md).

### DIFF
--- a/bin/create_asset
+++ b/bin/create_asset
@@ -2,22 +2,6 @@
 
 require File.expand_path("../../config/environment", __FILE__)
 
-filename = ARGV.any? ? ARGV.fetch(0) : nil
+require 'cli'
 
-unless filename
-  puts "You need to provide a filename as first argument when running this script"
-  abort
-end
-
-file = File.new(filename)
-asset = Asset.new(file: file)
-
-if asset.save
-  puts "Saved!"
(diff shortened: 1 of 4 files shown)


floehopper:
> I think at some point it'd be great if we added a small web UI to the app to allow people to manually manage assets instead of having to use the CLI.

Yes - I agree - see my comment here: https://github.com/alphagov/asset-manager/issues/326#issuecomment-354839401. I've added https://github.com/alphagov/asset-manager/issues/380 to capture the idea, because we've talked about it quite a few times and I couldn't see it in the list of issues.

PR #373, opened 2017-12-29
Delete underlying file when upload to S3 succeeds

floehopper:
Since #328 the app hasn't been capable of instructing Nginx to serve assets directly from the NFS mount and since alphagov/govuk-puppet#6971 Nginx hasn't been configured to allow assets to be served directly from the NFS mount.

This means that the only reason we need to store files on NFS is currently so that the Sidekiq workers (which may be running on different machines to that which handled the upload request) have access to the files.

This in turn means that it is safe to remove the underlying file once the asset has been scanned for viruses and uploaded to S3.

In order to achieve this I've introduced a new state transition callback which calls `Asset#remove_file!`. The latter seems to be the recommended way to remove the underlying file via the Carrierwave API and results in `Asset#file` becoming blank which seems to make sense to me.

I've had to restrict the validation which checks for the presence of the file attribute in order to ensure that an uploaded asset remains valid. I've also had to prevent the `store_metadata` callback from running when the asset is in the uploaded state, because it relies on the underlying file in order to e.g. compute the `ETag` from the file metadata. I think disabling the latter should be OK, because it should only need to be run on Asset creation or update, i.e. when a new file is specified.

Once this commit has been deployed, the underlying files for assets should only exist on NFS temporarily, i.e. until they have been scanned for viruses and uploaded to S3. This means we should then be in a position to delete all the underlying files for existing assets in the uploaded state and free up a significant amount of space on NFS. I've included a new Rake task (`govuk_assets:delete_file_from_nfs_for_assets_uploaded_to_s3`) to make this possible.

PR #390, opened 2018-01-09
Remove empty CarrierWave directories when new asset is created and file is removed after upload to S3

floehopper:
Since #373 we've been deleting an asset's underlying file from the filesystem once the file has been uploaded to S3. However, this was leaving behind a bunch of empty directories. This change means that the app now removes all empty parent directories when the file is removed.

The immediate motivation behind this change is that once alphagov/govuk-puppet#7016 is applied in production, the Asset Manager CarrierWave directory will no longer be sync-ed from the Asset Master to the Asset Slaves and there will therefore be a danger that the [`asset_master_and_slave_disk_space_similar_from_xxx` Icinga alert][1] might eventually be triggered by the non-zero disk space taken up by the empty directories.

The implementation makes use of `FileUtils.rmdir` with the `parents` option which is equivalent to the [`rmdir -p` unix command][2] and which recursively removes empty directories from the supplied path upwards.

Note that I think this means that in some circumstances top-level directories (e.g. `tmp/test_uploads/assets` in the test environment) might be removed, but I'm pretty confident CarrierWave creates directories using the equivalent of `mkdir -p` and so any missing intermediate directories will be created and whatever happens it will stop when it gets to the Rails project root which will never be empty.

As far as I can tell, the files and empty directories under the CarrierWave "cache" directory are automatically deleted by CarrierWave, so I don't think we need to worry about disk usage for that directory.

Note also that I've changed the specs so that the CarrierWave directory is cleaned out after *every* example, not just after all the specs have run. This was in order to make testing this change easier.

[1]: https://github.com/alphagov/govuk-puppet/blob/19837b4b351d97d84570bd8a7d01ef3420fbef94/modules/govuk/manifests/node/s_asset_slave.pp#L50-L63
[2]: http://www.linfo.org/rmdir.html

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -58,7 +58,9 @@ class Asset
     end
 
     after_transition to: :uploaded do |asset, _|
+      path = asset.file.path
       asset.remove_file!
+      FileUtils.rmdir(File.dirname(path), parents: true)
     end
   end
 
--- a/spec/models/asset_spec.rb
+++ b/spec/models/asset_spec.rb
@@ -525,6 +525,10 @@ RSpec.describe Asset, type: :model do
     context 'when asset is clean' do
       let(:asset) { FactoryBot.create(:clean_asset) }
       let(:path) { asset.file.path }
+      let(:pathname) { Pathname.new(path) }
+      let(:dir) { pathname.parent }
+      let(:parent_dir) { pathname.parent.parent }
+      let(:grandparent_dir) { pathname.parent.parent.parent }
 
       it 'changes asset state to uploaded' do
         asset.upload_success!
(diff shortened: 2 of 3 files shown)


floehopper:
I've just tested this on my development VM (see below). I can see that the directories under `uploads/assets` are removed after upload to S3 as intended. As I thought, it is possible for the `uploads/assets` directory itself to be removed if it's empty. However, this doesn't seem to prevent CarrierWave from saving new/updated assets to disk - it just re-creates the top-level directory.

## Initial directory state

```
~/Code/alphagov/asset-manager$ tree uploads/
uploads/

0 directories, 0 files
```

## Create new asset

```
$ echo `date` > tmp.txt
$ cat tmp.txt 
Wed 10 Jan 2018 11:19:54 GMT
$ curl http://asset-manager.dev.gov.uk/assets --form "asset[file]=@tmp.txt"
{"_response_info":{"status":"created"},"id":"http://asset-manager.dev.gov.uk/assets/5a55f73c759b74438e3806bd","name":"tmp.txt","content_type":"text/plain","file_url":"http://assets-origin.dev.gov.uk/media/5a55f73c759b74438e3806bd/tmp.txt","state":"unscanned"}
```

### Before upload to S3

```
~/Code/alphagov/asset-manager$ tree uploads/
uploads/
├── assets
│   └── 55
│       └── f7
│           └── 5a55f73c759b74438e3806bd
│               └── tmp.txt
└── tmp

5 directories, 1 file
```

### After upload to S3

```
$ curl http://asset-manager.dev.gov.uk/assets/5a55f73c759b74438e3806bd
{"_response_info":{"status":"ok"},"id":"http://asset-manager.dev.gov.uk/assets/5a55f73c759b74438e3806bd","name":"tmp.txt","content_type":"text/plain","file_url":"http://assets-origin.dev.gov.uk/media/5a55f73c759b74438e3806bd/tmp.txt","state":"uploaded"}

$ curl http://assets-origin.dev.gov.uk/media/5a55f73c759b74438e3806bd/tmp.txt
Wed 10 Jan 2018 11:19:54 GMT
```

```
~/Code/alphagov/asset-manager$ tree uploads/
uploads/
└── tmp

1 directory, 0 files
```

## Update existing asset

```
$ echo `date` > tmp.txt
$ cat tmp.txt 
Wed 10 Jan 2018 11:23:00 GMT
$ curl http://asset-manager.dev.gov.uk/assets/5a55f73c759b74438e3806bd/ --request PUT --form "asset[file]=@tmp.txt"
{"_response_info":{"status":"success"},"id":"http://asset-manager.dev.gov.uk/assets/5a55f73c759b74438e3806bd","name":"tmp.txt","content_type":"text/plain","file_url":"http://assets-origin.dev.gov.uk/media/5a55f73c759b74438e3806bd/tmp.txt","state":"unscanned"}
```

### Before upload to S3

```
~/Code/alphagov/asset-manager$ tree uploads/
uploads/
├── assets
│   └── 55
│       └── f7
│           └── 5a55f73c759b74438e3806bd
│               └── tmp.txt
└── tmp

5 directories, 1 file
```

### After upload to S3

```
$ curl http://asset-manager.dev.gov.uk/assets/5a55f73c759b74438e3806bd
{"_response_info":{"status":"ok"},"id":"http://asset-manager.dev.gov.uk/assets/5a55f73c759b74438e3806bd","name":"tmp.txt","content_type":"text/plain","file_url":"http://assets-origin.dev.gov.uk/media/5a55f73c759b74438e3806bd/tmp.txt","state":"uploaded"}

$ curl http://assets-origin.dev.gov.uk/media/5a55f73c759b74438e3806bd/tmp.txt
Wed 10 Jan 2018 11:23:00 GMT
```

```
~/Code/alphagov/asset-manager$ tree uploads/
uploads/
└── tmp

1 directory, 0 files
```

floehopper:
On the basis of @chrisroos' review and the tests above, I'm happy to merge this.

PR #393, opened 2018-01-10
WIP: Handle URI encoded legacy url paths

chrisroos:
We've found a number of Whitehall assets whose filenames (and therefore their `legacy_url_path`s) contain non-ascii characters. These non-ascii characters cause problems when used in URLs to request Whitehall assets because the resulting URL is invalid.

I've updated the `AssetPresenter` to encode the asset URI so that we're presenting valid URLs to the world.

I've updated `WhitehallAssetsController#show` to decode the `legacy_url_path`. This is in preparation for updating the gds-api-adapters to URI encode the legacy_url_path when making a request for a Whitehall asset.

### DIFF
--- a/app/controllers/whitehall_assets_controller.rb
+++ b/app/controllers/whitehall_assets_controller.rb
@@ -16,7 +16,7 @@ class WhitehallAssetsController < ApplicationController
 
   def show
     path = "/#{params[:path]}.#{params[:format]}"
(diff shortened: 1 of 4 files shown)


chrisroos:
I've marked this as WIP. Rubocop has suggested I use `CGI.escape` instead of `URI.escape` but making that change appears to break other tests!

chrisroos:
I'm closing this as it's now been superseded by PR #396.

Issue #395, opened 2018-01-11
URI::InvalidURIError when creating Whitehall assets with non-ascii characters in legacy_url_path


chrisroos:
https://sentry.io/govuk/app-asset-manager/issues/431066535/

Steps to replicate in development:

```
$ echo `date` > whitehall-assét.txt
$ curl http://asset-manager.dev.gov.uk/whitehall_assets \
  --form "asset[file]=@whitehall-assét.txt" \
  --form "asset[legacy_url_path]=/government/uploads/path/to/whitehall-assét.txt"
```

chrisroos:
I've opened PR #396 to address this issue.

chrisroos:
https://github.com/alphagov/asset-manager/pull/396 has just been merged to master.

chrisroos:
PR #396 was deployed to staging and production this morning.

chrisroos:
I've just tested the fix in integration and confirmed that the asset was created successfully and that we didn't see any exceptions in Sentry.

```
$ export BEARER_TOKEN=`cat /etc/govuk/manuals-publisher/env.d/ASSET_MANAGER_BEARER_TOKEN`

$ echo `date` > whitehall-assét.txt

$ curl https://asset-manager.integration.govuk-internal.digital/whitehall_assets \
--form "asset[file]=@whitehall-assét.txt" \
--form "asset[legacy_url_path]=/government/uploads/path/to/whitehall-assét.txt"  \
-H"Authorization: Bearer $BEARER_TOKEN" \
-H"Accept: application/json" 

{"_response_info":{"status":"created"},"id":"https://asset-manager.integration.govuk-internal.digital/assets/5a5cb53bcc6ebd6ae4f48d59","name":"whitehall-ass_t.txt","content_type":"text/plain","file_url":"
https://assets-origin.integration.publishing.service.gov.uk/government/uploads/path/to/whitehall-ass%C3%A9t.txt","state":"unscanned"}
```

chrisroos:
I've marked the exception as resolved in Sentry and am going to close this issue given that we're not expecting to see these issues anymore.

PR #396, opened 2018-01-11
Percent encode public_url_path in asset JSON

chrisroos:
This supersedes PR #393.

We've found a number of Whitehall assets whose filenames (and therefore
their `legacy_url_path`s) contain non-ascii characters. We can safely
store these non-ascii characters in the database but need to escape them
when constructing the `file_url` in the JSON representation of an asset.
This addresses the bug described in issue #395.

I initially used Ruby's `URI.escape` method but Rubocop doesn't like
that as it's been marked as obsolete.

After this change we can upload and request a Whitehall asset whose
legacy_url_path contains a non-ascii character:

```
$ echo `date` > whitehall-assét.txt

$ curl http://asset-manager.dev.gov.uk/whitehall_assets \
  --form "asset[file]=@whitehall-assét.txt" \
  --form "asset[legacy_url_path]=/government/uploads/path/to/whitehall-assét.txt"
{
  "_response_info":{"status":"created"},
  "id":"http://asset-manager.dev.gov.uk/assets/5a562f05759b745bcb180de6",
  "name":"whitehall-ass_t.txt",
  "content_type":"text/plain",
  "file_url":"http://assets-origin.dev.gov.uk/government/uploads/path/to/whitehall-ass%C3%A9t.txt",
  "state":"unscanned"
}

$ curl
http://asset-manager.dev.gov.uk/government/uploads/path/to/whitehall-ass%C3%A9t.txt
Wed Jan 10 15:18:02 UTC 2018
```

Note that the version of CarrierWave we're using (0.11.x) doesn't
support non-ascii characters in uploaded filenames which is why the
`name` in the example JSON above has replaced the 'é' with an
underscore.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,6 @@
 source 'https://rubygems.org'
 
+gem 'addressable'
 gem 'aws-sdk', '~> 2.10.10'
 gem 'carrierwave', '~> 0.11.2'
 gem 'carrierwave-mongoid', '~> 0.10.0', require: 'carrierwave/mongoid'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -39,6 +39,8 @@ GEM
       i18n (~> 0.7)
       minitest (~> 5.1)
       tzinfo (~> 1.1)
+    addressable (2.5.2)
+      public_suffix (>= 2.0.2, < 4.0)
     arel (8.0.0)
     ast (2.3.0)
     aws-sdk (2.10.105)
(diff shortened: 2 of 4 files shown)


chrisroos:
Thanks for the review, @floehopper. I've just force pushed the branch to force a build of the publishing-end-to-end tests. They failed last time for apparently unrelated reasons.

chrisroos:
The publishing-end-to-end tests have now passed so I'm going to get this merged.

PR #407, opened 2018-01-12
Include mongo & redis checks in healthcheck action

floehopper:
This was motivated by the recent discovery that Asset Manager wasn't properly configured to connect to Redis in the new AWS integration environment. If this healthcheck had been present, we would've been alerted to the problem much sooner.

This is modelled closely on the code in the [email-alert-api repo][1]. For now I've only included checks for mongo & redis connectivity. We can always add further checks later.

* I've included the logic for warnings as well as critical errors even though the checks I've included only report critical errors.
* I've included the logic for reporting details of the warnings/errors even though the checks I've included don't report any details.
* I had to make some changes to the code to fix some Rubocop RSpec violations, e.g. introduce a base class healthcheck, use named subjects & instance doubles in the specs, allow injection of healthchecks into the top-level Healthcheck class, etc.
* I couldn't find an equivalent of ActiveRecord::Base.connected?, so I ended up calling Asset.count and rescuing the relevant exception. Note that this may take some time to timeout if MongoDB isn't available - 30 secs in the development environment.

[1]: https://github.com/alphagov/email-alert-api

### DIFF
--- a/app/controllers/healthcheck_controller.rb
+++ b/app/controllers/healthcheck_controller.rb
@@ -1,0 +1,13 @@
+class HealthcheckController < ApplicationController
+  skip_before_action :authenticate_user!
+
+  def check
+    render json: healthcheck.details.merge(status: healthcheck.status)
+  end
+
+private
+
+  def healthcheck
+    @healthcheck ||= Healthcheck.build
+  end
+end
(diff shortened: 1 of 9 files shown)


floehopper:
@chrislo Thanks!

Issue #417, opened 2018-01-18
.odt files aren't served with the correct mimetype


boffbowsh:
The .odt file linked from https://www.gov.uk/government/consultations/great-western-rail-franchise 
(https://assets.publishing.service.gov.uk/government/uploads/system/uploads/consultation_response_form_data/file/629/great-western-consultation.odt) is served with `Content-Type: application/octet-stream`. This is causing some combinations of Windows and IE, particularly at DfT, to download the file with a .zip extension. Users are then not able to open the file without renaming to .odt.

chrisroos:
Thanks @boffbowsh. I suspect this might be related to "Investigate content-type difference between Whitehall and Asset Manager #408".

chrisroos:
I'm going to investigate this as the problem will only get worse as we migrate more Whitehall assets to Asset Manager.

chrisroos:
The difference in behaviour between Whitehall and Asset Manager is caused by the [additional mime types that have been added to Whitehall](https://github.com/alphagov/whitehall/blob/master/config/initializers/mime_types.rb).

We can observe the problem by comparing an Asset Manager Rails console to a Whitehall Rails console:

```
# Asset Manager
> Mime::Type.lookup_by_extension('odt')
=> nil

# Whitehall
> Mime::Type.lookup_by_extension('odt')
=> #<Mime::Type:0x0000556910e84a58 @hash=1499320519752563818, @string="application/vnd.oasis.opendocument.text", @symbol=:odt, @synonyms=[]>
```

chrisroos:
I've opened https://github.com/alphagov/asset-manager/pull/418 to address this issue.

chrisroos:
I've merged https://github.com/alphagov/asset-manager/pull/418 so I'll test the effect on integration once it's been deployed.

chrisroos:
I've tested this in integration by uploading a .odt file and requesting it using curl:

```
$ curl -v "https://assets-origin.integration.publishing.service.gov.uk/media/5a61e8f15f743b5834d6087b/great-western-consultation.odt" > /dev/null

> GET /media/5a61e8f15f743b5834d6087b/great-western-consultation.odt HTTP/2
> Host: assets-origin.integration.publishing.service.gov.uk
> User-Agent: curl/7.54.0
> Accept: */*
> 
* Connection state changed (MAX_CONCURRENT_STREAMS updated)!
< HTTP/2 200 
[...]

Issue #408, opened 2018-01-12
Investigate content-type difference between Whitehall and Asset Manager


chrisroos:
I noticed a difference in the `content-type` of a `.docx` asset when requesting it from Whitehall and Asset Manager as part of issue #297:

## Requesting from Whitehall
```
$ curl -v "https://assets-origin.integration.publishing.service.gov.uk/government/uploads/system/uploads/consultation_response_form_data/file/622/revised-draft-airports-nps-response-form.docx" > /dev/null

* Connection state changed (MAX_CONCURRENT_STREAMS updated)!
< HTTP/2 200
< content-type: application/vnd.openxmlformats-officedocument.wordprocessingml.document
```

## When from Asset Manager
```
$ curl -v "https://assets-origin.integration.publishing.service.gov.uk/government/uploads/system/uploads/consultation_response_form_data/file/622/revised-draft-airports-nps-response-form.docx" > /dev/null

* Connection state changed (MAX_CONCURRENT_STREAMS updated)!
< HTTP/2 200
< content-type: application/octet-stream
```

chrisroos:
I suspect this might be related to ".odt files aren't served with the correct mimetype #417".

chrisroos:
I've opened https://github.com/alphagov/asset-manager/pull/418 to address this issue.

chrisroos:
I'm closing this in favour of #417 as that's a more specific example of this more general problem.

PR #418, opened 2018-01-19
Ensure we support content types of whitehall assets

chrisroos:
This addresses https://github.com/alphagov/asset-manager/issues/417 and https://github.com/alphagov/asset-manager/issues/408.

I've ensured we're setting the correct `Content-Type` header for assets allowed by Whitehall's non-attachments assets that we're in the process of migrating to Asset Manager.

### DIFF
--- a/config/initializers/mime_types.rb
+++ b/config/initializers/mime_types.rb
@@ -2,3 +2,14 @@
 
 # Add new mime types for use in respond_to blocks:
 # Mime::Type.register "text/richtext", :rtf
+
+Mime::Type.register "text/rtf", :rtf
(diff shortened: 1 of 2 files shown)


chrisroos:
> Will this update the Content-Types for existing assets, or will we need to do a bulk update of assets already in S3?

We're not setting the `Content-Type` on the objects in S3 so there's nothing to update. This fix should ensure we're setting the `Content-Type` correctly as soon as it's deployed.

chrisroos:
Thanks for the review, @floehopper. I'll get this merged now.

Issue #156, opened 2017-08-24
Consider fixing or removing the private Asset Manager host alias & associated functionality


floehopper:
* [This pull request][1] (in combination with [this follow-up PR][2]) suggests the intention was for the Asset Manager application to be proxied via two different nginx vhosts:
  * `asset-manager.*`
  * `private-asset-manager.*`
* Requests via the `private-asset-manager` vhost require authentication via SSO and that the user is authorized to use the Asset Manager app (via [`GDS::SSO::ControllerMethods#require_signin_permission!`][3]).
* The private vhost is [configured as a vhost alias by Puppet][4].
* The `PRIVATE_ASSET_MANAGER_HOST` environment variable is [provided to the Asset Manager app by Puppet][5].
[This commit note][6] suggests that the intention was for the Asset Manager app to be the only app to know about the environment variable, because it hands out the asset URLs from its API.
* However, the host for the URLs returned by the Asset Manager API is [obtained from `Plek#asset_root`][7] whether or not the asset is marked as `access_limited`.
* If an asset is marked as `access_limited` the Asset Manager tries to check the user's organisation matches that of the asset.
  * **Non-private requests:**
    * The user will *not* be forced to sign in.
    * The authorization check will *not* happen, i.e. any SSO user will work.
    * Only allows access if the user is *already* signed in to SSO and their organisation matches that of the asset.
  * **Private requests:**
    * The user will be forced to sign in.
    * The SSO user must be authorized to access Asset Manager app.
    * Allows access if the user's organisation matches that of the asset.
* Note that `PRIVATE_ASSET_MANAGER_HOST` is set to `private-asset-manager.integration.publishing.service.gov.uk` in integration (see `/etc/govuk/asset-manager/env.d/PRIVATE_ASSET_MANAGER_HOST`) and while `asset-manager.integration.publishing.service.gov.uk` resolves to `backend-internal-lb.backend.integration.publishing.service.gov.uk` (via `/etc/hosts`), `private-asset-manager.integration.publishing.service.gov.uk` doesn't seem to resolve to anything. So it's not obvious to me that using this private hostname from a publishing app would actually work (e.g. the Manuals Publisher app runs on the same backend machine(s) in integration).

We should either make this work or remove the relevant functionality & configuration. Note that all my investigation so far has been by *reading code* and not by experimenting , e.g. on the developer VM.

[1]: https://github.com/alphagov/asset-manager/pull/23
[...]

PR #426, opened 2018-01-26
Remove unused access-limiting functionality

floehopper:
We're about to add functionality to the Asset Manager to protect draft assets behind SSO and this unused functionality just confuses matters. Note that this PR probably also addresses the concerns detailed in https://github.com/alphagov/asset-manager/issues/156.

### Remove authentication for private virtual host

Although in govuk-puppet [the `PRIVATE_ASSET_MANAGER_HOST` environment variable is set to `private-asset-manager.${app_domain}`][1] and [`private-asset-manager` is added as an alias to the `asset-manager` virtual host][2], the Asset Manager app never supplies URLs using this hostname to publishing apps and I haven't found any evidence that any of the publishing apps manually construct URLs using this hostname. So as far as I can tell this functionality is not currently (and possibly has never been) used.

We're about to add functionality to the Asset Manager to protect draft assets behind SSO and this unused functionality just confused matters.

I plan to remove the related code from govuk-puppet in a PR on that repo.

### Remove access-limited functionality for Mainstream assets

At some point it was possible to mark an asset as access-limited. However, since [the upgrade to Rails v4][3] this will have been prevented by [Rails' Strong Parameter functionality][4]. This won't have been noticed, because there are no specs testing that it's possible to supply the `access_limited` or `organisation_slug` parameters to the API endpoints for creating or updating an asset. If either of these parameters are supplied to these API endpoints, the parameters are stripped from the attributes used in the mass assignment and the a warning like the following is logged:

    Unpermitted parameters: :access_limited, :organisation_slug

I confirmed that there are currently no assets in production with either of these fields set by asking for the following queries to be run in the Asset Manager Rails console:

    irb> Asset.unscoped.where(access_limited: nil).count
    => 0
    irb> Asset.unscoped.where(access_limited: false).count
    => 1100658
    irb> Asset.unscoped.where(access_limited: true).count
    => 0

    irb> Asset.unscoped.where(organisation_slug: nil).count
    => 1100658
    irb> Asset.unscoped.where(:organisation_slug.ne =>
    nil).count
    => 0

### Tidy up

The rest of the PR just removes the now redundant code and database fields. It adds a new Rake task to achieve the latter.

[1]:
https://github.com/alphagov/govuk-puppet/blob/9a4ad116d8b495690415405e6ee3b0663959d06d/modules/govuk/manifests/apps/asset_manager.pp#L69-L74
[2]:
https://github.com/alphagov/govuk-puppet/blob/9a4ad116d8b495690415405e6ee3b0663959d06d/modules/govuk/manifests/apps/asset_manager.pp#L172
[3]: https://github.com/alphagov/asset-manager/pull/42
[4]:
http://guides.rubyonrails.org/v5.1.4/action_controller_overview.html#strong-parameters

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -28,7 +28,6 @@ protected
   def asset_servable?
     asset.filename_valid?(params[:filename]) &&
       asset.uploaded? &&
-      asset.accessible_by?(current_user) &&
       asset.mainstream?
   end
 
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -1,6 +1,4 @@
 class MediaController < BaseMediaController
-  before_action :authenticate_if_private
-
   def download
     unless asset_servable?
       error_404
@@ -34,16 +32,8 @@ protected
       asset.mainstream?
   end
 
-  def authenticate_if_private
-    authenticate_user! if requested_via_private_vhost?
-  end
-
   def asset
     @asset ||= Asset.find(params[:id])
-  end
-
-  def requested_via_private_vhost?
-    request.host == ENV['PRIVATE_ASSET_MANAGER_HOST']
   end
 
   def redirect_to_current_filename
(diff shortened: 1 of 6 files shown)


chrisroos:
I'll review this now.

Issue #432, opened 2018-01-26
Authenticate requests for draft assets


floehopper:
* Immediately authenticate requests that arrive via `draft-asset` virtual host

floehopper:
I've opened https://github.com/alphagov/asset-manager/pull/438 to address this.

floehopper:
I belatedly realised that #438 was overly simplistic and non-standard in its detection of the `draft-assets` domain. I've opened https://github.com/alphagov/govuk-puppet/pull/7173 to make `Plek#find` return the appropriate value for 'draft-assets'  and I plan to update #438 to make use of this. I've marked #438 as "Do not merge" in the meantime.

floehopper:
On the advice of @cbaines, I'm going to use the new `Plek` method introduced in https://github.com/alphagov/plek/pull/55 (supported by https://github.com/alphagov/govuk-puppet/pull/7107) instead of https://github.com/alphagov/govuk-puppet/pull/7173.

floehopper:
Closing in favour of #444.

floehopper:
Oops. I meant to close PR #438 not this issue. Re-opening.

floehopper:
I've opened #444 in place of #438.

floehopper:
I've merged https://github.com/alphagov/asset-manager/pull/444.

floehopper:
I've just tested this on my development VM as follows:

* Run `GDS_SSO_STRATEGY=real bowl asset-manager signon`

```
$ echo `date` > tmp.txt

$ cat tmp.txt 
Tue 30 Jan 2018 18:56:35 GMT

$ curl http://asset-manager.dev.gov.uk/assets --form "asset[file]=@tmp.txt" --form "asset[draft]=1"
{"_response_info":{"status":"created"},"id":"http://asset-manager.dev.gov.uk/assets/5a70bffe759b7464e73086ef","name":"tmp.txt","content_type":"text/plain","file_url":"http://assets-origin.dev.gov.uk/media/5a70bffe759b7464e73086ef/tmp.txt","state":"unscanned","draft":true}

$ curl -s http://asset-manager.dev.gov.uk/assets/5a70bffe759b7464e73086ef | jq '.state'
"uploaded"

$ curl -s http://asset-manager.dev.gov.uk/assets/5a70bffe759b7464e73086ef | jq '.draft'
true

$ curl -s http://asset-manager.dev.gov.uk/assets/5a70bffe759b7464e73086ef | jq '.file_url'
[...]

Issue #431, opened 2018-01-26
Allow assets to be marked as draft/published via API


floehopper:
* Change Asset Manager API to allow assets to be marked as draft and later marked as published
* API show action should report draft/published status of asset

floehopper:
I've opened https://github.com/alphagov/asset-manager/pull/435 to address this.

floehopper:
https://github.com/alphagov/asset-manager/pull/435 has been merged.

floehopper:
I've just confirmed that this is working on integration as follows:

### Create draft asset

```
$ echo `date` > tmp.txt

$ govuk_setenv manuals-publisher bash

$ curl \
>   -H"Authorization: Bearer $ASSET_MANAGER_BEARER_TOKEN" \
>   -H"Accept: application/json" \
>   https://asset-manager.$GOVUK_APP_DOMAIN/assets \
>   --form "asset[file]=@tmp.txt" --form "asset[draft]=1"
{"_response_info":{"status":"created"},"id":"https://asset-manager.integration.govuk-internal.digital/assets/5a704b225d7f095f67d1239c","name":"tmp.txt","content_type":"text/plain","file_url":"https://assets-origin.integration.publishing.service.gov.uk/media/5a704b225d7f095f67d1239c/tmp.txt","state":"unscanned","draft":true}

$ curl -s https://asset-manager.integration.govuk-internal.digital/assets/5a704b225d7f095f67d1239c -H"Authorization: Bearer $ASSET_MANAGER_BEARER_TOKEN" -H"Accept: application/json" | jq '.draft'
true
```

### Update asset to not be draft

```
$ curl \
> https://asset-manager.integration.govuk-internal.digital/assets/5a704b225d7f095f67d1239c \
> -H"Authorization: Bearer $ASSET_MANAGER_BEARER_TOKEN" \
> -H"Accept: application/json" \
> --request PUT --form "asset[draft]=0"
{"_response_info":{"status":"success"},"id":"https://asset-manager.integration.govuk-internal.digital/assets/5a704b225d7f095f67d1239c","name":"tmp.txt","content_type":"text/plain","file_url":"https://assets-origin.integration.publishing.service.gov.uk/media/5a704b225d7f095f67d1239c/tmp.txt","state":"uploaded","draft":false}

[...]

PR #435, opened 2018-01-26
Allow assets to be marked as draft

floehopper:
We want to migrate Whitehall attachments to Asset Manager, but Whitehall currently protects draft attachments behind SSO, so before we can do the migration we need to implement similar functionality in Asset Manager. This PR is the first small step towards doing that - it makes it possible to mark an asset as draft when calling the "create" & "update" API actions. It includes the draft status of an asset in the response to those API calls and the "show" API call.

See #431.

### Notes

* For the moment, public requests for assets marked as draft are treated the same as those for assets not marked as draft. I plan to introduce authentication, etc, as a separate PR when addressing #432.

* I've consciously chosen not to update the documentation for the moment since this is somewhat experimental functionality and I don't want to encourage people to use it yet.

* We may need to add a Whitehall-specific "update" API action to allow the Whitehall app to mark a draft Whitehall asset as draft using the `legacy_url_path` rather than the `id`, but I think that's better addressed in a separate PR if/when we know we definitely need it. This PR should be enough for us to try out the new functionality using tools like `curl`.

### DIFF
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -55,6 +55,6 @@ private
   end
 
   def asset_params
-    params.require(:asset).permit(:file)
+    params.require(:asset).permit(:file, :draft)
   end
 end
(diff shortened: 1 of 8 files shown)


floehopper:
Rebasing against master, resolving conflicts, and force-pushing in preparation for merging.

PR #438, opened 2018-01-29
[Do not merge] Require SSO authentication for requests from draft-assets host

floehopper:
This makes draft assets available to authenticated users via the new `draft-assets` host. Requests from non-authenticated users will be rejected. Requests from the existing `assets-origin` host will continue not to require authentication.

Note that by authenticating *before* we've made any queries to the database we can reduce the chances of a DOS attack causing problems.

Note also that this change applies to both Mainstream & Whitehall assets although we're only really interested in Whitehall assets at the moment.


floehopper:
Closing in favour of #444.

PR #444, opened 2018-01-30
Require authentication for requests from draft-assets host

floehopper:
The main change in this PR is in the last commit which means that any request from the `draft-assets` host requires SSO authentication.

In a future PR we plan to redirect requests from the
`assets-origin` host for assets marked as draft to the `draft-assets` host.

Note that this change applies to both Mainstream & Whitehall assets.

Note also that we're using the new `Plek#external_url_for` method to obtain the external version of the `draft-assets` host which should mean the functionality works in AWS environments (currently only integration).

Supersedes #432.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -179,7 +179,7 @@ GEM
     parallel (1.12.1)
     parser (2.4.0.2)
       ast (~> 2.3)
-    plek (2.0.0)
+    plek (2.1.0)
     powerpack (0.1.1)
     public_suffix (3.0.1)
     rack (2.0.3)
--- a/app/controllers/base_media_controller.rb
+++ b/app/controllers/base_media_controller.rb
@@ -1,7 +1,11 @@
 class BaseMediaController < ApplicationController
-  skip_before_action :authenticate_user!
+  skip_before_action :authenticate_user!, unless: :requested_from_draft_assets_host?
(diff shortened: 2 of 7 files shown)


chrisroos:
I'm going to review this now.

floehopper:
@chrisroos I've just noticed the build errors on this branch - I'm investigating them now.

floehopper:
I've asked @cbaines whether we've missed a place where `GOVUK_APP_DOMAIN_EXTERNAL` needs setting in order to be available on Jenkins CI.

floehopper:
It turns out the problem is that the env vars are set manually for the Jenkins CI build for the [`precompileAssets` step][1].

[1]: https://github.com/alphagov/govuk-puppet/blob/516885e2866dd3e7b1c381773bc2e2db205aa6e9/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy#L569

floehopper:
We're hoping that https://github.com/alphagov/publishing-e2e-tests/pull/171 which has just been merged will fix the e2e build for this branch.

floehopper:
I've kicked off another e2e build...

floehopper:
Force-pushed an "amended" commit to trigger new builds.

PR #446, opened 2018-01-30
Redirect requests for assets marked as draft to draft-assets host

floehopper:
The significant changes are in the final commit of this PR which redirects requests for assets marked as draft to the `draft-assets` host if they have come from another host, typically `assets-origin`.

Assets marked as draft should only be accessed via the draft-assets host which requires authentication as of #444.

Note that these changes apply to both Mainstream & Whitehall assets.

Although I've managed to extract the common functionality into `BaseMediaController#redirected_to_draft_assets_host_for?`, I've ended up adding examples to the specs for both `MediaController` & `WhitehallMediaController`, because I found it awkward to add them to `BaseMediaController` spec.

I'm not entirely happy with the resulting code, but it's the best compromise I could come up with. Hopefully as we bring the two media controllers into line with each other all this code can be simplified.

In preparation for this PR, I've done some work to bring the two media controller a bit more into line with each other. While this wasn't strictly necessary, I found it made it easier to see where to make the changes.

### DIFF
--- a/app/controllers/base_media_controller.rb
+++ b/app/controllers/base_media_controller.rb
@@ -19,4 +19,10 @@ protected
 
     head :ok, content_type: asset.content_type
   end
+
+  def redirected_to_draft_assets_host_for?(asset)
+    if asset.draft? && !requested_from_draft_assets_host?
+      redirect_to host: AssetManager.govuk.draft_assets_host, format: params[:format]
+    end
+  end
 end
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -1,5 +1,9 @@
 class MediaController < BaseMediaController
   def download
+    if redirected_to_draft_assets_host_for?(asset)
+      return
+    end
+
     unless asset_servable?
       error_404
       return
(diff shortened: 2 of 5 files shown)


app/controllers/base_media_controller.rb
```diff
@@ -19,4 +19,10 @@ def proxy_to_s3_via_nginx(asset)
 
     head :ok, content_type: asset.content_type
   end
+
+  def redirected_to_draft_assets_host_for?(asset)
```
chrisroos: I find this method a bit odd but I suggest we go with it and change it later if/when we revisit this code.

floehopper: Yeah. I'm not super happy with it either. The method needs to return a value so the controller sub-class can decide whether to `return` or not, otherwise we end up with more duplication in the controller sub-classes. And passing the `asset` in as a parameter matches how `#proxy_to_s3_via_nginx` works. With both those in mind I found it hard to come up with a good name for the method.

floehopper: @chrisroos I've addressed your comment in a new PR: https://github.com/alphagov/asset-manager/pull/452

PR #452, opened 2018-01-31
Standardize code for redirecting to draft-assets host

floehopper:
The previous implementation was non-standard and therefore confusing. In particular, the
`BaseMediaController#redirected_to_draft_assets_host_for?` method looked like a query method, but was actually a command method, i.e. it had side effects which might have been surprising. By splitting this method into a query method and a command method, we can avoid this confusion at the expense of a little duplication.

Addresses [this comment in an earlier PR](https://github.com/alphagov/asset-manager/pull/446#pullrequestreview-92674337).

### DIFF
--- a/app/controllers/base_media_controller.rb
+++ b/app/controllers/base_media_controller.rb
@@ -20,9 +20,11 @@ protected
     head :ok, content_type: asset.content_type
   end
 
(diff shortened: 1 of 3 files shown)


floehopper:
I've kicked the e2e CI build, because the error looks unrelated and possibly an intermittent error.

PR #456, opened 2018-02-01
Refactor assets controllers

floehopper:
I started work on this, because I wanted to add an `update` action to the `WhitehallAssetsController`, but I was painfully aware that we already had a bunch of duplication. However, having chatted to @chrisroos I've now convinced myself that we don't need this `update` action - we can just use a combination of `WhitehallAssetsController#show` & `AssetsController#update`.

I think some or all of the following work is still useful, because I fixed a couple of things and removed a bunch of duplication. However, I'm happy if people think it's not useful.

The fixes are in these commits:

* "Fix asset lookup in WhitehallAssetsController#show with no format"
* "Remove duplicate examples from AssetsController spec"
* "Move context into describe block in AssetsController spec"

The other commits are mostly about removing duplication or improving consistency.

Aside: I did start looking at adding a `WhitehallAssetsController#update` action and one thing I wondered was whether it might be worth making the `path` used in the `WhitehallAssetsController` routes into a proper `id` parameter, i.e. having `gds-api-adapters` encode the path so that it can cope with the forward-slashes. I think this would mean we could make `whitehall_assets` into a standard Rails resource, i.e. use resource-ful routes for all actions. However, since I decided not to add this action (see above), I didn't end up making this change.

### DIFF
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -1,16 +1,6 @@
 class AssetsController < BaseAssetsController
   before_action :restrict_request_format
 
-  def create
-    @asset = build_asset
-
-    if @asset.save
-      render json: AssetPresenter.new(@asset, view_context).as_json(status: :created), status: :created
-    else
-      error 422, @asset.errors.full_messages
-    end
-  end
-
   def update
     @asset = find_asset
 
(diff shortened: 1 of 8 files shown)


chrisroos:
I'm going to review this now.

floehopper:
@chrisroos Thanks! I'll get these changes merged.

PR #457, opened 2018-02-02
Improvements to Sidekiq workers

floehopper:
The main purpose of this PR is to make the `SaveToCloudStorageWorker` & `VirusScanWorker` jobs idempotent. These are the only two jobs used in standard operation, the only other one, `DeleteAssetFileFromNfsWorker`, has served its purpose and will soon be removed.

My motivation for making this change is that on a couple of occasions I've seen these jobs apparently fail, because the underlying file no longer exists. And on at least one occasion the state of the asset has still been changed to 'uploaded' and the asset appears to have been saved to the cloud. This suggests that the job may have executed more than once. If that is indeed what happened, these changes should avoid that happening again.

Also [the Sidekiq documentation recommends ensuring jobs are idempotent][1].

As part of this PR, I've also further constrained the asset state machine transitions. Hopefully this should make the code easier to reason about. I've also done some incidental tidying up in order to make my changes easier to apply.

[1]:
https://github.com/mperham/sidekiq/wiki/Best-Practices#2-make-your-job-idempotent-and-transactional

### DIFF
--- a/spec/models/asset_spec.rb
+++ b/spec/models/asset_spec.rb
@@ -187,16 +187,94 @@ RSpec.describe Asset, type: :model do
   end
 
   describe "when an asset is marked as clean" do
-    let(:asset) { FactoryBot.build(:asset) }
+    let(:state) { 'unscanned' }
+    let(:asset) { FactoryBot.build(:asset, state: state) }
 
     before do
       allow(SaveToCloudStorageWorker).to receive(:perform_async)
+    end
+
+    it 'sets the asset state to clean' do
+      asset.scanned_clean!
+
+      expect(asset.reload).to be_clean
     end
 
     it 'schedules saving the asset to cloud storage' do
       expect(SaveToCloudStorageWorker).to receive(:perform_async).with(asset.id)
 
       asset.scanned_clean!
+    end
+
+    context 'when asset is already clean' do
+      let(:state) { 'clean' }
+
+      it 'does not allow the state transition' do
+        expect { asset.scanned_clean! }
+          .to raise_error(StateMachines::InvalidTransition)
+      end
+    end
+
+    context 'when asset is already infected' do
+      let(:state) { 'infected' }
+
+      it 'does not allow the state transition' do
+        expect { asset.scanned_clean! }
+          .to raise_error(StateMachines::InvalidTransition)
+      end
+    end
+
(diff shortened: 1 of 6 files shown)


chrisroos:
I'm going to take a look at this now.

spec/models/asset_spec.rb
```diff
@@ -198,6 +199,79 @@
...
+          .to raise_error(StateMachines::InvalidTransition)
+      end
+    end
+
+    context 'when asset is already uploaded' do
+      let(:state) { 'uploaded' }
+
+      # before do
+      #   asset.file = load_fixture_file('asset.png')
+      # end
```
chrisroos: It looks as though you can remove this commented out `before` block.

spec/models/asset_spec.rb
```diff
@@ -198,6 +199,79 @@
...
```
chrisroos: It looks as though this commented out `before` block can be removed.

spec/models/asset_spec.rb
```diff
@@ -198,6 +199,79 @@
...
+
+      it 'does not allow the state transition' do
+        expect { asset.scanned_clean! }
+          .to raise_error(StateMachines::InvalidTransition)
+      end
+    end
+  end
+
+  describe 'when an asset is marked as infected' do
```
chrisroos: There doesn't appear to be a test within this `describe` block that asserts that the asset is transitioned to the `infected` state. Is that tested elsewhere?

spec/models/asset_spec.rb
floehopper: @chrisroos Well spotted - thanks!

spec/models/asset_spec.rb
floehopper: @chrisroos Well spotted - thanks!

spec/models/asset_spec.rb
floehopper: @chrisroos:

> Is that tested elsewhere?

No, it doesn't look like it. I guess I was focussing on testing the effects of the extra constraints. However, I've now added an example in both this and the 'when an asset is marked as clean' `describe` block for completeness.

floehopper:
I've rebased against `master`, fixed the issues identified by @chrisroos, and am about to force-push in preparation for merging.

PR #364, opened 2017-12-20
Introduce "uploaded" state for assets

floehopper:
This supersedes #363.

This introduces a new "uploaded" asset state. New assets are still scanned for viruses and if no viruses are found they are marked as "clean". However, now they will be marked as "uploaded" or if the subsequent upload to cloud storage is successful.

For the moment, assets in the "clean" and "uploaded" states are treated the same. I've made use of shared examples in the two media controller specs to check this without introducing duplication. However, once the `govuk:update_state_for_clean_assets_with_s3_object_to_uploaded` Rake task included in this PR has been run, we'll be in a position to treat these states differently and at that point we should be able to do away with the shared examples.

The motivation behind this PR is two-fold:

1. Now that all asset requests are proxied to S3, we shouldn't attempt to make an asset available via the media controllers until the asset has been successfully uploaded to S3. See #281 for details.

2. We want to remove files from NFS once they have been uploaded to S3. Having these new separate asset states should make this easier to reason about. See #323 for details.

Differences compared to #363 (mostly addressing comments from @chrisroos):

* Only add a single "uploaded" state; not "not_uploaded" as well. Failed uploads leave the asset in the "clean" state.
* Stricter constraints on state transition; only allow transition from "clean" to "uploaded".
* No need for generic `CloudStorage` exception base classes, because no explicit rescue block needed in `SaveToCloudStorageWorker.perform`.
* Include deleted assets in `govuk:update_state_for_clean_assets_with_s3_object_to_uploaded` Rake task.
* Better error messages in `S3Storage::ObjectUploadFailedError` exceptions.
* Extended/more representative end-to-end example in `spec/requests/virus_scanning_spec.rb`.

PR #458, opened 2018-02-05
Prevent caching of API endpoint for getting asset details

floehopper:
This changes the `AssetsController#show` and `WhitehallAssetsController#show` endpoints so that the `Cache-Control` header is always set to have `max-age=0`, i.e. indicating that the response should not be cached.

This logic became out-of-date when we introduced the uploaded state in [this PR][1]. I considered only updating the logic to fix this, i.e. by setting cache expiry to 0 for 'uploaded' or 'infected' assets (vs 'scanned' or 'infected' assets).

However, I realised that since we introduced the `draft` status of an asset in [this PR][2] and included it in the JSON response, it no longer makes sense to set the cache expiry to 0 for uploaded assets, because their `draft` status may change, even after they have been virus scanned and uploaded cloud storage.

Since very few assets are marked as infected, few (if any) of the publishing apps use this endpoint, and it isn't a very "expensive" request, so I decided it was better to simplify the app by always setting the cache expiry to 0.

Note that the `gds-api-adapters` gem used by the publishing apps [does respect the `Cache-Control` header][3] and so it is important to set the header appropriately.

[1]: https://github.com/alphagov/asset-manager/pull/364
[2]: https://github.com/alphagov/asset-manager/pull/435
[3]:
https://github.com/alphagov/gds-api-adapters/blob/5bdb71743ca94f11cad2756b37ab150a847ed53c/lib/gds_api/json_client.rb#L204-L241

### DIFF
--- a/spec/controllers/assets_controller_spec.rb
+++ b/spec/controllers/assets_controller_spec.rb
@@ -254,24 +254,10 @@ RSpec.describe AssetsController, type: :controller do
     describe "cache headers" do
       let(:asset) { FactoryBot.create(:asset) }
 
-      it "sets the cache-control headers to 0 for an unscanned asset" do
+      it "sets the cache-control headers to 0" do
         get :show, params: { id: asset.id }
 
         expect(response.headers["Cache-Control"]).to eq("max-age=0, public")
-      end
-
-      it "sets the cache-control headers to 30 minutes for a clean asset" do
-        asset.scanned_clean!
-        get :show, params: { id: asset.id }
-
-        expect(response.headers["Cache-Control"]).to eq("max-age=1800, public")
-      end
-
-      it "sets the cache-control headers to 30 minutes for an infected asset" do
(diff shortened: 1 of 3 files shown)


spec/controllers/assets_controller_spec.rb
```diff
@@ -254,25 +254,11 @@
     describe "cache headers" do
       let(:asset) { FactoryBot.create(:asset) }
 
-      it "sets the cache-control headers to 0 for an unscanned asset" do
+      it "sets the cache-control headers to 0" do
```
chrislo: s/headers/header

floehopper: Thanks. I've addressed this in [a new commit](https://github.com/alphagov/asset-manager/pull/458/commits/9e92fdebccb9c56f384e50f3b3d21a6525e7568a), because this wasn't something that I introduced.

PR #462, opened 2018-02-07
Allow location of clamscan to be configured by environment variable

adrianclay:
In the end to end tests environment, we want to be able to disable the virus scan easily as it can take more than 60 seconds to complete a scan when several requests are made to it at once.

By allowing the path to be passed in we can give it /bin/true when wanting to disable the scan.

### DIFF
--- a/lib/virus_scanner.rb
+++ b/lib/virus_scanner.rb
@@ -9,7 +9,8 @@ class VirusScanner
   class InfectedFile < StandardError; end
 
   def scan(file_path)
-    out_str, status = Open3.capture2e('govuk_clamscan', '--no-summary', file_path)
+    clamscan_path = AssetManager.govuk.clamscan_path
+    out_str, status = Open3.capture2e(clamscan_path, '--no-summary', file_path)
     case status.exitstatus
     when 0
       return true
--- a/spec/lib/govuk_configuration_spec.rb
+++ b/spec/lib/govuk_configuration_spec.rb
@@ -51,6 +51,28 @@ RSpec.describe GovukConfiguration do
     end
   end
 
+  describe '#clamscan_path' do
+    context 'when environment includes an ASSET_MANAGER_CLAMSCAN_PATH value' do
+      let(:env) {
+        {
+          'ASSET_MANAGER_CLAMSCAN_PATH' => 'alternative-path',
+        }
+      }
+
+      it 'returns environment variable' do
+        expect(config.clamscan_path).to eq('alternative-path')
+      end
+    end
+
+    context 'when environment does not include an ASSET_MANAGER_CLAMSCAN_PATH value' do
+      let(:env) { {} }
+
+      it 'returns govuk_clamscan' do
+        expect(config.clamscan_path).to eq('govuk_clamscan')
+      end
+    end
+  end
+
   describe '#draft_assets_host' do
     subject(:config) { described_class.new(env, plek) }
 
--- a/spec/lib/virus_scanner_spec.rb
+++ b/spec/lib/virus_scanner_spec.rb
@@ -1,4 +1,4 @@
-require 'spec_helper'
+require 'rails_helper'
 require 'virus_scanner'
 
 RSpec.describe VirusScanner do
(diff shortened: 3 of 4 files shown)


floehopper:
@adrianclay Did you consider either writing an empty `govuk_clamscan` file or a sym-link to `/bin/true` in the e2e test environment? It would be nice to avoid yet another e2e-specific environment variable option in the app.

floehopper:
@adrianclay If we are going to take this approach, I think it might be better to obtain this env var value in a Rails initializer - I'm pretty sure that's where most/all existing env var values are obtained. It could perhaps be added as a method on `GovukConfiguration` which would also mean you could add tests around it.

adrianclay:
Hey @floehopper we created an alternative here https://github.com/alphagov/asset-manager/compare/no-virus-scan-for-you

We preferred the ENV var solution because it makes the asset-manager more configurable as to if you want virus scan enabled or not within Docker. If you prefer that branch, we're happy to go with that.

floehopper:
@adrianclay:

> We preferred the ENV var solution because it makes the asset-manager more configurable as to if you want virus scan enabled or not within Docker.

Do you know if many people are currently using Docker to run individual apps? And/or is that the plan in the longer term? I currently still use my local machine and/or the development VM and in these use cases the ~10 secs it takes to scan a file doesn't cause me any issues.

> If you prefer that branch, we're happy to go with that.

I don't feel very strongly about it, but I am slightly wary about making the app more configurable if it's only needed in one use case.

@chrisroos, @chrislo: Do either of you have any thoughts about this?

chrislo:
I don't have particularly strong feelings, but I agree with @floehopper that less configuration is better - particularly if there's only a single use case.

RyanMacG:
@floehopper we've opened up #463 for the `/bin/true` implementation, sounds like this would be the best approach given your concerns.

floehopper:
@adrianclay, @RyanMacG:

I'm sorry for the slow response, I've been deep in some Whitehall code.

I've only just realised that the `Dockerfile` change would be in this repo (for some reason I had it in my head that it would be local to the e2e repo). As a consequence I'm now leaning towards the env var approach, because there's something nice about capturing the dependency on `clamav` in code (i.e. in the `Dockerfile`).

I'm sorry to have given you the run around, but would you be up for re-working **this** PR to add a method to `GovukConfiguration` and then accessing it via `AssetManager.govuk` c.f. `AssetManager.govuk.draft_assets_host`? That way you can add a test or two...

adrianclay:
I've made the requested change, ready for a rereview.

lib/virus_scanner.rb
```diff
@@ -9,7 +9,8 @@ class Error < StandardError; end
   class InfectedFile < StandardError; end
 
   def scan(file_path)
-    out_str, status = Open3.capture2e('govuk_clamscan', '--no-summary', file_path)
+    clamscan_path = AssetManager.govuk.clamscan_path
+    out_str, status = Open3.capture2e(clamscan_path, "--no-summary", file_path)
```
floehopper: Very minor: I think it would be better to keep the single quotes around `'--no-summary'`.

spec/lib/govuk_configuration_spec.rb
```diff
@@ -51,6 +51,28 @@
     end
   end
 
+  describe '#clamscan_path' do
+    context 'when environment includes an ASSET_MANAGER_CLAMSCAN_PATH value' do
+      let(:env) {
+        {
+          'ASSET_MANAGER_CLAMSCAN_PATH' => 'md5',
```
floehopper: Minor: The choice of 'md5' seems odd. I think it might be clearer to use something like 'alternative-path'.

spec/lib/virus_scanner_spec.rb
```diff
@@ -1,4 +1,4 @@
-require 'spec_helper'
+require 'rails_helper'
```
floehopper: Minor: I'd prefer this orthogonal change to be in a separate commit, but thanks for fixing it!

adrianclay:
I've updated the PR based on the comments.

PR #464, opened 2018-02-09
Configure GDS::SSO caching

cbaines:
The default is not to cache, and this means that large numbers of
requests are being made to Signon. Hopefully caching will reduce this.


tijmenb:
This was deployed on Friday and had the desired effect of reducing the number of requests to Signon from asset manager:

screen shot 2018-02-12 at 11 35 16

floehopper:
@tijmenb Thanks for checking it had the desired effect 👍

PR #471, opened 2018-02-12
Allow draft asset access to be restricted to specific users

chrisroos:
Certain documents in Whitehall can be marked as access limited; which means that they're only accessible to users within the same organisation(s) that the document belongs to. Any attachments belonging to those documents should be subject to the same restrictions. This branch adds functionality to allow access to draft assets to be restricted to a subset of users. Setting the `access_limited` attribute to an array of users' UIDs will allow _only_ those users to access the asset when its in draft.

This implementation is based on a cut-down version of the equivalent functionality in the Content Store; notably I'm not storing the `auth_bypass_ids` that it's possible to set on a `ContentItem` in the Content Store.

Note the additional duplication in the two media controllers. I imagine we'll DRY this up in future.

Note that I'm not emitting the access_limited data from the AssetPresenter. It doesn't feel quite right to emit this sensitive information.

### DIFF
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -38,7 +38,7 @@ private
   end
 
   def asset_params
-    params.require(:asset).permit(:file, :draft, :redirect_url)
+    params.require(:asset).permit(:file, :draft, :redirect_url, access_limited: [])
   end
 
   def find_asset(include_deleted: false)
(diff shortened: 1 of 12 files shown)


chrisroos:
Thanks for reviewing, @floehopper.

I've updated the commit note of the first commit to address your feedback about explaining the missing `auth_bypass_ids` and the use of an empty array to denote an asset being accessible to everyone.

PR #481, opened 2018-02-19
Add redirect_url to assets & redirect requests for Whitehall assets with redirect_url

floehopper:
The plan is for Whitehall to set `Asset#redirect_url` for an attachment if its parent document is unpublished. This relates to the logic in [Whitehall's `AttachmentController#fail`](https://github.com/alphagov/whitehall/blob/64bb168edf383fc62a7190f614f6822df9dff4a7/app/controllers/attachments_controller.rb#L43-L44) which determines whether the attachment's parent document has been unpublished.

This PR makes it possible to set a `redirect_url` attribute on any asset via the `update` API endpotin and to see its value via the `show` API endpoint. It also does the redirect if the attribute is set, but only for Whitehall assets.

In Asset Manager the concept of a "draft" asset means that requests for it must come via the `draft-assets` host and be associated with an authenticated user. Thus to avoid a double redirect and (more significantly) unnecessary authentication, Whitehall should not mark an asset with a redirect URL as draft.

Note that this redirect is intended to be used for non-asset URLs. More specifically in the case of an unpublished document, the user should be redirected to the original public URL of the relevant attachment's parent document. This will be set by the Whitehall app via the Asset Manager API.

### DIFF
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -38,7 +38,7 @@ private
   end
 
   def asset_params
-    params.require(:asset).permit(:file, :draft)
+    params.require(:asset).permit(:file, :draft, :redirect_url)
   end
 
   def find_asset(include_deleted: false)
--- a/app/controllers/whitehall_assets_controller.rb
+++ b/app/controllers/whitehall_assets_controller.rb
@@ -12,7 +12,10 @@ private
   def asset_params
     params
       .require(:asset)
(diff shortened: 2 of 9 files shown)


floehopper:
See that @chrisroos has added the access-limiting functionality for *all* assets in https://github.com/alphagov/asset-manager/pull/471, I think I should probably do the same for this redirect functionality, but I'm going to do it in a separate PR. Merging.

Issue #482, opened 2018-02-19
Allow Asset Manager API to return size


chrislo:
The `AssetManagerAndQuarentinedFileStorage` engine implements a `size` method that used the quarentined file to calculate the size. 

We plan to switch the storage engine for AttachmentUploader to `AssetManagerStorage` and therefore store all files exclusively in Asset Manger. This means we will need to implment the `size` method on `AssetManagerStorage` and it will have to fetch the data from the Asset Manager API. on remove all of the assets from the disk, so we need to instead calculate it by making a call to the Asset Manager API. We have [migrated all the existing attachments to Asset Manager](https://github.com/alphagov/asset-manager/issues/441) and are adding new ones to both the disk and AM. That means we can make this change now - it will make it easier to subsequently [store the attachments exlusively in Asset Manager](https://github.com/alphagov/asset-manager/issues/470). 

- [x] Store file size information in the Asset Manager Database when a new asset is added
- [x] Add file size information to the JSON returned by Asset Manager
- [x] Run `db:set_size_for_all_uploaded_assets` on `integration`
- [x] Run `db:set_size_for_all_uploaded_assets` on `staging`
- [x] Run `db:set_size_for_all_uploaded_assets` on `production`

chrislo:
The code in PR #483 has been deployed to production, so we are now saving asset sizes for newly uploaded assets.

chrislo:
The worker-based migration added in #484 is currently been run on integration and staging.

chrislo:
The migration took just under an hour to run (rake task + processing all jobs) on `staging`. It's considerably slower on integration (around 150 job/s compared to almost 3000 job/s on staging). 

<img width="1408" alt="screen shot 2018-02-20 at 10 22 21" src="https://user-images.githubusercontent.com/16707/36432353-00e49f44-1628-11e8-907e-12024bfcfddb.png">

I've asked 2ndline to run the task on `production`.

chrislo:
The migration completed in around 80 mins on production. 

<img width="1409" alt="screen shot 2018-02-20 at 11 34 52" src="https://user-images.githubusercontent.com/16707/36436417-283ae17a-1632-11e8-8b80-130d3edf9b68.png">

chrislo:
And, it completed in around 4.5 hours on integration. 

[...]

PR #483, opened 2018-02-19
Add asset file size to API

chrislo:
See: https://github.com/alphagov/asset-manager/issues/482

This PR adds an integer `size` attribtute to the JSON returned from a call to `GET http://asset-manager.<host>/assets/<id>`.  The JSON returned now looks like

```
{
  "_response_info": {
    "status": "ok"
  },
  "id": "http://asset-manager.dev.gov.uk/assets/5a689582759b744e3518efd8",
  "name": "minister-of-funk.960x640.jpg",
  "content_type": "image/jpeg",
  "size": 108106,
  "file_url": "http://assets-origin.dev.gov.uk/government/uploads/system/uploads/image_data/file/41/minister-of-funk.960x640.jpg",
  "state": "uploaded",
  "draft": false
}
```

The first 3 commits add the methods required to save the size to the database from the result of `stat`ing the file on disk before it is uploaded. These methods are very similar to the ones for setting `last_modified` and `md5_hexdigest` for example. 

2e5b016 adds the size to the JSON response. 

In 13d35c4 there is a migration rake task (which reuses `AssetProcessor` for progress indication) to set the `size` on all existing uploaded assets based on the size encoded in the `etag`. I've checked on the integration clone of the production data and as far as I can see the `etag` is set for all existing assets. I think this task should run in a reasonable time, but I intend to run it on integration to verify. Once migrated, this commit can be reverted.

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -14,4 +14,13 @@ namespace :db do
       puts 'Error: Unable to remove Asset#organisation_slug'
     end
   end
+
+  desc "Set the size field for all assets from the etag"
+  task set_size_for_all_uploaded_assets: :environment do
+    scope = Asset.unscoped.where(size: nil)
+    processor = AssetProcessor.new(scope: scope)
+    processor.process_all_assets_with do |asset_id|
+      scope.find(asset_id).set_size_from_etag
+    end
+  end
 end
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -122,6 +122,7 @@ protected
     self.etag = etag_from_file
     self.last_modified = last_modified_from_file
     self.md5_hexdigest = md5_hexdigest_from_file
+    self.size = size_from_file
   end
 
   def valid_filenames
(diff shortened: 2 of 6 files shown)


lib/tasks/db.rake
```diff
@@ -14,4 +14,13 @@ namespace :db do
       puts 'Error: Unable to remove Asset#organisation_slug'
     end
   end
+
+  desc "Set the size field for all uploaded assets from the etag"
+  task set_size_for_all_uploaded_assets: :environment do
+    scope = Asset.unscoped.where(state: 'uploaded').where(size: nil)
```
chrisroos: Having just approved this PR, I realise I have a question. Why do we only consider `uploaded` assets here? Should we update the size of _all_ assets?

chrislo: @chrisroos I'd added that because any assets that are in the process of being uploaded when this task runs will have their size set by the changes added earlier in the PR. But it's probably an unnecessary optimisation - I'll remove it.

chrislo:
Thanks @chrisroos - rebasing and force-pushing before merging.

PR #484, opened 2018-02-20
Use worker for asset size migration

chrislo:
In 16e323e we added a rake task to set the `size` field for all
existing assets. When we ran the task in integration however, we
estimated that it would take around 12 hours to complete. This runs
the risk of the rake task being killed and leaving the database in a
somewhat inconsistent state.

This PR moves the work of the rake task into a worker. Hopefully
this will improve the run time as the workers run with concurrency,
and the rake task that queues up the jobs will complete more quickly.

### DIFF
--- a/app/workers/set_asset_size_worker.rb
+++ b/app/workers/set_asset_size_worker.rb
@@ -1,0 +1,8 @@
+class SetAssetSizeWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: 'low_priority'
+
+  def perform(asset_id)
+    Asset.unscoped.find(asset_id).set_size_from_etag
+  end
+end
--- a/app/workers/set_asset_size_worker.rb
+++ b/app/workers/set_asset_size_worker.rb
@@ -3,6 +3,8 @@ class SetAssetSizeWorker
   sidekiq_options queue: 'low_priority'
 
   def perform(asset_id)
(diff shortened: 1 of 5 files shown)


app/workers/set_asset_size_worker.rb
```diff
@@ -0,0 +1,8 @@
+class SetAssetSizeWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: 'low_priority'
+
+  def perform(asset_id)
+    Asset.unscoped.find(asset_id).set_size_from_etag
```
floehopper: I'd be inclined to inline the `Asset#set_size_from_etag` method into the worker, because I don't think it's used by any production code and so it should minimise the chance that the code gets left behind when we remove this one-off migration.

floehopper: Are you confident that the call to `Asset#save` doesn't have any side-effects? As I mentioned in Slack earlier, I'd be inclined to use `Mongoid::Persistable::Settable#set` here c.f. https://github.com/alphagov/asset-manager/blob/314ddfcdee40e2e25273c0acb3327f3fd9771a99/app/workers/asset_file_metadata_worker.rb#L6-L10.

chrislo:
Thanks @floehopper - I've addressed your two suggestions in two new commits, thank you!

Issue #419, opened 2018-01-19
Ensure we're setting the correct Content-Type for Whitehall attachments


chrisroos:
Similar to "Ensure we support content types of whitehall assets #418". We'll need to ensure we're setting the correct `Content-Type` (or at least matching the `Content-Type` set by Whitehall) once we're serving Whitehall's attachments from Asset Manager.

chrislo:
I had a list of almost all of the attachments that we'd generated as part of <https://github.com/alphagov/asset-manager/issues/441>.

I stripped out everything after the final `.` in the file names to get a list of unique extensions.

    EXTENSTIONS=`cat asset_manager_assets_sorted.txt | awk -F'.' '{print tolower($NF)}' | sort | uniq`

    csv
    doc
    docx
    dot
    eps
    gif
    ics
    jpg
    odp
    ods
    odt
    pdf
    png
    ppt
    pptx
    ps
    rdf
    rtf
    txt
    xls
    xlsm
    xlsx
    xlt
    xml
    xsd
    zip

I then selected ten random examples of each file from the list

    for i in $EXTENSIONS; do
      grep "\.$i$" asset_manager_assets_sorted.txt | sort -R | head -n 10 > paths.txt
    done

Using this ruby script I called \`curl\` to follow redirects to their
final destination, extract the filename of that final destination in
case it was different to the original and report the content type.

    paths = File.read('paths.txt')
    
    paths.each_line do |path|
      path = path.strip
    
      whitehall_request = "curl -sLI https://www.gov.uk/government/uploads/system/uploads/#{path}"
      whitehall_response = `#{whitehall_request}`.split("\n")
    
      status = whitehall_response.grep(/^HTTP/).last
      content_type = whitehall_response.grep(/^Content-Type/).last
      content_disposition = whitehall_response.grep(/^Content-Disposition/).last
    
      if content_disposition
        final_filename = content_disposition.scan(/filename="(.+)"/).flatten.first
        extension = File.extname(final_filename)
        puts [extension, content_type.strip, final_filename, path, status].join("\t")
      end
    end

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">


<colgroup>
<col  class="org-left" />

<col  class="org-left" />

<col  class="org-left" />

<col  class="org-left" />

<col  class="org-left" />
</colgroup>
<thead>
<tr>
<th scope="col" class="org-left">Extension</th>
<th scope="col" class="org-left">Content-Type</th>
[...]

PR #496, opened 2018-02-27
Add Whitehall Attachment Content-Type headers

chrislo:
See: https://github.com/alphagov/asset-manager/issues/419

We are planning to change nginx so that requests for Whitehall attachments are served from Asset Manager instead of Whitehall. We've looked at all the existing attachments under `attachment_data` and requested examples of them to see what Content-Type header they are served with (#419). This PR registers the missing types so that the assets are served with the correct header.

### DIFF
--- a/config/initializers/mime_types.rb
+++ b/config/initializers/mime_types.rb
@@ -3,13 +3,10 @@
 # Add new mime types for use in respond_to blocks:
 # Mime::Type.register "text/richtext", :rtf
 
-Mime::Type.register "text/rtf", :rtf
-
-Mime::Type.register "application/vnd.ms-excel", :xls
 Mime::Type.register "application/msword", :doc
-
-Mime::Type.register "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet", :xlsx
-Mime::Type.register "application/vnd.openxmlformats-officedocument.wordprocessingml.document", :docx
-
+Mime::Type.register "application/vnd.ms-excel", :xls
 Mime::Type.register "application/vnd.oasis.opendocument.spreadsheet", :ods
 Mime::Type.register "application/vnd.oasis.opendocument.text", :odt
+Mime::Type.register "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet", :xlsx
+Mime::Type.register "application/vnd.openxmlformats-officedocument.wordprocessingml.document", :docx
+Mime::Type.register "text/rtf", :rtf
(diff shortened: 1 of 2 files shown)


config/initializers/mime_types.rb
```diff
@@ -4,9 +4,16 @@
...
 Mime::Type.register "application/vnd.ms-excel", :xls
+Mime::Type.register "application/vnd.ms-excel", :xlt
+Mime::Type.register "application/vnd.ms-excel.sheet.macroEnabled.12", :xlsm
+Mime::Type.register "application/vnd.ms-powerpoint", :ppt
 Mime::Type.register "application/vnd.oasis.opendocument.spreadsheet", :ods
 Mime::Type.register "application/vnd.oasis.opendocument.text", :odt
+Mime::Type.register "application/vnd.openxmlformats-officedocument.present", :pptx
 Mime::Type.register "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet", :xlsx
 Mime::Type.register "application/vnd.openxmlformats-officedocument.wordprocessingml.document", :docx
+Mime::Type.register "text/plain; charset=utf-8", :txt
```
floehopper: Will this mean there will be a change in behaviour for Mainstream assets with a `txt` file extension?

Is this intentional? If so, I think it would be useful to highlight this in the commit note and explain what the content type used to be for this file extension and why we're happy to change it.

floehopper: Will this mean there will be a change in behaviour for Mainstream assets with a `txt` file extension?

Is this intentional? If so, I think it would be useful to highlight this in the commit note and explain what the content type used to be for this file extension and why we're happy to change it.

floehopper: Are there any other file extensions which will now have a different content type for Mainstream assets? If so, it would be good to make this clear in the commit note and explain why we're happy with the changes.

chrislo: The change to `.txt` affects 3 non-whitehall assets. I've explained more in the commit note. 

The other file extensions affected are ones where Asset Manager would previously have returned `application/octet-stream`, which were incorrect. More detail of this is in the [linked comment](https://github.com/alphagov/asset-manager/issues/419#issuecomment-368575610). I don't think I'll count all of these up and list out which ones are affected now, as this change is an improvement (it sets the correct content-type).

chrislo:
Thanks @floehopper! I've addressed your feedback in commit notes, and added a new commit to add the 2 additional, unused mime-type registrations in whitehall.

PR #497, opened 2018-02-27
Improvements to cache control

floehopper:
This fixes a minor bug in the existing code. The bug hadn't led to any incorrect external behaviour, but it meant that the code was confusing and could've led to unintended incorrect external behaviour in the future.

I've also reduced the amount of duplication slightly which will make it easier to apply some upcoming changes.


floehopper:
Rebasing against `master` and force-pushing in preparation for merging.

Issue #498, opened 2018-02-27
Avoid multiple redirects if a replacement asset is itself replaced


floehopper:
Currently Whitehall implements this [at the end of `AttachmentData#replace_with!`][1]. In the short term I'm not planning to move this behaviour into Asset Manager, but I think it probably does make sense to move it in the longer term. This is why I'm capturing it in an issue outside the ["Serve Whitehall attachment assets from Asset Manager" milestone][2].

[1]: https://github.com/alphagov/whitehall/blob/446e60b80fee198540e5abe3a7077a5a7f5f63e5/app/models/attachment_data.rb#L101-L103
[2]: https://github.com/alphagov/asset-manager/milestone/7

Issue #494, opened 2018-02-26
Set appropriate Cache-Control header for Whitehall assets when user is signed in


floehopper:
It looks as if Whitehall attachments are currently being served [with a `Cache-Control` header of `no-cache`][1] if the user is signed in. I assume this is meant to be used to ensure that editors see the most up-to-date version of an attachment.

I have a feeling this behaviour might've been present for non-attachments in Whitehall in the old `PublicUploadsController`. It would be worth checking in order to inform the decision of how/whether to implement equivalent behaviour in Asset Manager.

It's not immediately obvious to me whether we should do something similar in Asset Manager, because I'm not entirely sure the fact that the `current_user` is set in Asset Manager is semantically equivalent to `current_user` in Whitehall.

[1]: https://github.com/alphagov/whitehall/blob/master/app/controllers/attachments_controller.rb#L79

chrisroos:
@chrislo and I have opened https://github.com/alphagov/asset-manager/pull/517 to update the caching behaviour in Asset Manager.

chrisroos:
I'm going to test that this is working as expected.

chrisroos:
I've tested this in production by requesting the same image from `assets-origin` and from `draft-assets`:

* https://assets-origin.publishing.service.gov.uk/media/55bb8be1ed915d1565000016/large-goods-vehicle-markings.jpg
* https://draft-assets.publishing.service.gov.uk/media/55bb8be1ed915d1565000016/large-goods-vehicle-markings.jpg

I used the Chrome Developer tools to inspect the Response headers and have confirmed that the `Cache-Control` header is set to `max-age=86400, public` when requested from `assets-origin` and is set to `no-cache` when requested from `draft-assets`.

I'm happy that this is working as expected so I'm closing this issue.

PR #499, opened 2018-02-27
Allow asset to be replaced with another asset

floehopper:
When a Whitehall attachment is "replaced", requests for its URL result in a redirect to the attachment which replaced it. This PR implements this functionality in Asset Manager so that we'll be able to serve Whitehall attachments from Asset Manager without changing the externally visible behaviour.

This PR introduces the `Asset#replacement` association. If an asset has such a replacement then requests for it are redirected to the path for the replacement. This is equivalent to [the behaviour implemented in Whitehall's `AttachmentsController#fail` method][1].

This PR also allows a `replacement_id` to be specified when creating or updating an asset via the API. I plan to call this [when an attachment is replaced in Whitehall][2].

Notes:

* I have not addressed the issue of setting a different value for the `Cache-Control` response header when the user is not signed in, because I think this is a wider concern which has been captured in [this GitHub issue][3].

* I have not implemented the optimisation which avoids multiple redirects where a replacement asset is itself replaced. For the moment, we can retain this behaviour by leaving it in Whitehall, but it would probably make sense to move it into Asset Manager at some point so we don't need to implement it in other publisher apps. See [this GitHub issue][4] for more details.

* This work is based on the work in #497. The first commit is a squashed version of that PR and should be removed/rebased-out before merging this PR.

[1]: https://github.com/alphagov/whitehall/blob/446e60b80fee198540e5abe3a7077a5a7f5f63e5/app/controllers/attachments_controller.rb#L45-L47
[2]: https://github.com/alphagov/whitehall/blob/446e60b80fee198540e5abe3a7077a5a7f5f63e5/app/models/attachment_data.rb#L100
[3]: https://github.com/alphagov/asset-manager/issues/494
[4]: https://github.com/alphagov/asset-manager/issues/498

### DIFF
--- a/spec/controllers/assets_controller_spec.rb
+++ b/spec/controllers/assets_controller_spec.rb
@@ -135,6 +135,24 @@ RSpec.describe AssetsController, type: :controller do
         end
       end
 
+      context 'and replacement_id does not match an existing asset' do
+        let(:replacement_id) { 'non-existent-asset-id' }
+
+        it 'responds with unprocessable entity status' do
+          post :create, params: { asset: attributes }
+
+          expect(response).to have_http_status(:unprocessable_entity)
+        end
+
+        it 'includes error message in response' do
+          post :create, params: { asset: attributes }
+
+          body = JSON.parse(response.body)
+          status = body['_response_info']['status']
+          expect(status).to include('Replacement not found')
+        end
+      end
+
       it 'includes the replacement_id in the response' do
         post :create, params: { asset: attributes }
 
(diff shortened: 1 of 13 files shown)


spec/controllers/assets_controller_spec.rb
```diff
@@ -135,6 +135,24 @@
         end
       end
 
+      context 'and replacement_id is does not match an existing asset' do
```
chrislo: typo

floehopper: Thanks! 👍

floehopper:
I've fixed the typo spotted by @chrislo in the last commit and I'm rebasing against `master` and force-pushing in preparation for merging.

PR #502, opened 2018-03-01
Add support for parent document URL to Asset Manager

chrisroos:
Whitehall attachments can include a `Link` header containing the URL of their parent document (added to Whitehall in alphagov/whitehall#2202). We need to add support for this functionality to Asset Manager so that we can serve Whitehall attachments from this app.

Although this is primarily for Whitehall attachments I felt that it'd be less confusing to add support for it to both non-Whitehall and Whitehall assets.

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -45,6 +45,7 @@ class Asset
 
   validate :check_specified_replacement_exists
   validate :prevent_transition_from_published_to_draft_if_replaced
+  validate :ensure_parent_document_url_is_valid
 
   mount_uploader :file, AssetUploader
 
@@ -174,6 +175,20 @@ protected
       if redirect_url.present?
         errors.add(:draft, 'cannot be true, because already redirected')
       end
+    end
+  end
+
+  def ensure_parent_document_url_is_valid
+    return unless parent_document_url.present?
+
+    begin
+      uri = Addressable::URI.parse(parent_document_url)
+    rescue Addressable::URI::InvalidURIError
+      uri = nil
+    end
+
+    unless uri && %w(http https).include?(uri.scheme)
+      errors.add(:parent_document_url, 'must be an http(s) URL')
     end
   end
 end
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -32,6 +32,8 @@ class Asset
 
   field :access_limited, type: Array, default: []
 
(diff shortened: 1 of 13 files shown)


app/models/asset.rb
```diff
@@ -32,6 +32,8 @@ class Asset
 
   field :access_limited, type: Array, default: []
 
+  field :parent_document_url, type: String
```
chrislo: Do you think it is worth adding any validation to ensure that the value stored is a valid URL?

chrisroos: Yeah, I reckon it is. I did consider initially but discounted it for reasons I no longer remember. I'll add it to this PR.

floehopper: I also considered doing this for `Asset#redirect_url`, but discounted it, because I thought it would be complicated to know what hosts were allowed in the various environments + draft/public stacks. Are you imagining just checking it's a valid URL or that it's a valid GOV.UK URL?

chrislo: Probably just that it's a valid url - so that any header added is valid against https://tools.ietf.org/html/rfc5988#section-5.

floehopper: Do you think we should add some validation for `Asset#redirect_url` and, if so, what?

chrisroos: I remember now - it's because it's not particularly easy to add such validation :-)

@chrislo - What do you think of the validation I've added in https://github.com/alphagov/asset-manager/pull/502/commits/f51bb68965e01280aec9ec72b14068341000e57a?

chrisroos: Sorry, @floehopper. I've only just seen your comments (even though they were posted before mine). What do you think about the validation I've added? Does it seem reasonable?

floehopper: It seems OK to me. Did you consider using a validator class? There might be some inspiration in https://github.com/perfectline/validates_url or it might even be worth using it...?

chrisroos: I thought a validator class was overkill for validating a single attribute, but can see that it might be worth introducing if we do need to validate multiple URLs.

I've had a quick look at the validates_url Gem. It seems reasonable but I don't feel particularly strongly about whether we should use it.

chrisroos:
I've rebased this on master and force pushed in preparation for merging.

Issue #450, opened 2018-01-30
Implement access-limiting behaviour for Whitehall attachments in Asset Manager


floehopper:
e.g. limited to groups of users in an organisation

chrisroos:
I'm going to start working on this.

chrisroos:
I'm working on the changes necessary to Asset Manager in "WIP: Allow draft asset access to be restricted to specific users #471".

I'm working on the changes necessary to Whitehall in "WIP: Send access_limited data to Asset Manager #3767".

chrisroos:
The changes in https://github.com/alphagov/asset-manager/pull/471 and https://github.com/alphagov/whitehall/pull/3767 have been deployed to production.

@floehopper realised that we also need to handle attachments created by the `BulkUploadsController` in Whitehall. I had hoped that the `BulkUploadsController` might not be used any longer but there are (a small number of) requests for it within the last 14 days so I think we do need to handle these attachments.

chrisroos:
I made the changes to the `BulkUploadsController` in Whitehall in https://github.com/alphagov/whitehall/pull/3822 which has also been merged to master.

I plan to test the functionality in integration to ensure that it's working as expected.

floehopper:
@chrisroos Does this work for `OfficialStatistics` & `NationalStatistics` publication types which are access-limited by default?

* [`PublicationType::OfficialStatistics`][1]
* [`PublicationType::NationalStatistics`][2]

[1]: https://github.com/alphagov/whitehall/blob/32bb13cf3ae3ae9ca71fa14bd4c620c631660419/app/models/publication_type.rb#L100
[2]: https://github.com/alphagov/whitehall/blob/32bb13cf3ae3ae9ca71fa14bd4c620c631660419/app/models/publication_type.rb#L101

chrisroos:
Good question, @floehopper. I'll investigate.

chrisroos:
> @chrisroos Does this work for `OfficialStatistics` & `NationalStatistics` publication types which are access-limited by default?

I've checked and believe that this will just work because setting `Publication#publication_type` to something that's access limited by default will set the `access_limited` attribute of that publication to `true`:

```
> Publication.new.access_limited?
=> nil
> Publication.new(publication_type: PublicationType::OfficialStatistics).access_limited?
=> true
> Publication.new(publication_type: PublicationType::PolicyPaper).access_limited?
=> false
```

chrisroos:
[...]

PR #510, opened 2018-03-06
Allow access_limited to be reset on update

chrisroos:
While investigating issue #450, I noticed that resetting/removing the access_limited status of an asset wasn't working as expected.

This is because when we pass an empty `access_limited` array to gds-api-adapters/rest-client it's turned into an empty string in Asset Manager. This results in an `UnpermittedParameter` exception for the `access_limited` parameter in Asset Manager as it's expecting an Array but receives a String. The fix is to detect the empty string in the controller and convert it back into the empty array that's intended. I don't think this is particularly nice but I think it's good enough for now.

Note that I haven't had to use `handle_empty_access_limited_param` in `WhitehallAssetsController#asset_params` as the problem only occurs when _updating_ an access-limited asset, and the `WhitehallAssetsController` is only responsible for _creating_ assets.

I also considered using two fields to implement the access-limiting functionality: a boolean to control whether access-limiting was enabled in conjunction with an array of user UIDs who are authorised to view the asset. I decided against that approach because it's more work and because it means the implementation in Asset Manager diverges from the access-limiting functionality in the Content Store.

### DIFF
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -39,7 +39,7 @@ private
 
   def asset_params
     normalize_redirect_url(
-      handle_empty_access_limited_param(params)
+      normalize_access_limited(params)
         .require(:asset)
         .permit(:file, :draft, :redirect_url, :replacement_id, :parent_document_url, access_limited: [])
     )
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -38,8 +38,9 @@ private
   end
 
   def asset_params
+    asset_params = handle_empty_access_limited_param(params)
     exclude_blank_redirect_url(
-      params
+      asset_params
         .require(:asset)
         .permit(:file, :draft, :redirect_url, :replacement_id, :parent_document_url, access_limited: [])
     )
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -38,9 +38,8 @@ private
   end
 
   def asset_params
-    asset_params = handle_empty_access_limited_param(params)
     exclude_blank_redirect_url(
-      asset_params
+      handle_empty_access_limited_param(params)
         .require(:asset)
         .permit(:file, :draft, :redirect_url, :replacement_id, :parent_document_url, access_limited: [])
     )
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -38,7 +38,7 @@ private
   end
 
   def asset_params
-    exclude_blank_redirect_url(
+    normalize_redirect_url(
       handle_empty_access_limited_param(params)
         .require(:asset)
         .permit(:file, :draft, :redirect_url, :replacement_id, :parent_document_url, access_limited: [])
--- a/app/controllers/base_assets_controller.rb
+++ b/app/controllers/base_assets_controller.rb
@@ -22,7 +22,7 @@ protected
     params.reject { |k, v| (k.to_sym == :redirect_url) && v.blank? }
   end
 
-  def handle_empty_access_limited_param(params)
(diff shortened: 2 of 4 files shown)


app/controllers/assets_controller.rb
```diff
@@ -38,8 +38,9 @@ def restrict_request_format
   end
 
   def asset_params
+    asset_params = handle_empty_access_limited_param(params)
```
floehopper: I think it would be useful to explain why you have to operate on the `params` *before* the calls to `require` and `permit`.

app/controllers/assets_controller.rb
```diff
@@ -38,8 +38,9 @@ def restrict_request_format
...
     exclude_blank_redirect_url(
-      params
+      asset_params
         .require(:asset)
         .permit(:file, :draft, :redirect_url, :replacement_id, :parent_document_url, access_limited: [])
     )
```
floehopper: A possible alternative approach which I think looks a bit cleaner:

```ruby
  def asset_params
      exclude_blank_redirect_url(
        handle_empty_access_limited_param(params)
          .require(:asset)
          .permit(:file, :draft, :redirect_url, :replacement_id, :parent_document_url, access_limited: [])
      )
    )
  end
```

app/controllers/base_assets_controller.rb
```diff
@@ -22,6 +22,13 @@ def exclude_blank_redirect_url(params)
     params.reject { |k, v| (k.to_sym == :redirect_url) && v.blank? }
   end
 
+  def handle_empty_access_limited_param(params)
```
floehopper: Minor: I think the `_param` suffix is unnecessary, given `params` are passed in as an argument, and inconsistent with `exclude_blank_redirect_url`. I'd prefer the method name to be `handle_empty_access_limited`.

floehopper: Minor: Possibly rename both these methods as follows:

* `exclude_blank_redirect_url ` -> `normalize_redirect_url`
* `handle_empty_access_limited_param` -> `normalize_access_limited`

app/controllers/base_assets_controller.rb
```diff
@@ -22,6 +22,13 @@ def exclude_blank_redirect_url(params)
...
+    if params.has_key?(:asset) && params[:asset].has_key?(:access_limited) && params[:asset][:access_limited].empty?
+      params[:asset][:access_limited] = []
+    end
+    params
+  end
+
```
floehopper: This seems quite "ugly", but I can't see a way to simplify it. I think it would be useful to mention other solutions you considered, e.g. having two fields: `Asset#access_limit?` & `Asset#authorized_users`, and ideally capturing these ideas in a GitHub issue.

app/controllers/assets_controller.rb
chrisroos: I've updated the commit note to explain why I need to modify the params before calling `require` and `permit`.

app/controllers/assets_controller.rb
chrisroos: I've changed this in a new commit in this branch.

app/controllers/base_assets_controller.rb
chrisroos: I've renamed the two methods as you've suggested - in two new commits in this branch.

app/controllers/base_assets_controller.rb
chrisroos: I've updated the commit note in the initial commit to explain the alternative approach of having two fields.

I've captured the idea of using two fields in https://github.com/alphagov/asset-manager/issues/511.

chrisroos:
Thanks for reviewing, @floehopper. I've addressed your feedback, rebased on master, force pushed and merged!

PR #512, opened 2018-03-08
Use external signon URL for OAuth exchange

chrisroos:
I noticed that visiting the draft-assets host was resulting in a redirect to the internal signon URL (e.g. https://signon.integration.govuk-internal.digital). It should be redirecting to the external URL (e.g. https://signon.integration.publishing.service.gov.uk) as we need to be able to follow it in a browser.

Note that We're not yet sending any traffic to draft-assets so this incorrect behaviour wasn't actually causing us a problem.

I've [added a test to Smokey][1] to help avoid this problem in future.

[1]: https://github.com/alphagov/smokey/pull/341

### DIFF
--- a/config/initializers/gds-sso.rb
+++ b/config/initializers/gds-sso.rb
@@ -2,6 +2,6 @@ GDS::SSO.config do |config|
   config.user_model   = "User"
   config.oauth_id     = ENV["OAUTH_ID"] || "abcdefghjasndjkasndassetmanager"
   config.oauth_secret = ENV["OAUTH_SECRET"] || "secret"
-  config.oauth_root_url = Plek.current.external_url_for("signon")
+  config.oauth_root_url = Plek.new.external_url_for("signon")
   config.cache = Rails.cache
 end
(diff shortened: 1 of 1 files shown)


config/initializers/gds-sso.rb
```diff
@@ -2,6 +2,6 @@
   config.user_model   = "User"
   config.oauth_id     = ENV["OAUTH_ID"] || "abcdefghjasndjkasndassetmanager"
   config.oauth_secret = ENV["OAUTH_SECRET"] || "secret"
-  config.oauth_root_url = Plek.current.find("signon")
+  config.oauth_root_url = Plek.current.external_url_for("signon")
```
floehopper: Minor: I think `Plek.current` is "sort of" deprecated - see [this comment](https://github.com/alphagov/plek/blob/3b838fd022a5890946a3a3ce7e774d758e0c6320/lib/plek.rb#L148-L152). I think approved way is `Plek.new.external_url_for("signon")`. It might be nice to make this change while we're at it.

chrisroos: Good spot. I've updated this to use `Plek.new` in a new commit in this branch.

chrisroos:
This has been deployed to integration and I've tested that it's working as expected: requesting draft-assets now redirects me to signon.integration.publishing.service.gov.uk as expected.

PR #517, opened 2018-03-12
Avoid caching responses made to draft-assets host

chrisroos:
Draft assets can be optionally access-limited to a subset of users.
Caching the response to a request for an access-limited draft asset
might mean that unauthorised users can see things they shouldn't be able
to.

This is similar to the behaviour in Whitehall where [attachment requests
made by signed-in users aren't cached][1].

[1]: https://github.com/alphagov/whitehall/blob/cc57fbae96042952584d90583952d077be1c59a8/app/controllers/base_attachments_controller.rb#L35

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -33,7 +33,11 @@ class MediaController < BaseMediaController
 
     respond_to do |format|
       format.any do
-        set_expiry(cache_control)
+        if requested_from_draft_assets_host?
+          expires_now
+        else
+          set_expiry(cache_control)
+        end
         add_link_header(asset)
         headers['X-Frame-Options'] = AssetManager.frame_options
         proxy_to_s3_via_nginx(asset)
--- a/app/controllers/whitehall_media_controller.rb
+++ b/app/controllers/whitehall_media_controller.rb
@@ -36,7 +36,11 @@ class WhitehallMediaController < BaseMediaController
       return
     end
 
-    set_expiry(cache_control)
+    if requested_from_draft_assets_host?
+      expires_now
+    else
+      set_expiry(cache_control)
+    end
     add_link_header(asset)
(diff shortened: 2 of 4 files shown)


app/controllers/media_controller.rb
```diff
@@ -33,7 +33,11 @@ def download
 
     respond_to do |format|
       format.any do
-        set_expiry(cache_control)
+        if requested_from_draft_assets_host?
+          expires_now
```
floehopper: Given that we're already using `set_expiry(cache_control.expires_in(0.minutes))` [in `BaseAssetsController#show`][1], I think it would be more consistent (and avoid breaking encapsulation) to use something like `set_expiry(cache_control.expires_now)`. In fact it might also make sense to change `BaseAssetsController#show` to use that too!

Note that I can imagine this would all become simpler (and we might no longer need `CacheControlConfiguration`) if/when we merge more of the behaviour in the two media controllers.

[1]: https://github.com/alphagov/asset-manager/blob/2f509d96d852c552e184d7e2b6135af6848a0cbe/app/controllers/base_assets_controller.rb#L5

app/controllers/whitehall_media_controller.rb
```diff
@@ -36,7 +36,11 @@ def download
       return
     end
 
-    set_expiry(cache_control)
+    if requested_from_draft_assets_host?
+      expires_now
```
floehopper: Same comment as above.

app/controllers/media_controller.rb
chrisroos: We've just discussed this. It's not entirely trivial to add something to `CacheControlConfiguration` as it doesn't look as though `expires_in` (used in `ApplicationController#set_expiry`) allows us to generate the `no_cache` header that we're interested in. We're going to use `expires_now` for now and remove the `CacheControlConfiguration` class by standardising on a public cache of 4 hours.

Issue #518, opened 2018-03-12
Standardise Whitehall and Mainstream Cache-Control headers 


chrisroos:
Whitehall assets are cached for 4 hours while Mainstream assets are cached for 24 hours. Making these the same will allow us to simplify Asset Manager.

chrisroos:
We've used Kibana to count the number of 304 responses for Mainstream assets from Asset Manager over the last 7 days (`http_host:assets.publishing.service.gov.uk AND status:304 AND request:/GET \/media\/.*/ #=> 219,087`). 

Mainstream assets are cached for 24 hours. By changing this to 4 hours, we'd expect to see around 6 times the number of requests resulting in 304 responses from Asset Manager (i.e. 1,200,000 instead of 200,000). In the same 7 day period Asset Manager responded to 2,208,911 GET requests (`application:asset-manager AND request:/GET.*/`). The additional 1,000,000 requests represents nearly a 50% increase in total requests so it's not obvious this is the way to go.

chrisroos:
We've used Kibana to count the number of 304 responses for Whitehall assets from Asset Manager over the last 7 days (`http_host:assets.publishing.service.gov.uk AND status:304 AND request:/GET \/government\/.*/ #=> 1,047,031).

Whitehall assets are cached for 4 hours. By changing this to 24 hours, we'd expect to see around 6 times fewer requests resulting in 304 responses from Asset Manager (i.e. 174,505 instead of 1,047,031).

chrisroos:
We're currently preparing to start serving about 1.5m Whitehall attachments from Asset Manager so we're currently thinking that it might be better to increase the cache time across the board to 24 hours.

chrisroos:
@chrislo and I have started working on standardising the caching in https://github.com/alphagov/asset-manager/pull/523.

chrisroos:
The changes in https://github.com/alphagov/asset-manager/pull/523 were merged yesterday. I've confirmed that they're working as expected in integration by requesting the same asset from integration and production and comparing the `Cache-Control` header. You can see that it was 4 hours in production (the previous value) and it's now 24 hours in integration (the new value).

```
# Production
$ curl -v "https://assets.publishing.service.gov.uk/government/uploads/system/uploads/person/image/6/s216_PM_portrait_960x640.jpg" > /dev/null

< Cache-Control: max-age=14400, public

# Integration
[...]

PR #523, opened 2018-03-14
Standardise cache time for all assets

chrisroos:
This addresses issue https://github.com/alphagov/asset-manager/issues/518 by standardising on a 24 hour cache for both mainstream and Whitehall assets.

I've removed the `CacheControlConfiguration` class as it feels like overkill now that we're not using the flexibility it provides.


chrisroos:
Can you review this when you have a minute please, @chrislo.

chrisroos:
Thanks for reviewing, @chrislo.

I've rebased on master in preparation for merging.

Issue #525, opened 2018-03-15
Update metadata of Whitehall attachments in Asset Manager


chrisroos:
We've added a Rake task to bulk update the metadata of Whitehall attachments in Asset Manager in https://github.com/alphagov/whitehall/pull/3864.

chrisroos:
@floehopper and I have been through the 6 `AssetManagerAttachment*` workers in Whitehall and have confirmed that they appear to be safe to run during this migration (i.e. outside of the context in which they're normally run).

chrisroos:
We've discovered that there are a number of `AttachmentData` records (in integration at least) without a corresponding file on the file system. Requesting a URL for one of these `AttachmentData` records can result in the user being redirected elsewhere by the [`BaseAttachmentsController#fail` method](https://github.com/alphagov/whitehall/blob/71dee913bae15186e43be930eb1a0eae855e33c2/app/controllers/base_attachments_controller.rb#L8). When we migrated all the Whitehall attachments to Asset Manager we did it based on files on disk (instead of records in the database) which means that we might need to create assets in Asset Manager representing some of these `AttachmentData` records.

chrislo:
I ran the rake task `asset_manager:update_attachment_metadata[0,10000]` [on integration](https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/276/console) at 16:03:09 on 2018-03-15. It completed at 16:04:12 (~1min). 

At peak (16:04:22) the rake task had added 47345 jobs to the `asset_migration` queue. The queue was empty at 16:29:02.

chrisroos:
I've written a [script to save all the `AttachmentData` URLs to a text file](https://gist.github.com/chrisroos/7f728a0668289b6473d939acfc74a92d). The idea being that we can request these paths from `assets-origin` (Whitehall) and `draft-assets` (Asset Manager) to ensure that the responses are identical. I've not saved the output of the script anywhere publicly because I'm not sure we necessarily want it to be easy for someone to crawl such a list.

chrisroos:
I've written a [script to print the IDs of all the `AttachmentData` records that don't have a corresponding file on the filesystem](https://gist.github.com/chrisroos/3c22337d23c015c88ae78efdf516d27a). We might need to create assets in Asset Manager for these `AttachmentData` records if requesting them results in a redirect in Whitehall.

2ndline are currently running this script in staging for me.

chrisroos:
[...]

PR #528, opened 2018-03-19
Add temporary Rake task for creating Whitehall Attachments

chrisroos:
In issue #525 we've identified a number of Whitehall attachments that don't have a corresponding file on the filesystem. A subset of these attachments have a `replaced_by` object which results in requests for the original URL being redirected to the `replaced_by` URL. This Rake task creates assets for these attachments so that the redirects continue to work once we start serving the attachments from Asset Manager.

The scripts we used to come up with the hardcoded list of redirects are linked to from issue #525.

### DIFF
--- a/lib/tasks/create_missing_whitehall_assets.rake
+++ b/lib/tasks/create_missing_whitehall_assets.rake
@@ -12,7 +12,6 @@ task create_missing_whitehall_assets: :environment do
     ["/government/uploads/system/uploads/attachment_data/file/9559/att0212.xls", "/government/uploads/system/uploads/attachment_data/file/230269/att0212.xls"],
     ["/government/uploads/system/uploads/attachment_data/file/11510/climate-change-2011-tables-xls.zip", "/government/uploads/system/uploads/attachment_data/file/230270/climate-change-2011-tables-xls.zip"],
     ["/government/uploads/system/uploads/attachment_data/file/11723/Water_Efficiency_Calculator_Rev_02.xls", "/government/uploads/system/uploads/attachment_data/file/205789/The_water_efficiency_calculator_tool.xls"],
-    ["/government/uploads/system/uploads/attachment_data/file/28419/Form_NSV001.pdf", "/government/uploads/system/uploads/attachment_data/file/303346/Form_NSV001.pdf"],
     ["/government/uploads/system/uploads/attachment_data/file/210359/positive-for-youth-consultation-responses.doc", "/government/uploads/system/uploads/attachment_data/file/210381/positive-for-youth-consultation-responses.doc"],
     ["/government/uploads/system/uploads/attachment_data/file/210360/positive-for-youth-young-peoples-role-in-its-development.doc", "/government/uploads/system/uploads/attachment_data/file/210382/positive-for-youth-young-peoples-role_in_its_development.doc"],
     ["/government/uploads/system/uploads/attachment_data/file/211761/110713_Local_CO2_NS_-_Annex_A__Statistical_release_.pdf", "/government/uploads/system/uploads/attachment_data/file/322819/20140624_Statistical_release_Local_Authority_CO2_emissions.pdf"],
--- a/lib/tasks/create_missing_whitehall_assets.rake
+++ b/lib/tasks/create_missing_whitehall_assets.rake
@@ -1,0 +1,53 @@
+# rubocop:disable Metrics/BlockLength
+task create_missing_whitehall_assets: :environment do
+  placeholder_path = Rails.root.join('tmp/whitehall-attachment-placeholder.txt')
+  File.open(placeholder_path, 'w') do |file|
+    file.puts('whitehall-attachment-placeholder.txt')
+  end
+
+  redirects = [
+    ["/government/uploads/system/uploads/attachment_data/file/2205/business-plan-12.pdf", "/government/uploads/system/uploads/attachment_data/file/282785/business-plan-12.pdf"],
+    ["/government/uploads/system/uploads/attachment_data/file/2206/business-plan-12-annexes.pdf", "/government/uploads/system/uploads/attachment_data/file/282786/business-plan-12-annexes.pdf"],
+    ["/government/uploads/system/uploads/attachment_data/file/2319/epc.pdf", "/government/uploads/system/uploads/attachment_data/file/49997/1790388.pdf"],
+    ["/government/uploads/system/uploads/attachment_data/file/9559/att0212.xls", "/government/uploads/system/uploads/attachment_data/file/230269/att0212.xls"],
+    ["/government/uploads/system/uploads/attachment_data/file/11510/climate-change-2011-tables-xls.zip", "/government/uploads/system/uploads/attachment_data/file/230270/climate-change-2011-tables-xls.zip"],
+    ["/government/uploads/system/uploads/attachment_data/file/11723/Water_Efficiency_Calculator_Rev_02.xls", "/government/uploads/system/uploads/attachment_data/file/205789/The_water_efficiency_calculator_tool.xls"],
+    ["/government/uploads/system/uploads/attachment_data/file/28419/Form_NSV001.pdf", "/government/uploads/system/uploads/attachment_data/file/303346/Form_NSV001.pdf"],
+    ["/government/uploads/system/uploads/attachment_data/file/210359/positive-for-youth-consultation-responses.doc", "/government/uploads/system/uploads/attachment_data/file/210381/positive-for-youth-consultation-responses.doc"],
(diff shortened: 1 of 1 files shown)


lib/tasks/create_missing_whitehall_assets.rake
```diff
@@ -0,0 +1,53 @@
...
+    ["/government/uploads/system/uploads/attachment_data/file/225280/SIN_Officer_Internal_Advert.docx", "/government/uploads/system/uploads/attachment_data/file/236779/Press_Reader_JD_and_Advert.pdf"],
+    ["/government/uploads/system/uploads/attachment_data/file/296338/mca-legislation-si.csv", "/government/uploads/system/uploads/attachment_data/file/297534/mca-legislation-si.csv"],
+    ["/government/uploads/system/uploads/attachment_data/file/490383/Local_Plans_Procedural_Guidance.pdf", "/government/uploads/system/uploads/attachment_data/file/531000/Procedural_Practice_in_the_Examination_of_Local_Plans_-_final.pdf"],
+    ["/government/uploads/system/uploads/attachment_data/file/627713/National_Lottery_Distribution_Fund_Investment_Account_2016-2017__web_.pdf", "/government/uploads/system/uploads/attachment_data/file/628399/National_Lottery_Distribution_Fund_Investment_Account_2016-2017__web_.pdf"]
+  ]
+
+  redirects.each do |(redirect_from, redirect_to)|
+    replacement = WhitehallAsset.find_by(legacy_url_path: redirect_to)
+
+    existing_asset = WhitehallAsset.where(legacy_url_path: redirect_from).first
```
floehopper: Minor: I think `WhitehallAsset.find_by` would be OK here - it returns `nil` if no asset found which I think is what you want.

lib/tasks/create_missing_whitehall_assets.rake
```diff
@@ -0,0 +1,53 @@
+# rubocop:disable Metrics/BlockLength
+task create_missing_whitehall_assets: :environment do
+  placeholder_path = Rails.root.join('tmp/whitehall-attachment-placeholder.txt')
```
floehopper: Minor: Using the word "placeholder" might be confusing, because it's used in both Whitehall and Asset Manager for the target of the redirects that happen if an asset has not yet been virus scanned. Perhaps something like `replaced-whitehall-attachment` would be better...?

lib/tasks/create_missing_whitehall_assets.rake
```diff
@@ -0,0 +1,53 @@
...
+  File.open(placeholder_path, 'w') do |file|
+    file.puts('whitehall-attachment-placeholder.txt')
```
floehopper: Minor: It might be nice to add some more useful explanatory text in there, e.g. reference this PR, explain that the attachment has been replaced, etc.

lib/tasks/create_missing_whitehall_assets.rake
```diff
@@ -0,0 +1,53 @@
...
+        puts "Asset identified by #{redirect_from} already exists so there's nothing to do"
+      else
+        puts "Asset identified by #{redirect_from} found but doesn't have matching redirect URL!"
+      end
+    else
+      puts "Asset identified by #{redirect_from} created"
+      WhitehallAsset.create!(
+        legacy_url_path: redirect_from,
+        file: File.open(placeholder_path),
+        replacement: replacement
```
floehopper: I think it's possible (?) that `replacement` might be `nil`, in which case this `create!` call will raise an exception. Is that what you intended?

lib/tasks/create_missing_whitehall_assets.rake
```diff
@@ -0,0 +1,53 @@
...
+      )
+    end
+  end
+
+  FileUtils.rm(placeholder_path)
```
floehopper: Minor: If you intend that the earlier code might raise an exception, you might want to consider putting this in an `ensure` block.

lib/tasks/create_missing_whitehall_assets.rake
```diff
@@ -0,0 +1,53 @@
...
+
+  redirects.each do |(redirect_from, redirect_to)|
+    replacement = WhitehallAsset.find_by(legacy_url_path: redirect_to)
+
+    existing_asset = WhitehallAsset.where(legacy_url_path: redirect_from).first
+    if existing_asset
+      if existing_asset.replacement == replacement
+        puts "Asset identified by #{redirect_from} already exists so there's nothing to do"
+      else
+        puts "Asset identified by #{redirect_from} found but doesn't have matching redirect URL!"
```
floehopper: Minor: I found the wording a bit confusing. I suspect it's left over from when you were using `Asset#redirect_url`. Perhaps it should be something more like: "but doesn't have matching replacement asset"...?

lib/tasks/create_missing_whitehall_assets.rake
```diff
@@ -0,0 +1,53 @@
...
+    ["/government/uploads/system/uploads/attachment_data/file/490383/Local_Plans_Procedural_Guidance.pdf", "/government/uploads/system/uploads/attachment_data/file/531000/Procedural_Practice_in_the_Examination_of_Local_Plans_-_final.pdf"],
+    ["/government/uploads/system/uploads/attachment_data/file/627713/National_Lottery_Distribution_Fund_Investment_Account_2016-2017__web_.pdf", "/government/uploads/system/uploads/attachment_data/file/628399/National_Lottery_Distribution_Fund_Investment_Account_2016-2017__web_.pdf"]
+  ]
+
+  redirects.each do |(redirect_from, redirect_to)|
+    replacement = WhitehallAsset.find_by(legacy_url_path: redirect_to)
+
+    existing_asset = WhitehallAsset.where(legacy_url_path: redirect_from).first
+    if existing_asset
+      if existing_asset.replacement == replacement
```
floehopper: Are you missing the case where `replacement` might be `nil` and you might need to create the replacement...?

lib/tasks/create_missing_whitehall_assets.rake
chrisroos: Unfortunately not. It raises an exception:

```
>> WhitehallAsset.find_by(legacy_url_path: 'foo')
Mongoid::Errors::DocumentNotFound: 
message:
  Document not found for class WhitehallAsset with attributes {:legacy_url_path=>"foo"}.
```

lib/tasks/create_missing_whitehall_assets.rake
chrisroos: This Rake task is going away as soon as it's been run so I'm not worried about the potential confusion.

lib/tasks/create_missing_whitehall_assets.rake
floehopper: Thanks, Mongoid! 😮

lib/tasks/create_missing_whitehall_assets.rake
chrisroos: Good idea! We've updated the text in the commit titled "Improve content of temporary file use in Rake task".

lib/tasks/create_missing_whitehall_assets.rake
chrisroos: Yup. We want it to fail fast if something doesn't work as expected.

lib/tasks/create_missing_whitehall_assets.rake
chrisroos: If the earlier code raises an exception we'd need to fix the problem and then run the script again which would then clear this file.

lib/tasks/create_missing_whitehall_assets.rake
chrisroos: Given that it's a temporary script I'm not worried about the confusion.

lib/tasks/create_missing_whitehall_assets.rake
floehopper: While the Rake task is going away the file content will stay on S3, no?

lib/tasks/create_missing_whitehall_assets.rake
chrisroos: We're not expecting this to happen but if it does then we'll be alerted by the "Asset identified by #{redirect_from} found but doesn't have matching redirect URL!" message. We'll work out what to do about it if it happens.

chrisroos:
Thanks for the review, @floehopper. We've addressed your comments and are planning to merge as soon as the tests have passed.

lib/tasks/create_missing_whitehall_assets.rake
chrisroos: I think the "whitehall-attachment-placeholder.txt" filename will only be present in the database, and there were only 19 assets created with this filename so I'm still not too worried about the potential confusion.

PR #531, opened 2018-03-21
Avoid redirecting to non-accessible assets

chrisroos:
It's possible to replace a public asset with a draft asset. Normal replacement behaviour is to redirect from the old to the new asset; but we shouldn't redirect requests for the public asset in this instance as it'll mean we send the user to an asset they don't have permission to view. Note that we do redirect if the user requested the public asset from the draft-assets host as it means they're authenticated and therefore authorised (barring any additional access-limiting) to view draft assets.

This change follows a similar problem in Whitehall where we were inadvertently redirecting a subset of Attachments to a non-accessible draft version. See https://github.com/alphagov/whitehall/pull/3894 for some more information.

### DIFF
--- a/spec/controllers/media_controller_spec.rb
+++ b/spec/controllers/media_controller_spec.rb
@@ -210,7 +210,7 @@ RSpec.describe MediaController, type: :controller do
           replacement.update_attribute(:draft, true)
         end
 
-        it 'serves the asset when requested via something other than the draft-assets host' do
+        it 'serves the original asset when requested via something other than the draft-assets host' do
           request.headers['X-Forwarded-Host'] = "not-#{AssetManager.govuk.draft_assets_host}"
 
           expect(controller).to receive(:proxy_to_s3_via_nginx).with(asset)
@@ -218,7 +218,7 @@ RSpec.describe MediaController, type: :controller do
           get :download, params
         end
 
-        it 'redirects to the replacement when requested via the draft-assets host' do
+        it 'redirects to the replacement asset when requested via the draft-assets host by a signed-in user' do
           request.headers['X-Forwarded-Host'] = AssetManager.govuk.draft_assets_host
           allow(controller).to receive(:authenticate_user!)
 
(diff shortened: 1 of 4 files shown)


spec/controllers/media_controller_spec.rb
```diff
@@ -204,6 +204,29 @@
 
         expect(response.headers['Cache-Control']).to eq('max-age=86400, public')
       end
+
+      context 'and the replacement is draft' do
+        before do
+          replacement.update_attribute(:draft, true)
+        end
+
+        it 'serves the asset when requested via something other than the draft-assets host' do
```
floehopper: Minor: "serves the asset" -> "serves the original asset"; or "serves the asset" -> "serves the replaced asset"

spec/controllers/media_controller_spec.rb
```diff
@@ -204,6 +204,29 @@
...
+          request.headers['X-Forwarded-Host'] = "not-#{AssetManager.govuk.draft_assets_host}"
+
+          expect(controller).to receive(:proxy_to_s3_via_nginx).with(asset)
+
+          get :download, params
+        end
+
+        it 'redirects to the replacement when requested via the draft-assets host' do
```
floehopper: Minor: "when requested via the draft-assets host" -> "when requested via the draft-assets host by a signed-in user"

floehopper:
Note that I've added some integration tests to Whitehall in https://github.com/alphagov/whitehall/pull/3900 which serve to illustrate the dependency on this new Asset Manager behaviour.

chrisroos:
Thanks @floehopper. I've improved the test names as you've suggested.

> Also I think it might be worth explaining a bit more about the "access-limited" scenario in the commit note / PR description. I think I'm right in saying that given that the parent document would be access-limited, a user without access wouldn't even be able to see the parent document and therefore neither would they be able to see the links to the attachments. So a replacement redirect resulting in a `403 Forbidden` would be a very unlikely scenario and in any case the result would be perfectly sensible.

You're quite right when thinking about this in terms of Whitehall's use of Asset Manager, but there's nothing in Asset Manager itself that prevents you from replacing a non access-limited asset with one that is access-limited, so I'm happy enough with the description as it stands.

Issue #533, opened 2018-03-21
It should be possible to update an asset if its replacement has been deleted


chrisroos:
This stems from https://github.com/alphagov/asset-manager/issues/525#issuecomment-375037308.

You can use this rspec example to see the problem:

```
it 'should be updateable if the replacement has been deleted' do
  asset = FactoryBot.create(:asset)
  replacement = FactoryBot.create(:asset)
  asset.replacement = replacement
  asset.save!

  replacement.delete!

  asset.reload.save!
end
```

The error is:

```
Failure/Error: asset.reload.save!

Mongoid::Errors::Validations:

  message:
    Validation of Asset failed.
  summary:
    The following errors were found: Replacement not found
  resolution:
    Try persisting the document with valid data or remove the validations.
```

chrisroos:
We've addressed this in https://github.com/alphagov/asset-manager/pull/535.

PR #535, opened 2018-03-22
Assets should be valid if their replacement has been deleted

chrisroos:
We saw some exceptions when updating the metadata of Whitehall attachments stored in Asset Manager. Some of these attachments had been replaced by assets that had subsequently been deleted. This left the assets in an invalid state and meant that we couldn't update them.

We considered removing the `replacement_id` foreign key when a replacement is deleted but weren't keen on losing that data. Allowing an asset to be valid if it's been replaced by a deleted asset feels slightly safer.

See https://github.com/alphagov/asset-manager/issues/533 for some more information.

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -162,7 +162,8 @@ protected
   end
 
   def check_specified_replacement_exists
-    if replacement_id.present? && replacement.blank?
+    unscoped_replacement = Asset.unscoped.where(id: replacement_id)
+    if replacement_id.present? && unscoped_replacement.blank?
       errors.add(:replacement, 'not found')
     end
   end
--- a/spec/models/asset_spec.rb
+++ b/spec/models/asset_spec.rb
@@ -60,6 +60,19 @@ RSpec.describe Asset, type: :model do
           expect(asset.errors[:replacement]).to include('not found')
         end
       end
+
+      context 'and the replacement exists but has been deleted' do
+        let(:replacement) { FactoryBot.create(:asset) }
+        let(:replacement_id) { replacement.id.to_s }
+
+        before do
+          replacement.destroy
+        end
+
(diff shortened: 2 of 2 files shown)


spec/models/asset_spec.rb
```diff
@@ -60,6 +60,19 @@
           expect(asset.errors[:replacement]).to include('not found')
         end
       end
+
+      context 'and the replacement exists but has been deleted' do
+        let(:replacement) { FactoryBot.create(:asset) }
+        let(:replacement_id) { replacement.id.to_s }
+
+        before do
+          replacement.destroy
```
floehopper: Minor: I think it would be better to use `replacement.destroy!` so we fail fast if the destroy fails for some unexpected reason.

app/models/asset.rb
```diff
@@ -162,7 +162,8 @@ def file_stat
   end
 
   def check_specified_replacement_exists
-    if replacement_id.present? && replacement.blank?
+    unscoped_replacement = Asset.unscoped.where(id: replacement_id)
```
floehopper: Minor: Using `Asset.unscoped.where` vs `Asset.unscoped.find_by` feels a bit odd here. I think we're only ever expecting at most a single asset to be returned so I think it would be better to use the latter.

spec/models/asset_spec.rb
chrisroos: We can't do that as `#destroy!` actually destroys the object rather than just marking it as deleted.

app/models/asset.rb
chrisroos: We can't easily use `Asset.unscoped.find_by` as that raises an exception when the record doesn't exist.

chrisroos:
Thanks for the review, @floehopper. I'll get this merged.

PR #542, opened 2018-04-04
Allow deleted assets to be retrieved using the API

chrisroos:
We're seeing a number of problems in Whitehall when it's trying to update attachments in Asset Manager. A large number of these are being caused by us not being able to retrieve information about assets that have been deleted. This change should help improve the situation.

More generally, I think this is a reasonable change to make given that there's an API restore endpoint that undeletes a previously deleted asset. I think it's probably useful to be able to query the API to determine the state of the asset before attempting to delete it.

Note that it's not possible to update a previously deleted asset, nor is it possible to request the content of a deleted asset.

### DIFF
--- a/app/controllers/base_assets_controller.rb
+++ b/app/controllers/base_assets_controller.rb
@@ -1,6 +1,6 @@
 class BaseAssetsController < ApplicationController
   def show
-    @asset = find_asset
+    @asset = find_asset(include_deleted: true)
 
     expires_now
     render json: AssetPresenter.new(@asset, view_context)
--- a/app/controllers/whitehall_assets_controller.rb
+++ b/app/controllers/whitehall_assets_controller.rb
@@ -25,8 +25,9 @@ private
     WhitehallAsset.where(legacy_url_path: asset_params[:legacy_url_path])
   end
 
-  def find_asset
-    WhitehallAsset.from_params(
+  def find_asset(include_deleted: false)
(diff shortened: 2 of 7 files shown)


chrislo:
This looks good to me @chrisroos - the only thing I'm worried about is if other applications are relying on the status code returned from the GET action to determine whether an asset is available or not. We talked a bit yesterday about passing a parameter with the request - something like `include_deleted` which we could then add to the GET requets in Whitehall where we need it.

chrisroos:
> the only thing I'm worried about is if other applications are relying on the status code returned from the GET action to determine whether an asset is available or not.

I'm fairly confident that Whitehall is the only app that uses the API to `GET` information about assets, so I don't think we need to worry about this.

chrisroos:
> on second reading, this parameter has only been added to the Whitehall controller's find_asset_by method so as long as nothing else in Whitehall is relying on that status we should be ok?

You're correct that the parameter has only been added to `WhitehallAssetsController#find_asset` method although that's because the same parameter is already present in the equivalent `AssetsController#find_asset` method.

The main change in this branch is the passing of `include_deleted: true` in the call to `find_asset` from `BaseAssetsController#show`. This means that `GET` requests to `/assets\<id>` and to `/whitehall_assets/\<path>` will now return assets in the deleted state where they wouldn't have previously.

chrisroos:
Thanks @chrislo :-)

PR #554, opened 2018-05-15
Set X-Frame-Options: DENY for whitehall assets

### LINKED ISSUE #249: Standardise use of X-Frame-Options response header for Whitehall vs Mainstream assets
Once #246 and alphagov/govuk-puppet#6602 have been merged & deployed and once we are [serving Whitehall assets from Asset Manager](https://github.com/alphagov/asset-manager/milestone/6), Whitehall assets will be served with `X-Frame-Options` set to `SAMEORIGIN` and Mainstream assets will be served with it set to `DENY`. We can probably standardise on one of these values and simplify the Asset Manager code.

barrucadu:
This breaks anything which embeds whitehall assets in an iframe,
however:

- whitehall only uses iframes for external videos
- government-frontend doesn't use iframes at all

So this more conservative header should be safe.

OWASP recommends DENY unless there is a specific reason to use
something else, which is why I didn't change everything to use
SAMEORIGIN instead.

---

Closes #249 

[Trello card](https://trello.com/c/Rx2IiQPR/156-standardise-use-of-x-frame-options-response-header-for-whitehall-vs-mainstream-assets-2)

### DIFF
--- a/app/controllers/base_media_controller.rb
+++ b/app/controllers/base_media_controller.rb
@@ -37,4 +37,8 @@ protected
       headers['Link'] = %(<#{asset.parent_document_url}>; rel="up")
     end
   end
+
+  def add_frame_header
+    headers['X-Frame-Options'] = 'DENY'
+  end
 end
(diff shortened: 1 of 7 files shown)


barrucadu:
That sounds good to me.  Should we deploy this branch to integration, or should I just merge this now and add a note to the trello card to monitor for a bit?

thomasleese:
Fine to merge it now I think!

PR #555, opened 2018-05-15
Reconcile whitehall and mainstream asset serving code

barrucadu:
There's a lot of code duplication between `MediaController::download` and `WhitehallMediaController::download`, which can be reduced by moving the different behaviours to protected methods and letting `MediaController::download` orchestrate everything.

This is a bit less than ideal, as `MediaController` doesn't make use of `temporary_redirect?` and `WhitehallMediaController` doesn't make use of `filename_current?`.  So there's a small bit of complication in the final code, but I think it's worth all the shared bits being specified in only one place.

---

[Trello card](https://trello.com/c/h0bsn8rL/153-rationalise-mainstream-asset-filename-redirect-whitehall-asset-replacement-redirect-3)


app/controllers/media_controller.rb
```diff
@@ -69,4 +70,10 @@ def redirect_to_current_filename
       only_path: true,
     )
   end
+
+  def temporary_redirect?
+    false
+  end
+
+  def perform_temporary_redirect; end
```
thomasleese: We can remove this method since it shouldn't get called if `temporary_redirect?` is `false`. I think it's better anyway for an error to happen if `temporary_redirect?` was changed to `true` without implementing said method.

app/controllers/whitehall_media_controller.rb
```diff
@@ -1,56 +1,31 @@
-class WhitehallMediaController < BaseMediaController
```
thomasleese: Can we get rid of `BaseMediaController` now? Perhaps merge it into `MediaController`?

app/controllers/whitehall_media_controller.rb
```diff
@@ -1,56 +1,31 @@
...
+      path: params[:path], format: params[:format], path_prefix: 'government/uploads/'
+    )
+  end
 
-    if asset.redirect_url.present?
-      redirect_to asset.redirect_url
-      return
-    end
+  def asset_servable?
+    not asset.infected?
```
thomasleese: Although it doesn't matter in this case, I think it's better to use `!` because that has more expected precedence rules than `not` which has the lowest.

https://stackoverflow.com/questions/27429711/can-someone-explain-not-vs-in-ruby

.rubocop_todo.yml
```diff
@@ -69,7 +69,7 @@ RSpec/SubjectStub:
 # Configuration parameters: IgnoreSymbolicNames.
 RSpec/VerifiedDoubles:
   Exclude:
-    - 'spec/controllers/base_media_controller_spec.rb'
+    - 'spec/controllers/media_controller_spec.rb'
```
thomasleese: 😢 but since the old one was in this file I'll let it slip 😉

PR #560, opened 2018-05-23
Add a Rake task to unreplace an asset

barrucadu:
My only thought is: should this be two tasks?  One to unreplace an asset, one to delete one?

---

[Trello card](https://trello.com/c/uIPJV2kI/150-fix-old-replaced-whitehall-attachments-on-discarded-editions-2)

### DIFF
--- a/lib/tasks/govuk_assets.rake
+++ b/lib/tasks/govuk_assets.rake
@@ -16,4 +16,28 @@ namespace :govuk_assets do
     status = result.successful? ? 'OK' : 'Error'
     puts "#{status}: #{result.written_count} documents updated"
   end
+
+  def unreplace_asset(asset)
+    if asset.replacement.nil?
+      puts "#{asset.uuid} has no replacement"
+      return
+    end
+
+    puts "found replacement #{asset.replacement.uuid}"
+
+    asset.replacement.update!(deleted: true)
+    asset.update!(replacement: nil)
+  end
+
+  desc 'Unreplace an asset and delete the replacement'
+  task :unreplace_asset, [:uuid] => :environment do |_, args|
+    asset = Asset.unscoped.find_by!(uuid: args[:uuid])
+    unreplace_asset(asset)
+  end
+
+  desc 'Unreplace a Whitehall asset and delete the replacement'
+  task :unreplace_whitehall_asset, [:legacy_url_path] => :environment do |_, args|
+    asset = WhitehallAsset.unscoped.find_by!(legacy_url_path: args[:legacy_url_path])
+    unreplace_asset(asset)
+  end
 end


lib/tasks/govuk_assets.rake
```diff
@@ -16,4 +16,28 @@ namespace :govuk_assets do
     status = result.successful? ? 'OK' : 'Error'
     puts "#{status}: #{result.written_count} documents updated"
   end
+
+  desc 'Unreplace an asset and delete the replacement'
+  task :unreplace_asset, [:uuid] => :environment do |_, args|
+    uuid = args[:uuid]
+    asset = Asset.unscoped.where(uuid: uuid)
```
thomasleese: This should be `.find_by` I think, `.where` will return a query.

lib/tasks/govuk_assets.rake
```diff
@@ -16,4 +16,27 @@ namespace :govuk_assets do
     status = result.successful? ? 'OK' : 'Error'
     puts "#{status}: #{result.written_count} documents updated"
   end
+
+  desc 'Unreplace an asset and delete the replacement'
+  task :unreplace_asset, [:uuid] => :environment do |_, args|
+    begin
+      uuid = args[:uuid]
+      asset = Asset.unscoped.find_by(uuid: uuid)
```
thomasleese: If you use `find_by!` it can raise an exception if no asset is found and we avoid the subsequent `if` block.

lib/tasks/govuk_assets.rake
```diff
@@ -16,4 +16,27 @@ namespace :govuk_assets do
...
+
+      if asset.replacement.nil?
+        puts "#{uuid} has no replacement"
+        return
+      end
+
+      puts "found replacement #{asset.replacement.uuid}"
+
+      asset.replacement.deleted = true
```
thomasleese: You can use `asset.replacement.update!(deleted: true)` to `update` the attributes and then call `save!`. Incidentally, it's better to use `save!` in a place like this because it will raise an exception if any of the validations fail, whereas `save` will just return `true` or `false`.

lib/tasks/govuk_assets.rake
```diff
@@ -16,4 +16,27 @@ namespace :govuk_assets do
...
+      asset.replacement.save
+
+      asset.replacement = nil
+      asset.save
+    rescue Mongoid::Errors::DocumentNotFound
```
thomasleese: I wouldn't worry too much about catching this exception or showing nice error messages, I think if the Rake task fails, it's fine to show the full traceback because it will be being used by developers.

lib/tasks/govuk_assets.rake
```diff
@@ -16,4 +16,28 @@ namespace :govuk_assets do
...
+
+  def unreplace_asset(asset)
+    if asset.replacement.nil?
+      puts "#{asset.uuid} has no replacement"
+      return
+    end
+
+    puts "found replacement #{asset.replacement.uuid}"
+
+    asset.replacement.upate!(deleted: true)
```
thomasleese: `update!` :)

barrucadu:
I am going to *tentatively* close this, as I don't think it's actually needed.

https://github.com/alphagov/whitehall/pull/4052 deletes invalid whitehall attachments.  When an attachment is deleted, whitehall sends an update to the asset-manager.  So I suspect that running the whitehall task and waiting for all the asset-manager jobs to be processed will be enough.

PR #563, opened 2018-05-30
Improve healthcheck performance

thomasleese:
There is a default scope on assets which makes this query more complicated to execute. If we remove this Mongo can respond to this query much quicker.


kevindew:
I think you can add the index concurrently with mongoid. I seem to remember wishing I'd done that last time I added a mongo index.

thomasleese:
I've actually already added the index, I ran:

```js
db.assets.createIndex({deleted_at: 1}, {background: true})
```

I wanted to add it to the model so we don't lose the index.

PR #583, opened 2018-08-03
Add a rake task to create a whitehall asset

barrucadu:
This is probably useful, and is an indirect stepping stone to HMRC assets

### DIFF
--- a/lib/tasks/govuk_assets.rake
+++ b/lib/tasks/govuk_assets.rake
@@ -16,4 +16,14 @@ namespace :govuk_assets do
     status = result.successful? ? 'OK' : 'Error'
     puts "#{status}: #{result.written_count} documents updated"
   end
+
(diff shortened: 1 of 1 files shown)


lib/tasks/govuk_assets.rake
```diff
@@ -16,4 +16,15 @@ namespace :govuk_assets do
     status = result.successful? ? 'OK' : 'Error'
     puts "#{status}: #{result.written_count} documents updated"
   end
+
+  desc 'Create a whitehall asset with the given legacy URL path'
+  task :create_whitehall_asset, [:file_path, :legacy_url_path] => :environment do |_, args|
+    file_path = args[:file_path]
+    legacy_url_path = args[:legacy_url_path]
+    asset = WhitehallAsset.new(
```
thomasleese: To save lines, you can actually do a `WhitehallAsset.create!(` instead of a `new` and then a `save`.

PR #585, opened 2018-08-03
Do not make assumptions about files in HMRC PAYE task

barrucadu:
The documented list of files does not match the reality.

This also makes it much easier to load all the current files into the asset manager:

```bash
for file in *; do
    sudo -udeploy govuk_setenv asset-manager bundle exec rake govuk_assets:create_hmrc_paye_asset[$file]
done
```

---

[Trello card](https://trello.com/c/SFeIQ9Dp/371-serve-hmrc-assets-from-the-asset-manager)

### DIFF
--- a/lib/tasks/govuk_assets.rake
+++ b/lib/tasks/govuk_assets.rake
@@ -31,10 +31,22 @@ namespace :govuk_assets do
     puts "Uploaded '#{file_path}' to '#{legacy_url_path}'"
   end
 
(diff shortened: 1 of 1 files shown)


barrucadu:
I've only made it do all zip files, as the xml file needs its file name changing when uploaded.

barrucadu:
`create_or_replace_whitehall_asset` already prints out on-disk filenames and legacy URL paths as it uploads things, so I don't think there's a need for extra logging

PR #589, opened 2018-08-14
Return a 404 error when attempting to access a deleted Whitehall asset

brucebolt:
Currently, accessing a soft deleted asset returns a 404.  However, accessing a soft deleted Whitehall asset still attempts to proxy to S3, then returns the resulting XML and 404 proxied back from S3.  This commit changes that behaviour so that soft deleted Whitehall assets are not proxied and a 404 is served from Asset Manager.

Note: the asset media controller uses `find` to get the records, which raises an exception if no records are found.  Whitehall media controller uses `find_by` which does not raise an exception, therefore we have to raise our own exception and rescue with a 404 in the event of no records being returned.

### DIFF
--- a/app/controllers/whitehall_media_controller.rb
+++ b/app/controllers/whitehall_media_controller.rb
@@ -2,10 +2,15 @@ class WhitehallMediaController < MediaController
 
 protected
 
+  class WhitehallAssetNotFound < StandardError
+  end
+
+  rescue_from WhitehallAssetNotFound, with: :error_404
+
   def asset
-    @asset ||= WhitehallAsset.from_params(
+    @asset ||= WhitehallAsset.undeleted.from_params(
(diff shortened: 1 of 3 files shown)


app/controllers/whitehall_media_controller.rb
```diff
@@ -2,10 +2,15 @@ class WhitehallMediaController < MediaController
 
 protected
 
+  class NoRecordsFound < StandardError
```
thomasleese: I think it might be clearer if this was named `WhitehallAssetNotFound` and inherited from the rails `ActiveRecord::RecordNotFound` exception.

brucebolt: I've renamed the exception.  We're using `Mongoid` in this app instead of `ActiveRecord`.  I don't think there's any benefit to inheriting from a Mongoid exception here.

PR #590, opened 2018-08-15
Delete and remove from S3

thomasleese:
This PR adds a Rake task which makes it possible to delete and asset, and remove the asset from S3 if required.

```sh
$ rake assets:delete_and_remove_from_s3[100]
```

Depends on https://github.com/alphagov/govuk-terraform-provisioning/pull/165.

[Trello Card](https://trello.com/c/bod2J0d5/393-make-the-process-of-deleting-assets-off-s3-easier)

### DIFF
--- a/lib/tasks/assets.rake
+++ b/lib/tasks/assets.rake
@@ -1,4 +1,9 @@
 namespace :assets do
+  desc "Mark an asset as deleted"
+  task :delete, [:id] => :environment do |_t, args|
+    Asset.find(args.fetch(:id)).delete
+  end
+
   desc "Mark an asset as deleted and remove from S3"
   task :delete_and_remove_from_s3, [:id] => :environment do |_t, args|
     asset = Asset.find(args.fetch(:id))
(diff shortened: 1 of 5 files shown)


lib/tasks/assets.rake
```diff
@@ -0,0 +1,18 @@
+namespace :assets do
```
barrucadu: Is it worth duplicating these for whitehall assets too?  This approach works, but it means that the user will need to look up the asset by legacy URL path to get its ID first, in which case they may as well just delete it from the app console.

thomasleese: I wanted to avoid having special cases for Whitehall because it's just more to delete when Whitehall eventually goes away. I think it if ends up being a big problem we could add them, but I wanted to experiment with not having them for now.

As it is, conveniently, the most common place this will get used in the docs you already have the ID of the asset easily available even if it's a whitehall asset because it recommends having an `asset` variable ready in the console.

PR #595, opened 2018-09-05
Delete files uploaded to s3 from disk

barrucadu:
This will allow us to shrink the asset NFS share with confidence, as the disk space used by assets will no longer grow continuously.

---

[Trello card](https://trello.com/c/M3Dfrj8S/443-remove-attachments-from-nfs)


barrucadu:
This is actually unnecessary, as it happens already!

https://github.com/alphagov/asset-manager/blob/3d1496932b06f7ee965f3e7ba2124b8dccc8a37b/app/models/asset.rb#L80-L83

*However*, 12GB of empty directories have built up.

PR #608, opened 2018-11-20
View draft assets with token authentication

kevindew:
Trello: https://trello.com/c/dawXVwdW/439-asset-manager-files-are-not-available-when-using-access-tokens-or-if-someone-doesnt-have-asset-manager-permission

This is a feature that allows the presence of a token, in either a cookie or a query string, to bypass the authentication needed to view that asset on the draft stack. See https://github.com/alphagov/authenticating-proxy/pull/108 for more info.

This allows the storing of auth_bypass_ids on an asset and allows checking JWT tokens when requesting a draft asset to determine whether the asset should be served or not.

### DIFF
--- a/spec/controllers/assets_controller_spec.rb
+++ b/spec/controllers/assets_controller_spec.rb
@@ -32,6 +32,13 @@ RSpec.describe AssetsController, type: :controller do
         expect(assigns(:asset).access_limited).to eq(['user-id'])
       end
 
+      it 'stores auth_bypass_ids on asset' do
+        attributes = valid_attributes.merge(auth_bypass_ids: %w[id1 id2])
+        post :create, params: { asset: attributes }
+
+        expect(assigns(:asset).auth_bypass_ids).to eq(%w[id1 id2])
+      end
+
       it 'stores parent_document_url on asset' do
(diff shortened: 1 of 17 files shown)


spec/controllers/assets_controller_spec.rb
```diff
@@ -206,6 +213,24 @@
...
 
+      it 'stores auth_bypass_ids on existing asset' do
+        attributes = valid_attributes.merge(auth_bypass_ids: ['bypass-id'])
+        put :update, params: { id: asset.id, asset: attributes }
+
+        expect(assigns(:asset).auth_bypass_ids).to eq(['bypass-id'])
+      end
+
+      it 'copes when auth_bypass_ids are passed in as an empty string' do
+        asset.update_attributes!(auth_bypass_ids: %w[bypass-1 bypass-2])
```
leenagupte: Is this line needed?

kevindew: Yup, as by default an asset will have an empty array as auth_bypass_ids and thus this is set to a populated array to test that it is modified to be an empty array

kevindew:
Thanks @leenagupte 👍

PR #610, opened 2018-11-23
Authenticate user on draft host, even if asset doesn't exist

barrucadu:
This is a useful behaviour so that we can check requesting the draft host sends you to signon.


app/controllers/media_controller.rb
```diff
@@ -52,6 +52,12 @@ def download
     add_link_header(asset)
     add_frame_header
     proxy_to_s3_via_nginx(asset)
+  rescue Mongoid::Errors::DocumentNotFound => e
+    if requested_from_draft_assets_host? && !authenticated?
+      authenticate_user!
+    else
+      raise e
```
thomasleese: Small nitpick, but to re-raise an exception you can just have `raise` on its own.

PR #653, opened 2019-05-10
Create a database user seed

thomasleese:
This will be used in the e2e tests combined with GDS SSO mock mode to enable authenticated access to the API without needing to pass in a bearer token.

This is based off the one in email-alert-api: https://github.com/alphagov/email-alert-api/blob/master/db/seeds.rb

Once this is merged, it should enable https://github.com/alphagov/travel-advice-publisher/pull/602 to pass correctly.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,7 +1,7 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the rake db:seed (or created alongside the db with db:setup).
-#
-# Examples:
-#
-#   cities = City.create([{ name: 'Chicago' }, { name: 'Copenhagen' }])
-#   Mayor.create(name: 'Emanuel', city: cities.first)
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -1,7 +1,9 @@
...
-#   Mayor.create(name: 'Emanuel', city: cities.first)
+return if User.where(name: "Test user").present?
+
+gds_organisation_id = "af07d5a5-df63-4ddc-9383-6a666845ebe9"
+
+User.create!(
+  name: "Test user",
+  permissions: %w[signin],
+  organisation_content_id: gds_organisation_id,
+)
```
cbaines: Adding the check on to the end of this expression might be neater, especially if more things are added to the db/seeds.rb file in the future. Also, maybe use `.exists?`?

So something like `) unless User.where(name: "Test user").exists?`

thomasleese: Yep, that is much nice... I've amended the commit to include that change.

PR #654, opened 2019-05-16
Preserve draft redirect query

benthorner:
https://trello.com/c/xpS0zlU4/803-download-an-attachment-file-through-content-publisher

This ensures we preserve any query params when downloading draft asset, which supports anonymous preview of draft assets via Content Publisher.

### DIFF
--- a/spec/controllers/media_controller_spec.rb
+++ b/spec/controllers/media_controller_spec.rb
@@ -206,7 +206,7 @@ RSpec.describe MediaController, type: :controller do
         it "redirects to the new file name" do
           get :download, params: { id: asset, filename: old_file_name }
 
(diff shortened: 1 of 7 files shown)


spec/controllers/media_controller_spec.rb
```diff
@@ -206,7 +206,7 @@ def download
         it "redirects to the new file name" do
           get :download, params: { id: asset, filename: old_file_name }
 
-          expect(response.location).to match(%r(\A/media/#{asset.id}/asset.png))
+          expect(response.location).to match(download_media_path(id: asset, filename: "asset.png"))
```
1pretz1: Nice cleanup ⭐️

PR #668, opened 2019-07-02
Remove request test for positive virus scan

kevindew:
Trello: https://trello.com/c/hwMoAOau/3-%F0%9F%90%B3-get-all-apps-working-on-docker

This test no longer runs successfully on GDS issued laptops as the ESET
software detects the contents as a virus and quarantines the file, which
is very hard to debug.

This test is also rather strange as it will still pass even if the file
is not identified as having a virus. This is because the endpoint will
return a 404 until a file has been uploaded to cloud storage.

### DIFF
--- a/spec/requests/virus_scanning_spec.rb
+++ b/spec/requests/virus_scanning_spec.rb
@@ -32,40 +32,4 @@ RSpec.describe "Virus scanning of uploaded images", type: :request, disable_clou
     get cloud_url
     expect(response).to have_http_status(:success)
   end
-
-  # Extension to UploadedFile to represent an uploaded virus
-  # without having to have a 'virus' committed into source control.
-  #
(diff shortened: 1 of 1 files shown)


benthorner:
I'm in two minds about this: it's good to remove flakey things, but it looks like there aren't any other tests that cover Asset Manager actually doing any virus scanning. Do you think that's important?

kevindew:
> I'm in two minds about this: it's good to remove flakey things, but it looks like there aren't any other tests that cover Asset Manager actually doing any virus scanning. Do you think that's important?

The worker is unit tested - so all the virus scanning code is covered.

This test itself is bogus as it succeeds even if the file passes the virus scan.

kevindew:
> Fair enough - is the 'success' can similarly misleading?

No, the success case tests enough steps.

PR #677, opened 2019-07-25
Support access limits by user organisation id

theseanything:
This PR enables assets to be access limited by the user's organisation id by the following changes:

- Rename the existing `access_limited` field on the Asset model to `access_limited_user_ids`
- Add migration to update existing documents to the new field name.
- Add new field called `access_limited_organisation_ids` to Asset model to store permitted user organisations
- Supports existing `access_limited` parameter passed via API calls to override `access_limited_user_ids`. This is to prevent breaking API changes.


spec/models/asset_spec.rb
```diff
@@ -232,6 +232,18 @@
       asset = FactoryBot.build(:asset, draft: true, access_limited_user_ids: ['another-user-id'])
       expect(asset).not_to be_accessible_by(user)
     end
+
+    it 'returns true if the asset is draft and access limited and the users org is authorised to view it' do
```
benthorner: Minor, but this feels like it wants a context, but I understand it's consistent with the one above :-).

app/controllers/base_assets_controller.rb
```diff
@@ -24,8 +24,8 @@ def base_asset_params
         asset[:redirect_url] = nil
       end
 
-      if asset.has_key?(:access_limited) && asset[:access_limited].empty?
-        asset[:access_limited] = []
+      if asset.has_key?(:access_limited_user_ids) && asset[:access_limited_user_ids].empty?
```
benthorner: I'm a bit confused why we're doing this, and why it doesn't apply to the new organisation array. Could you help me understand?

spec/controllers/assets_controller_spec.rb
```diff
@@ -25,13 +25,27 @@
...
+      end
+
       it 'stores access_limited_user_ids on asset' do
         attributes = valid_attributes.merge(access_limited_user_ids: ['user-id'])
         post :create, params: { asset: attributes }
 
         expect(assigns(:asset).access_limited_user_ids).to eq(['user-id'])
       end
 
+      it 'stores access_limited_organisation_ids on asset' do
```
benthorner: Does this belong in a different commit?

spec/controllers/assets_controller_spec.rb
```diff
@@ -25,13 +25,27 @@
         expect(assigns(:asset).file.path).to match(/asset\.png$/)
       end
 
+      it 'stores access_limited as access_limited_user_ids on asset' do
```
benthorner: Do we need to cope with the weird empty-string-array case too?

app/models/asset.rb
```diff
@@ -87,10 +89,9 @@ class Asset
   end
 
   def accessible_by?(user)
-    return true unless draft?
-    return true if access_limited.empty?
+    return true unless draft? && access_limited?
```
benthorner: I think the `access_limited?` question needs to be inverted here - it's accessible if it's *not* access limited :-).

app/controllers/base_assets_controller.rb
theseanything: > why we're doing this

This is was the existing functionality - keeping it to prevent breaking changes

> why it doesn't apply to the new organisation array

we are doing the same for organisations ids on lines 35-37

app/models/asset.rb
theseanything: I think the logic is right - unless (i.e. if not) `draft` AND `access limited` then return true

spec/controllers/assets_controller_spec.rb
theseanything: Yup moved to previous commit

spec/controllers/assets_controller_spec.rb
theseanything: Added test case

app/models/asset.rb
benthorner: Oops - smug logic smile vanquished :-|.

kevindew: @benthorner this brings back a memory 😄 I stand by "unless" and an "and" read wrong 😉

spec/models/asset_spec.rb
theseanything: Fixed - refactored tests using contexts

theseanything:
> * It looks like this will perform a sequential operation over each Asset, which could take a long time and potentially disrupt other ops; there [seem to be ways to make this more efficient](https://stackoverflow.com/questions/9254351/how-can-i-rename-a-field-for-all-documents-in-mongodb)

From testing locally and in integration it appears that the operation isn't taking longer than a minute - `RenameAccessLimitField: migrated (19.4255s)`. This relatively short and should have limited impact on other operations.

> * It looks like Mongo isn't very transaction-friendly, and we could [end up with inconsistent data](https://stackoverflow.com/questions/4813160/mongodb-1-6-5-how-to-rename-field-in-collection) if we're not careful

The [rename action for individual documents appear atomic](https://github.com/mongodb/mongoid/blob/664f8f66a8ef3292dd4e1023698034ea9bd5d621/lib/mongoid/persistable/renamable.rb#L27). We still have the potential issue where the collection is partially updated - (due to lack of transactions). However, I feel okay moving forward with this as the risk of the operation being interrupted/not completing is low and it doesn't take long to re-run.

spec/models/asset_spec.rb
```diff
@@ -211,26 +211,48 @@
...
-      asset = FactoryBot.build(:asset, draft: true, access_limited: ['user-id'])
-      expect(asset).to be_accessible_by(user)
+    context 'asset is a draft thats access_limited' do
+      let(:asset) do
+        FactoryBot.build(
+          :asset, draft: true, access_limited_user_ids: ['user-id'], access_limited_organisation_ids: ['org-id']
+)
+      end
+
+      it 'returns true if users id is authorised to view it' do
```
benthorner: Minor, but we usually go to double quotes in order to be grammatical ;-).

app/controllers/base_assets_controller.rb
```diff
@@ -24,8 +24,16 @@ def base_asset_params
         asset[:redirect_url] = nil
       end
 
-      if asset.has_key?(:access_limited) && asset[:access_limited].empty?
-        asset[:access_limited] = []
+      if asset.has_key?(:access_limited) && asset[:access_limited].present?
+        asset[:access_limited_user_ids] = asset[:access_limited]
+      end
+
+      if asset.has_key?(:access_limited_user_ids) && asset[:access_limited_user_ids].empty?
```
benthorner: I think we're missing the old 'reset' behaviour when the old `access_limited` param is empty? Looks like there's no longer a (reset) test for it either, but since the attribute is still allowed we should support this case as well. We should be able to do it in one like this using `to_a`.

benthorner:
...forgot to say thanks for looking into the migration behaviour. Might still be worth letting Platform Health know this is happening, in case it causes a problem/ticket.

spec/models/asset_spec.rb
theseanything: fixed!

app/controllers/base_assets_controller.rb
theseanything: Yup! So I think the "reset" behaviour applies to the new `access_limited_user_ids` with this if statement. And the old  `access_limited` param also adopts this behaviour as its simply mapped to `access_limited_user_ids` in lines 27 to 29. Should be covered by this [test](https://github.com/alphagov/asset-manager/pull/677/files#diff-d438e8bf052be84189bf1bcbed65e9e4R35).

benthorner: Hmm...I'm still unsure and I'll try to explain why:

   - In the code, we only do the mapping behaviour if the old `access_limited` param is `present?`, but the edge case here is when it's `''`, which isn't "present"
   - In the [test](https://github.com/alphagov/asset-manager/pull/677/files#diff-d438e8bf052be84189bf1bcbed65e9e4R35), I think we're just seeing [the default behaviour for the asset](https://github.com/alphagov/asset-manager/blob/master/app/models/asset.rb#L35) (i.e. the existing test was dodgy)

The test that's notably missing is [the reset one for `access_limited_user_ids`](https://github.com/alphagov/asset-manager/pull/677/files#diff-d438e8bf052be84189bf1bcbed65e9e4R233), but using the old name. Am I missing something?

theseanything: Yup - I've fixed resetting an existing parameter using a blank string on the access_limit param and added test case.

PR #679, opened 2019-08-01
Support access limits by user organisation id without migrations

theseanything:
Following #677 (which was reverted due to incorrect migrations), this PR aims to support access limiting by organisation id without modifying the existing data schema. 

This PR adds a new field called access_limited_organisation_ids to Asset model to store permitted user organisations.

This also introduces a new parameter `access_limited_user_ids` which maps to `access_limited`

### DIFF
--- a/spec/controllers/assets_controller_spec.rb
+++ b/spec/controllers/assets_controller_spec.rb
@@ -32,6 +32,13 @@ RSpec.describe AssetsController, type: :controller do
         expect(assigns(:asset).access_limited).to eq(['user-id'])
       end
 
+      it 'stores access_limited blank string as empty array on access_limited' do
+        attributes = valid_attributes.merge(access_limited: '')
+        post :create, params: { asset: attributes }
+
+        expect(assigns(:asset).access_limited).to eq([])
+      end
+
       it 'stores access_limited_organisation_ids on asset' do
         attributes = valid_attributes.merge(access_limited_organisation_ids: ['org-id'])
         post :create, params: { asset: attributes }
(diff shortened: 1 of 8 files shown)


spec/controllers/assets_controller_spec.rb
```diff
@@ -25,13 +25,41 @@
...
+      end
+
+      it 'stores access_limited_user_ids blank string as empty array on access_limited' do
+        attributes = valid_attributes.merge(access_limited_user_ids: '')
+        post :create, params: { asset: attributes }
+
+        expect(assigns(:asset).access_limited).to eq([])
+      end
+
+      it 'stores access_limited_organisation_ids on asset' do
```
benthorner: I think we're missing a test for the empty string case with this new attribute.

spec/controllers/assets_controller_spec.rb
```diff
@@ -202,7 +230,18 @@
         expect(assigns(:asset).access_limited).to eq(['user-id'])
       end
 
-      it 'resets access_limited to an empty array for an existing asset with an access_limited array' do
+      it 'resets access_limits an existing asset with a blank acess_limited_user_ids param' do
```
benthorner: Minor, but small grammar issue with 'resets access_limits an'.

theseanything: Fixed!

spec/controllers/assets_controller_spec.rb
theseanything: Is it covered by the test case on lines 49-54?

PR #693, opened 2019-10-11
Allow an auth bypass token to grant access to all draft assets

AgaDufrat:
## User story
**As a** burgeoning fact checker
**I want** to check content related to step by steps and see the assets
**So that** the pages don't look broken

## What
Opens up access to draft assets when the `draft_asset_manager_access` value is present in the JWT payload but will not grant access to draft assets that are marked as access limited to users or organisations.

## Why
Asset manager should be aware of a flag which was added in https://github.com/alphagov/collections-publisher/pull/815 that can be set in an auth_bypass_token that allows users access to draft assets.

This is a key part in allowing assets to be available as part of a multi-page fact check. This allows a user access to draft assets they normally wouldn’t have access to.

[Trello card](https://trello.com/c/DoD8SPbI/145-in-asset-manager-allow-an-auth-bypass-token-to-grant-access-to-all-draft-assets)

####  Co-authored by @brucebolt

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -2,6 +2,7 @@
 
 class MediaController < ApplicationController
   skip_before_action :authenticate_user!
+  before_action :set_token_payload
 
   def download
     if redirect_to_draft_assets_host_for?(asset)
@@ -59,8 +60,9 @@ class MediaController < ApplicationController
 protected
 
   def has_bypass_id_for_asset?(asset)
-    token = params.fetch(:token, cookies[:auth_bypass_token])
-    asset.valid_auth_bypass_token?(token)
+    return false if @token_payload.nil?
+
+    asset.valid_auth_bypass_token?(@token_payload["sub"])
   end
 
   def authorized_for_asset?(asset)
@@ -68,12 +70,20 @@ protected
 
     return true if has_bypass_id_for_asset?(asset)
 
+    return true if draft_asset_manager_access? && !asset.access_limited?
+
     authenticate_user!
     asset.accessible_by?(current_user)
   end
 
   def requested_from_draft_assets_host?
     request.host == AssetManager.govuk.draft_assets_host
+  end
+
+  def draft_asset_manager_access?
+    return false if @token_payload.nil?
+
+    @token_payload["draft_asset_manager_access"] == true
   end
 
   def proxy_to_s3_via_nginx(asset)
@@ -147,5 +157,15 @@ protected
 
   def temporary_redirect?
     false
+  end
+
+  def set_token_payload
+    token = params.fetch(:token, cookies[:auth_bypass_token])
+    @token_payload = if token
+                       secret = Rails.application.secrets.jwt_auth_secret
+                       JWT.decode(token, secret, true, algorithm: "HS256").first
+                     end
+  rescue JWT::DecodeError
+    @token_payload = nil
   end
 end
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -94,14 +94,8 @@ class Asset
     access_limited.include?(user.uid) || access_limited_organisation_ids.include?(user.organisation_content_id)
   end
 
-  def valid_auth_bypass_token?(token)
-    payload, = JWT.decode(token,
-                          Rails.application.secrets.jwt_auth_secret,
-                          true,
-                          algorithm: "HS256")
-    payload["sub"].present? && auth_bypass_ids.include?(payload["sub"])
-  rescue JWT::DecodeError
-    false
+  def valid_auth_bypass_token?(auth_bypass_id)
+    auth_bypass_ids.include?(auth_bypass_id)
   end
 
   def public_url_path
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -185,11 +185,11 @@ class Asset
     deleted_at.present?
   end
 
-protected
-
   def access_limited?
     access_limited.any? || access_limited_organisation_ids.any?
   end
+
+protected
 
   def store_metadata
     self.etag = etag_from_file
--- a/spec/controllers/media_controller_spec.rb
+++ b/spec/controllers/media_controller_spec.rb
@@ -303,31 +303,55 @@ RSpec.describe MediaController, type: :controller do
     end
 
     context "with an access limited draft asset" do
-      let(:user) { FactoryBot.build(:user) }
-      let(:asset) { FactoryBot.create(:uploaded_asset, draft: true) }
-      let(:scope) { double(:undeleted_scope) }
+      context "when a user is authenticated" do
+        let(:user) { FactoryBot.build(:user) }
+        let(:asset) { FactoryBot.create(:uploaded_asset, draft: true) }
+        let(:scope) { double(:undeleted_scope) }
 
-      before do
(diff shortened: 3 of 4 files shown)


app/controllers/media_controller.rb
```diff
@@ -76,6 +74,22 @@ def requested_from_draft_assets_host?
     request.host == AssetManager.govuk.draft_assets_host
   end
 
+  def draft_asset_manager_access?
+    token = params.fetch(:token, cookies[:auth_bypass_token])
+    payload = decode_auth_bypass_token(token)
```
kevindew: It might be worth having this set as part of a before action or in a method of it's own so it doesn't need to be replicated. You could have something like:
```
before_action :set_token_payload

def set_token_payload
  token = params.fetch(:token, cookies[:auth_bypass_token])
  @token_payload = if token
                     secret = Rails.application.secrets.jwt_auth_secret
                     JWT.decode(token, secret, true, algorithm: "HS256").first
                   end
rescue JWT::DecodeError
  @token_payload = nil
end
```

app/controllers/media_controller.rb
```diff
@@ -56,18 +59,13 @@ def download
...
-    token = params.fetch(:token, cookies[:auth_bypass_token])
-    asset.valid_auth_bypass_token?(token)
-  end
-
   def authorized_for_asset?(asset)
     return true unless requested_from_draft_assets_host?
 
     return true if has_bypass_id_for_asset?(asset)
 
+    return true if draft_asset_manager_access? && asset.access_limited? == false
```
kevindew: You might want to change this to: `return true if draft_asset_manager_access? && !asset.access_limited?`

app/controllers/media_controller.rb
```diff
@@ -76,6 +74,22 @@ def requested_from_draft_assets_host?
...
+  def draft_asset_manager_access?
+    token = params.fetch(:token, cookies[:auth_bypass_token])
+    payload = decode_auth_bypass_token(token)
+
+    return false if payload == false
+
+    payload["draft_asset_manager_access"].present? && payload["draft_asset_manager_access"] == true
+  end
+
+protected
```
kevindew: It's somewhat unconventional to expose controller helper methods as public methods. I assume that this has been done as this is now a monster controller with bags of logic that is hard to test?

I'd suggest that you instead keep the methods protected but test them through actions that exercise the controller. 

For examples if you see in the media_controller_spec how the tests run a request and check a response: https://github.com/alphagov/asset-manager/blob/10537aa1437a18b3416e61c86c9bf19093bd3c44/spec/controllers/media_controller_spec.rb#L526-L528

app/controllers/media_controller.rb
```diff
@@ -76,6 +74,22 @@ def requested_from_draft_assets_host?
     request.host == AssetManager.govuk.draft_assets_host
   end
 
+  def draft_asset_manager_access?
+    token = params.fetch(:token, cookies[:auth_bypass_token])
+    payload = decode_auth_bypass_token(token)
+
+    return false if payload == false
+
+    payload["draft_asset_manager_access"].present? && payload["draft_asset_manager_access"] == true
```
kevindew: As you're checking for true here you could simplify this whole line down to `payload["draft_asset_manager_access"] == true`

app/controllers/media_controller.rb
```diff
@@ -148,4 +158,14 @@ def redirect_to_current_filename
   def temporary_redirect?
     false
   end
+
+  def set_token_payload
+    token = params.fetch(:token, cookies[:auth_bypass_token])
+    @token_payload = if token
+                       secret = Rails.application.secrets.jwt_auth_secret
+                       JWT.decode(token, secret, true, algorithm: "HS256").first
```
ChrisBAshton: Wondering why `.first` is needed here 🤔 `JWT.decode` takes a token and returns the unencrypted payload, not an array...?

spec/controllers/media_controller_spec.rb
```diff
@@ -366,6 +379,17 @@ def download
         end
       end
 
+      context "when a user is not authenticated and has provided a valid token with draft_asset_manager_access" do
+        before { not_logged_in }
+
+        it "grants access to the file" do
+          expect(controller).not_to receive(:authenticate_user!)
+          query_params = params.tap { |p| p[:params].merge!(token: token_with_draft_asset_manager_access) }
```
ChrisBAshton: Couldn't this be simplified to this...?

```ruby
          query_params = params.merge(token: token_with_draft_asset_manager_access)
```

app/models/asset.rb
```diff
@@ -185,12 +179,12 @@ def deleted?
     deleted_at.present?
   end
 
-protected
-
   def access_limited?
     access_limited.any? || access_limited_organisation_ids.any?
   end
 
+protected
```
ChrisBAshton: I wonder if it's worth writing unit tests for `access_limited?` now that it's a public function...? 🤔

app/controllers/media_controller.rb
AgaDufrat: If I remove it, I'm getting: 🤷‍♀ 
```
asset.valid_auth_bypass_token?(@token_payload["sub"])
no implicit conversion of String into Integer
```
and the previous version was also taking the first element of an array by adding the comma after payload 
https://github.com/alphagov/asset-manager/blob/afcd1f5b0f05a5ad42f11d208d178c06c494c655/app/models/asset.rb#L98

spec/controllers/media_controller_spec.rb
AgaDufrat: I'm getting `unknown keyword: token` 
I copied that line from another test 🙈

app/controllers/media_controller.rb
ChrisBAshton: 🤢 well `.first` is much better than an easy to miss `,` 😄

ChrisBAshton: 🤢 well `.first` is much better than an easy to miss `,` 😄

spec/controllers/media_controller_spec.rb
ChrisBAshton: Ah I misunderstood what it's doing. It's merging `token` into the `:params` field of the `params` hash - but my code sample would merge `:token` as a top level field. The use of `tap` makes sense now.

app/models/asset.rb
ChrisBAshton: Nice that you kept this in its own commit 👍 but I'd add to the commit message to explain why it's being made public.

spec/models/asset_spec.rb
```diff
@@ -1116,4 +1116,22 @@
...
+
+    it "is true if it is an entry in access_limited array" do
+      asset.access_limited = %w[123]
+      expect(asset.access_limited?).to be true
+    end
+
+    it "is true if it is an entry in access_limited_organisation_ids array" do
+      asset.access_limited_organisation_ids = %w[435]
+      expect(asset.access_limited?).to be true
+    end
```
ChrisBAshton: These look like they could do with a proof read 🤔 

```
  it "is false if the `access_limited` and `access_limited_organisation_ids` arrays are empty" do
  it "is true if there is a non-empty `access_limited` array"
  it "is true if there is a non-empty `access_limited_organisation_ids` array"
```

PR #703, opened 2019-11-13
Add the option to use username and password with mongodb

fredericfran-gds:
There is a desire to move from mongodb from AWS documentdb but
documentdb requires compulsory authentication.

This PR attempts to modify config of the mongoid to read
the username and password from the environment.
If the environment variables are not set, we use null
to hopefully login as before to the non-authenticated
mongodb.


fredericfran-gds:
no longer required

PR #708, opened 2019-11-29
Enable rails-specific rules for rubocop-govuk

benjamineskola:
trello: https://trello.com/c/Lijki8aj/1584-3-enable-rails-rubocop-rules-via-rubocop-govuk-on-all-our-rails-apps

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -1,6 +1,8 @@
 require "virus_scanner"
 require "services"
 
+# Mongoid doesn't seem to have find_each so rubocop would break it
+# rubocop:disable Rails/FindEach
 class Asset
   include Mongoid::Document
   include Mongoid::Timestamps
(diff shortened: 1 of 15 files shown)


app/models/asset.rb
```diff
@@ -1,6 +1,8 @@
 require "virus_scanner"
 require "services"
 
+# Mongoid doesn't seem to have find_each so rubocop would break it
+# rubocop:disable Rails/FindEach
```
issyl0: If this will never work in this app because Mongoid is weird, could we globally disable it (in `.rubocop.yml`) with the same comment? There may be some disadvantages, like if we ever switch this app away from Mongoid to a different ORM that doesn support it, we might forget to re-enable this cop?

app/models/user.rb
```diff
@@ -15,11 +15,14 @@ class User
   field "organisation_content_id", type: String
   field "organisation_slug", type: String
 
+  # rubocop:disable Rails/FindBy
```
issyl0: Same as above.

PR #750, opened 2020-04-08
Fix transitive redirection breaking with draft assets

barrucadu:
Redirecting to replacements is done in media_controller.rb, as the
first thing in the download method:

    if asset.replacement.present? && (!asset.replacement.draft? || requested_from_draft_assets_host?)
      set_default_expiry
      redirect_to_replacement_for(asset)
      return
    end

In prose, redirection to a replacement happens if and only if these
two requirements are met:

1. The asset has a replacement.
2. The replacement is not a draft (or it is being accessed from the
   draft host).

So if a publisher creates a new draft of a document, and replaces some
assets, the live assets won't redirect until the document is published
and the assets become live.  So far so good.

The problem is to do with `asset.replacement`.  When an asset is
replaced, anything which has the original asset as its replacement is
updated to point to the new replacement.  This is done in asset.rb:

    def update_indirect_replacements
      return unless replacement.present?

      Asset.unscoped.where(replacement_id: self.id).each do |asset|
        asset.replacement = replacement
        asset.save
      end
    end

This behaviour was implemented 2 years ago (by me) in commit 93b225e
to prevent long chains of redirects.  Web browsers typically impose a
limit on the number of redirects they follow, so this feature is
needed.

So this is how the problem occurs:

1. Asset 1 is created and published
2. A new draft is created, which replaces Asset 1 with Asset 2.
   Asset 1 does not redirect because Asset 2 is a draft.
3. The draft is published.
   Asset 1 now redirects to Asset 2.
4. A new draft is created, which replaces Asset 2 with Asset 3.
   Asset 2 does not redirect because Asset 3 is a draft.
5. Asset Manager automatically updates Asset 1 to point directly to Asset 3.
   Asset 1 no longer redirects to Asset 2 because it has no connection to Asset 2.
   Asset 1 doesn't redirect to Asset 3 because Asset 3 is a draft.
   So Asset 1 is visible again!  Oh no!

The solution has 2 parts:

1. Make `update_indirect_replacements` only apply when the replacement
   is live.

2. Add another `after_save` handler to update the `replacement_id` of
   grandchild assets when an asset is published.

The new handler will kick off `update_indirect_replacements`, though
I'm not sure if that's strictly necessary any more.  Since the
transitive closure logic has been in place for two years now, any
asset replaced since then will have had any redirect chains collapsed
to one step already.  So `update_indirect_replacements` only does
something useful for assets which haven't been touched since then.

---

[Trello card](https://trello.com/c/dsIeWe0W/1904-5-plan-how-to-address-issues-with-asset-manager-drafts)

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -59,7 +59,8 @@ class Asset
 
   before_save :store_metadata, unless: :uploaded?
   after_save :schedule_virus_scan
-  after_save :update_indirect_replacements
+  after_save :update_indirect_replacements_on_publish
+  after_save :backpropagate_replacement
 
   scope :deleted, -> { where(:deleted_at.ne => nil) }
   scope :undeleted, -> { where(deleted_at: nil) }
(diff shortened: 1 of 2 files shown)


app/models/asset.rb
```diff
@@ -154,12 +155,23 @@ def size_from_file
...
+    end
+  end
+
+  def backpropagate_replacement
+    return if replacement.blank? || replacement.draft?
+
+    Asset.where(replacement_id: self.id).each do |replaced_by_me|
+      replaced_by_me.replacement = self
+      replaced_by_me.save
     end
```
cbaines, line 175: I'm not sure this here changes anything... for assets where they're replaced by this asset, it sets them to be replaced by this asset?

barrucadu: Whoops, that should be `replaced_by_me.replacement = self.replacement`

cbaines: Ah, cool :+1:

PR #770, opened 2020-06-02
Sync this repo with RuboCop GOV.UK

benthorner:
https://github.com/alphagov/rubocop-govuk/issues/73

This removes all custom config and fixes the resulting
linting issues. Please see the commits for more details.


app/controllers/assets_controller.rb
```diff
@@ -13,22 +13,14 @@ def update
 
   def destroy
     @asset = find_asset
-
-    if @asset.destroy
-      render json: AssetPresenter.new(@asset, view_context).as_json(status: :success)
-    else
-      error 422, @asset.errors.full_messages
-    end
+    @asset.destroy
```
kevindew, line 16: I saw from the commit message about switching this to a 500 which makes sense, what I wondered is whether we're confident this definitely produces a 500?

If this was ActiveRecord the `ActiveRecord::RecordInvalid` exception would be caught and converted to a 422 by action_dispatch.rescue_responses by default and it looks like Mongoid applies the same in a Railtie: https://github.com/mongodb/mongoid/blob/3cbbcfb42b091261f7f255959c48d98a1b5948cb/lib/mongoid/railtie.rb#L26

spec/lib/s3_storage/fake_spec.rb
```diff
@@ -18,17 +18,17 @@
...
 
-  context "saving a file" do
+  context "when saving a file" do
     let(:asset_path) { root_path.join(relative_path_to_asset) }
 
     it "writes file to fake S3 storage directory" do
       storage.save(asset)
 
-      expect(File.exist?(asset_path)).to be_truthy
+      expect(File).to exist(asset_path)
```
kevindew, line 27: That's a lot nicer.

app/controllers/assets_controller.rb
benthorner: That's true. I was naively assuming that most records would be valid, and the text in the (now) deleted test suggested the developer was more concerned about a scenario where the DB operation itself failed, although admittedly that wouldn't populate `errors`!

How about I rephrase the commit to be more: it'll do the right thing without us testing it?

benthorner:
Bump 😸

app/controllers/assets_controller.rb
kevindew: Oh so sorry for the long delay.

The commit message change sounds reasonable, every other route I consider seems to end up in massive rabbit hole.  It may mean a client gets a 422 in an unexpected format, but then no-one can do anything about a 422 here as it's the system state that is wrong (and thus the rabbit hole).

benthorner:
Testing on Integration (creating a doc with an image and then removing it) works fine.

PR #771, opened 2020-06-04
Upgrade Rails to 6, Ruby to 2.7.1, and other dependencies

catalin-ursachi:
As part of Softwire's investigation for [this opportunity](https://www.digitalmarketplace.service.gov.uk/digital-outcomes-and-specialists/opportunities/12323), I have begun to upgrade the Asset Manager to Ruby 2.7 and Rails 6; so I've raised this pull request in case it's useful! 🙂 Below is a summary of the changes made and the testing done:

### Upgrading Rails

* Upgraded the `rails` gem version to 6.0.3.1; ran `bundle update --conservative`.
* Ran `app:update` - which updated some of the auto-generated Rails files, and created some new ones.
* Deleted the transitional `new_framework_defaults_6_0.rb` and `new_framework_defaults_5_1.rb` files; set `config.load_defaults` to `6.0`.
* Updated one of the `healthcheck_spec.rb` tests to expect the charset to be included in Content-Type headers (as is the default on Rails 6.0).
* Added the `listen` gem to the development dependencies - without this, the Dockerised app wouldn't start correctly.

### Upgrading Ruby

* Switched to using a Ruby 2.7.1 interpreter.
* Updated the parent image in the Dockerfile to 2.7.1.

### Upgrading other dependencies

* Upgraded `carrierwave-mongoid` to 1.3.0.
* Upgraded `mongoid` to 7.1.1.
* Upgraded the AWS SDK to version 3 (and included `aws-sdk-s3` only, rather than the full set of services).
* Updated the tests that depended on Mongoid behaviour with regard to read-only properties: as of version 7, the gem no longer raises an error on read-only property update attempts.

### Upgrading CarrierWave

* Upgraded `carrierwave` to 2.1.0 (via 1.0.0).
* This required a production code change:
  * Before version 1.0.0, the `def remove_#{column}!` method (which is used as `asset.remove_file!`) would delete the physical file from disk, but not remove the `file` property from the Asset model.
  * As of version 1.0.0, `def remove_#{column}!` also removes the `file` property.
  * But the app tries to access `file` even after the physical file is deleted (because the asset has been successfully uploaded to S3).
 * Consequently, I have replaced `asset.remove_file!` with `FileUtils.remove_file(file.path)` in the two places it occurred.
* This also required a test update:
  * As of version 1.0.0, `asset.reload` does not refresh the file itself.
  * This is due to the issue addressed by this outstanding PR: https://github.com/carrierwaveuploader/carrierwave-mongoid/pull/190.
 * Consequently, in one test I have replaced `asset.reload` with `Asset.find(asset.id)`.

## Testing

* At every stage of the update, I have run the full unit test suite and ensured all tests pass (subject to the necessary changes enumerated above).
* I manually tested the S3 file upload journey (including the virus scan) to a real S3 bucket.
  * However, I did shortcircuit the async execution logic to make it synchronous, so as to remove the dependency on govuk_sidekiq.
* I tested that the Docker image builds and starts up successfully (and serves the default Rails root page).
* I have of course not tested integration with any other gov.uk services, or deployment to a server instance.


app/models/asset.rb
```diff
@@ -85,11 +85,15 @@ class Asset
 
     after_transition to: :uploaded do |asset, _|
       asset.save!
-      asset.remove_file!
```
catalin-ursachi, line 88: As of CarrierWave 1.0.0, calling `remove_file!` method would remove the `file` property from the Asset model; but other code, called later, depends on the `file` property being present even after the file is uploaded to S3.

So I've replaced this with `FileUtils.remove_file(file.path)`, which is functionally equivalent to the pre-upgrade logic.

I've also extracted these two lines to the `delete_file_from_nfs` method below, so that it can be reused in `delete_asset_file_from_nfs_worker.rb`.

config/initializers/new_framework_defaults_5_1.rb
```diff
@@ -1,14 +0,0 @@
-# Be sure to restart your server when you modify this file.
-#
-# This file contains migration options to ease your Rails 5.1 upgrade.
-#
-# Once upgraded flip defaults one by one to migrate to the new default.
```
catalin-ursachi, line 5: This file was, I believe, redundant, as we were already using the Rails 5.1 defaults, via the `config.load_defaults 5.1` statement in `application.rb`.

spec/models/asset_spec.rb
```diff
@@ -284,9 +284,11 @@
     end
 
     it "cannot be changed after creation" do
-      uuid = "11111111-1111-1111-1111-11111111111111"
-      asset = FactoryBot.create(:asset, uuid: uuid)
-      expect { asset.uuid = "22222222-2222-2222-2222-222222222222" }.to raise_error(Mongoid::Errors::ReadonlyAttribute)
```
catalin-ursachi, line 289: As of Mongoid 7, attempting to modify a read-only parameter raises an exception only if done via `update_attribute`, but not if done via the setter.

So `asset.update_attribute(:uuid, "222...")` would raise a `Mongoid::Errors::ReadonlyAttribute`, but `asset.uuid = "222..." does not.

The test could be easily modified to check for the error-raising behaviour, if that were functionally important here; but I'm assuming it is not.

spec/requests/healthcheck_spec.rb
```diff
@@ -5,7 +5,7 @@
     get "/healthcheck"
 
     expect(response.status).to eq(200)
-    expect(response.content_type).to eq("application/json")
+    expect(response.content_type).to eq("application/json; charset=utf-8")
```
catalin-ursachi, line 8: This behaviour is controlled by the: ```Rails.application.config.action_dispatch.return_only_media_type_on_content_type = false``` setting. I've assumed the default setting (including the charset) is preferable; if not, it would be easy to exclude it.

app/models/asset.rb
catalin-ursachi, line 88: As of CarrierWave 1.0.0, calling `remove_file!` method would remove the `file` property from the Asset model; but other code, called later, depends on the `file` property being present even after the file is uploaded to S3.

So I've replaced this with `FileUtils.remove_file(file.path)`, which is functionally equivalent to the pre-upgrade logic.

I've also extracted these two lines to the `delete_file_from_nfs` method below, so that it can be reused in `delete_asset_file_from_nfs_worker.rb`.

config/initializers/new_framework_defaults_5_1.rb
catalin-ursachi, line 5: This file was, I believe, redundant, as we were already using the Rails 5.1 defaults, via the `config.load_defaults 5.1` statement in `application.rb`.

spec/models/asset_spec.rb
catalin-ursachi, line 289: As of Mongoid 7, attempting to modify a read-only parameter raises an exception only if done via `update_attribute`, but not if done via the setter.

So `asset.update_attribute(:uuid, "222...")` would raise a `Mongoid::Errors::ReadonlyAttribute`, but `asset.uuid = "222..." does not.

The test could be easily modified to check for the error-raising behaviour, if that were functionally important here; but I'm assuming it is not.

spec/requests/healthcheck_spec.rb
catalin-ursachi, line 8: This behaviour is controlled by the: ```Rails.application.config.action_dispatch.return_only_media_type_on_content_type = false``` setting. I've assumed the default setting (including the charset) is preferable; if not, it would be easy to exclude it.

MuriloDalRi (approved):

PR #773, opened 2020-06-05
Provide better logs for "File can't be blank" errors

ChrisBAshton:
This PR:

* Fixes one broken test locally
* Adds a missing test for directory deletion
* Catches an edge case where the directory can't be deleted, and provides better logging information so we can figure out the root cause

Trello: https://trello.com/c/FDu35RHS/1977-5-investigate-whitehall-sending-blank-file-when-updating-assets

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -86,7 +86,16 @@ class Asset
     after_transition to: :uploaded do |asset, _|
       asset.save!
       asset.remove_file!
-      FileUtils.rmdir(File.dirname(asset.file.path))
+      dir = File.dirname(asset.file.path)
+      begin
+        FileUtils.rmdir(dir)
+      rescue Errno::ENOTEMPTY => e
+        GovukError.notify(e, extra: {
+          asset_id: asset.id,
+          filename: asset.filename,
+          other_files_in_dir: Dir["#{dir}/*"].map { |path| File.basename(path) },
+        })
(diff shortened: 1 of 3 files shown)


app/models/asset.rb
```diff
@@ -86,7 +86,13 @@ class Asset
     after_transition to: :uploaded do |asset, _|
       asset.save!
       asset.remove_file!
-      FileUtils.rmdir(File.dirname(asset.file.path))
+      dir = File.dirname(asset.file.path)
+      begin
+        FileUtils.rmdir(dir)
+      rescue Errno::ENOTEMPTY
+        files = Dir["#{dir}/*"].map { |path| path.split("/").last }
+        raise "Attempted to delete non-empty directory #{dir}. Contents: #{files}"
```
thomasleese, line 94: I think rather than raising a custom error, we should send the original exception to Notify but include the contents of the directory in the context.

See https://docs.sentry.io/clients/ruby/context/ for more detail (although `Raven.capture_event` is [replaced with `GovukError.notify`](https://github.com/alphagov/govuk_app_config/blob/56ae9cf05a1958fe41e0e8d53cb5ff918f82234b/lib/govuk_app_config/govuk_error.rb#L5))

ChrisBAshton: Nice idea! Have now implemented 👍

PR #784, opened 2020-07-07
Upgrade to rails 6

cdccollins:
Upgrade to rails 6

The versions of the gems carrierwave, carrierwave-mongoid, and mongoid were very out of date, and have been bumped up significantly. Most importantly, carrierwave-mongoid doesn't appear to be supported at the moment and so I had to add a fix directly in the code.


config/environments/development.rb
```diff
@@ -9,16 +9,27 @@
...
 
-  # Raises error for missing translations
+  # Raises error for missing translations.
   # config.action_view.raise_on_missing_translations = true
+
+  # Use an evented file watcher to asynchronously detect changes in source code,
+  # routes, locales, etc. This feature depends on the listen gem.
+  # config.file_watcher = ActiveSupport::EventedFileUpdateChecker
+
+  config.hosts << "asset-manager.dev.gov.uk"
```
pixeltrix, line 34: Unsure whether this should be `config.hosts.clear` or the following:

```
config.hosts = [
  "asset-manager.dev.gov.uk",
]
```

Trying to look through other repos for a pattern.

pudinha: As pointed in another comment, better to do with `+=` than `=`, so we don't override the addresses (like `127.0.0.1`) that are already on the array by default

pixeltrix: Yes, I changed my thinking 😄

spec/models/asset_spec.rb
```diff
@@ -1033,7 +1034,7 @@
       it "removes the underlying directory" do
         asset.upload_success!
 
-        expect(File.exist?(File.dirname(path))).to be_falsey
```
pudinha, line 1036: hmm did it stop working?

cdccollins: Yeah some new rubocop rules were introduced in this repo but somehow this one was missed. Andrew said just to put it in this PR too :)

Gemfile
```diff
@@ -1,20 +1,21 @@
 source "https://rubygems.org"
 
+gem "activemodel-serializers-xml"
```
pixeltrix, line 3: Where was the error being generated that caused you to add this? The class in question was removed from Rails 5 so seems odd to be popping up now.

cdccollins: I was getting this error `uninitialized constant ActiveModel::Serializers::Xml`, can't exactly remember where.

pixeltrix: Need to track this down - try removing the gem and seeing what the backtrace is.

cdccollins: Ok, running it now without that gem I don't seem to be getting the error again. From my commit history it seems it appeared when I ran `bundle update rails`. But now the error's gone I'll remove it

spec/models/asset_spec.rb
```diff
@@ -286,7 +286,8 @@
     it "cannot be changed after creation" do
       uuid = "11111111-1111-1111-1111-11111111111111"
       asset = FactoryBot.create(:asset, uuid: uuid)
-      expect { asset.uuid = "22222222-2222-2222-2222-222222222222" }.to raise_error(Mongoid::Errors::ReadonlyAttribute)
+      expect { asset.uuid = "22222222-2222-2222-2222-222222222222" }
+        .not_to change(asset, :uuid).from("11111111-1111-1111-1111-11111111111111")
```
pixeltrix, line 290: Seems a bit wasteful to type this out again when you have it in the variable `uuid` ;-)

cdccollins: d'oh 🤦‍♀️

spec/models/asset_spec.rb
```diff
@@ -286,7 +286,8 @@
     it "cannot be changed after creation" do
       uuid = "11111111-1111-1111-1111-11111111111111"
       asset = FactoryBot.create(:asset, uuid: uuid)
-      expect { asset.uuid = "22222222-2222-2222-2222-222222222222" }.to raise_error(Mongoid::Errors::ReadonlyAttribute)
+      expect { asset.uuid = "22222222-2222-2222-2222-222222222222" }
+        .not_to change(asset, :uuid).from(uuid)
```
pudinha, line 290: why specify the `.from(uuid)`? isn't just `not_to change(...)` enough?

Gemfile
```diff
@@ -1,20 +1,21 @@
 source "https://rubygems.org"
 
+gem "rails", "~> 6.0.3", ">= 6.0.3.2"
```
pudinha, line 3: fix version

EGiataganas:
Hey @cdccollins this looks great, I've compared the pr changes with [railsdiff][1] and there are some changes that didn't happen with `bundle exec rails app:update` that could also be done:

1. Remove `bin/bundle`, `bin/update`
2. Update [system! "bin/rails db:setup"][2] with `system! "bin/rails db:prepare"`
3. In the `config/environments/test.rb` the `config.cache_classes = true` has changed to `false` and there is `config.action_view.cache_template_loading = true`
4. In the `config/spring.rb` the `%w[]` needs to be updated to `Spring.watch()`
```ruby
Spring.watch(
  ".ruby-version",
  ".rbenv-vars",
  "tmp/restart.txt",
  "tmp/caching-dev.txt"
)
```
5. Remove `config/initializers/new_framework_defaults.rb` as per the reason in alphagov/travel-advice-publisher@8b0745e
6. Update `app/views/layouts/application.html.erb`to include `csp_meta_tag` after `csrf_meta_tags`. Maybe it will be good to have a `csrf_meta_tags`, `csp_meta_tag`, `stylesheet_link_tag` and then `javascript_include_tag`

[1]: http://railsdiff.org/5.2.4.3/6.0.3.2
[2]: https://github.com/alphagov/asset-manager/blob/master/bin/setup#L31

PR #796, opened 2020-07-27
Add whitehall-specific tasks to avoid SSH

benthorner:
https://trello.com/c/HfMApI04/947-make-it-easier-to-cleanup-stray-whitehall-assets

Please see the commits for more details.

### DIFF
--- a/lib/tasks/assets.rake
+++ b/lib/tasks/assets.rake
@@ -15,4 +15,10 @@ namespace :assets do
     asset.destroy
     Services.cloud_storage.delete(asset)
   end
+
+  desc "Mark a Whitehall asset as deleted and remove from S3"
+  task :whitehall_delete_and_remove_from_s3, [:legacy_url_path] => :environment do |_t, args|
+    asset = WhitehallAsset.find_by(legacy_url_path: args.fetch(:legacy_url_path))
+    Rake::Task["assets:delete_and_remove_from_s3"].invoke(asset.id)
+  end
 end
--- a/lib/tasks/assets.rake
+++ b/lib/tasks/assets.rake
@@ -1,14 +1,4 @@
 namespace :assets do
-  desc "Mark an asset as deleted"
-  task :delete, [:id] => :environment do |_t, args|
-    Asset.find(args.fetch(:id)).destroy
-  end
-
-  desc "Restore an asset after being deleted"
-  task :restore, [:id] => :environment do |_t, args|
-    Asset.find(args.fetch(:id)).restore
-  end
-
   desc "Mark an asset as deleted and remove from S3"
   task :delete_and_remove_from_s3, [:id] => :environment do |_t, args|
     asset = Asset.find(args.fetch(:id))
--- a/lib/tasks/assets.rake
+++ b/lib/tasks/assets.rake
@@ -1,14 +1,14 @@
 namespace :assets do
-  desc "Mark an asset as deleted and remove from S3"
-  task :delete_and_remove_from_s3, [:id] => :environment do |_t, args|
+  desc "Mark an asset as deleted and (optionally) remove from S3"
+  task :delete, %i[id permanent] => :environment do |_t, args|
     asset = Asset.find(args.fetch(:id))
     asset.destroy
-    Services.cloud_storage.delete(asset)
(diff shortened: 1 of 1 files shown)


benthorner:
- [ ] Deploy to Integration and check the new task works

lib/tasks/assets.rake
```diff
@@ -1,14 +1,4 @@
 namespace :assets do
-  desc "Mark an asset as deleted"
-  task :delete, [:id] => :environment do |_t, args|
-    Asset.find(args.fetch(:id)).destroy
-  end
-
-  desc "Restore an asset after being deleted"
-  task :restore, [:id] => :environment do |_t, args|
-    Asset.find(args.fetch(:id)).restore
-  end
```
ChrisBAshton, line 10: Do we ever use these? I can't see a justification for it in the commits where this was added, but I wonder if a 'soft delete' may be a desirable feature in some situations?

thomasleese: I believe we added soft-delete tasks because we wanted that to be the default unless otherwise specified. For example, if an asset needs to be removed because it's outdated and still showing up in Google, we would do a soft delete. If an asset needs to be removed because it affects national security, then it would be hard deleted.

benthorner: That's interesting. Can you help me understand the distinction better? I believe we only run these tasks when an asset has been orphaned. What's the risk of always doing a hard delete?

thomasleese: Well, the risk is that if we needed to restore an asset for whatever reason or get access to an asset that's not live on the site anymore. The normal behaviour of Asset Manager is not to delete the files from S3 when they get replaced or removed from documents (only [one search result for `cloud_storage.delete`](https://github.com/alphagov/asset-manager/search?q=cloud_storage.delete&unscoped_q=cloud_storage.delete)) so this matches that behaviour.

benthorner: Ah, OK. That's quite an obscure distinction in [the docs](https://docs.publishing.service.gov.uk/manual/howto-manually-remove-assets.html) - a choice is given, but without a way to make it. If we do support both kinds of behaviour, then we should indicate when to use them.

I've added another commit to make soft deletion the default. However, I don't think we need a task for restoring assets, since it's not a common or documented thing to do.

I can also update the docs with some guidance between the two tasks.

How does that sound?

lib/tasks/assets.rake
```diff
@@ -1,18 +1,14 @@
...
-
   desc "Mark an asset as deleted and remove from S3"
-  task :delete_and_remove_from_s3, [:id] => :environment do |_t, args|
+  task :delete, %i[id permanent] => :environment do |_t, args|
     asset = Asset.find(args.fetch(:id))
     asset.destroy
-    Services.cloud_storage.delete(asset)
+    Services.cloud_storage.delete(asset) if permanent
```
thomasleese, line 6:
```suggestion
    Services.cloud_storage.delete(asset) if args[:permanent]
```

Although I'm not sure if this works for booleans, in case someone passes in `"false"` which would evaluate to true...

lib/tasks/assets.rake
thomasleese: Sounds reasonable to me.

As a side point, maybe it's just me, but I would consider the Rake tasks themselves to be documentation. `rake -T` can be used to list the tasks, and that prints useful information like the description and the arguments. Following on from that, I would suggest improving the documentation of the Rake tasks themselves rather than having something in the dev docs which is far more likely to end up out of date with the specifics in the task itself.

lib/tasks/assets.rake
benthorner: Oops - fixed.

lib/tasks/assets.rake
benthorner: While I would normally agree, the problem is that these tasks are referred to from dev docs as part of the overall process of removing an asset. The links in the docs go straight to Jenkins, so there isn't an opportunity to run `rake -T`. Since we already have docs about removing an asset, which extend beyond asset manager, it makes sense to explain the distinction there, since that's probably where we will be looking to understand it.

PR #866, opened 2021-08-11
Fix dependencies to allow Dependabot to raise a Rails 6.1 PR

brucebolt:
Dependabot did not raise a Rails 6.1 update PR as there were incompatible dependencies.

Bumping `mongoid` to the latest version to remove the requirement for `activemodel < 6.1`.

Also adding `mongoid` to the dependabot configuration, so we don't get into this situation with the next Rails update.

[Trello card](https://trello.com/c/V56Gndqj)


.github/dependabot.yml
```diff
@@ -26,6 +26,8 @@ updates:
         dependency-type: direct
       - dependency-name: rspec-rails
         dependency-type: direct
+      - dependency-name: mongoid
```
ChrisBAshton, line 29: Could this be moved to between lines 22 and 23? So far, the list is alphabetical - let's keep it that way!

ChrisBAshton (approved):
LGTM 👍  one nitpick comment below.

.github/dependabot.yml
brucebolt: Good point.  I've moved this up the list.

brucebolt:
This has allowed dependabot to raise https://github.com/alphagov/asset-manager/pull/867.

PR #903, opened 2022-04-12
Add auth bypass id to whitehall endpoint

BeckaL:
Some foundational work to allow shareable previews for Whitehall to access assets - Whitehall has its own create endpoint, so this PR adds auth_bypass_id (a UUID that allows for non-sign-on access to an edition and its related content via a token) to the allowed parameters for Whitehall assets. This is identical to what is done for non-Whitehall assets.

This commit also introduces some tests for the WhitehallMediaController to ensure that an asset can be accessed with an access token containing an auth_bypass_id. This repeats some tests in the non-Whitehall media controller, and the behaviour is already defined for the WhitehallMediaController via inheritance, but I thought the tests were useful from a documentation standpoint.

### DIFF
--- a/spec/controllers/whitehall_media_controller_spec.rb
+++ b/spec/controllers/whitehall_media_controller_spec.rb
@@ -205,6 +205,84 @@ RSpec.describe WhitehallMediaController, type: :controller do
       end
     end
 
+    context "with draft uploaded asset with auth_bypass_ids" do
+      let(:user) { FactoryBot.build(:user) }
+      let(:auth_bypass_id) { "bypass-id" }
+      let(:asset) { FactoryBot.create(:uploaded_whitehall_asset, draft: true, auth_bypass_ids: [auth_bypass_id]) }
+      let(:valid_token) do
+        JWT.encode(
+          { "sub" => auth_bypass_id },
+          Rails.application.secrets.jwt_auth_secret,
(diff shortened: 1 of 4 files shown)


spec/controllers/whitehall_media_controller_spec.rb
```diff
@@ -205,6 +205,84 @@
...
+        end
+      end
+
+      context "when a user is not authenticated and has provided an invalid token" do
+        before { not_logged_in }
+
+        it "authenticates the user" do
+          cookies["auth_bypass_token"] = "bad-token"
+          expect(controller).to receive(:authenticate_user!)
+          get :download, params: { path: path, format: format }
```
brucebolt, line 272: Is there an expectation here (e.g. that the response is unsuccessful) that could be added?

brucebolt (commented):
Looks good to me 👍 .

There's just one test that I think is missing the assertion.

spec/controllers/whitehall_media_controller_spec.rb
BeckaL: Just writing this here for ref - as discussed, have added some requests specs for whitehall that cover this scenario.

brucebolt (approved):
Looks good to me 👍 .

Thanks for persevering with the tests!

PR #915, opened 2022-05-11
Stopping Freshclam from writing separate log file and adding app user

nsabri1:
There is no need for Freshclam daily updates to log to a separate log file as we already run it a s background process with logging to STDOUT.
`freshclam -d --stdout --foreground`

Also adding new user with ID 1001 and GID 1001


desenhar (approved):
theres a failing check I can't seem to view, besides that lgtm.

PR #920, opened 2022-06-06
Adding softlink to /tmp filesystem.

nsabri1:
This should allow writing to /app/tmp directory and fix the current permission denied errors
in EKS.

This is part of non-root container work:
https://trello.com/c/Jef1t4xU/903-fix-app-permission-errors-due-to-nonroot


fredericfran-gds (approved):
ok

PR #929, opened 2022-06-30
Add healthcheck for connectivity with S3 bucket

BeckaL:
This PR adds a healthcheck that checks for connectivity with the s3 bucket, as part of work to improve checks before enabling continuous deployment for the app.

It uses the head_bucket method from the aws s3 api, which returns a successful response if the bucket requested exists and the requester has access to it. In the case where we get this successful response, we return a GovukHealthcheck::OK response.

We return a GovukHealthcheck::CRITICAL status in the following cases:

* an error is raised by the call to head_bucket
* head_bucket returns an unsuccessful response

This is because there is some ambiguity in the [documentation](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html#head_bucket-instance_method) vs observed behaviour, so we want to protect against both cases.

[Trello card](https://trello.com/c/Sj4UTatV/126-enable-continuous-deployment-for-asset-manager)

### DIFF
--- a/spec/lib/s3_storage_spec.rb
+++ b/spec/lib/s3_storage_spec.rb
@@ -315,4 +315,33 @@ RSpec.describe S3Storage do
       end
     end
   end
+
+  describe "#healthy?" do
+    let(:head_bucket_params) { { bucket: bucket_name } }
+
+    let(:response) { instance_double("Seahorse::Client::Response") }
+
+    it "returns true when s3 bucket is reachable" do
+      allow(response).to receive(:successful?).and_return(true)
+      allow(s3_client).to receive(:head_bucket)
+                            .with(head_bucket_params).and_return(response)
+
+      expect(storage.healthy?).to eq true
+    end
+
+    it "returns false when struct returns returns false to successful?" do
+      allow(response).to receive(:successful?).and_return(false)
+      allow(s3_client).to receive(:head_bucket)
+                            .with(head_bucket_params).and_return(response)
+
+      expect(storage.healthy?).to eq false
+    end
+
+    it "returns false when calling head_bucket returns an error" do
+      allow(s3_client).to receive(:head_bucket)
+                            .with(head_bucket_params).and_raise(Aws::S3::Errors::NotFound.new(nil, nil))
+
+      expect(storage.healthy?).to eq false
+    end
+  end
 end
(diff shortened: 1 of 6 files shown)


spec/lib/s3_storage_spec.rb
```diff
@@ -315,4 +315,29 @@
...
+    let(:head_bucket_params) { { bucket: bucket_name } }
+
+    it "returns true when s3 bucket is reachable" do
+      allow(s3_client).to receive(:head_bucket)
+        .with(head_bucket_params).and_return(Aws::EmptyStructure.new)
+
+      expect(storage.healthy?).to eq true
+    end
+
+    it "returns false when s3 bucket is not found" do
```
kevindew, line 329:
```suggestion
    it "returns false when an AWS exception is raised" do
```

I think we should just have one test for AWS exceptions since we only catch one - it's ok if it's a specific one (like not found though).

If you take the suggestion above about using response.successful? we'd want a test for that - it looks a bit of a pain to create a Seahorse::Client::Response so may just want to set up a mock double.

spec/lib/s3_storage_spec.rb
```diff
@@ -315,4 +315,29 @@
       end
     end
   end
+
+  describe "#healthy?" do
+    let(:head_bucket_params) { { bucket: bucket_name } }
+
+    it "returns true when s3 bucket is reachable" do
+      allow(s3_client).to receive(:head_bucket)
+        .with(head_bucket_params).and_return(Aws::EmptyStructure.new)
```
kevindew, line 324: Bizarrely it looks like it returns a `Seahorse::Client::Response` object which - when represented with inspect shows "#<struct Aws::EmptyStructure>"

Evidence

```
irb(main):023:0> resp = client.head_bucket({ bucket: "govuk-assets-integration" })
[Aws::S3::Client 200 0.035634 0 retries] head_bucket(bucket:"govuk-assets-integration")

=> #<struct Aws::EmptyStructure>
irb(main):024:0> resp.class
=> Seahorse::Client::Response
irb(main):025:0> resp.to_s
=> "{}"
irb(main):026:0> resp.inspect
=> "#<struct Aws::EmptyStructure>"
```

kevindew (commented):
Looking great Becka

spec/lib/s3_storage_spec.rb
BeckaL: Oh thank you!! I was extremely confused as to how the EmptyStructure managed to get a successful? method on it

PR #938, opened 2022-08-05
Add clamdemon and sidekiq monitoring

fredericfran-gds:
1. Add clamdaemon, previously using it in standalone mode which was giving errors.
2. Add sidekiq monitoring, via update to govuk_app_config


Dockerfile
```diff
@@ -22,15 +22,12 @@ FROM $base_image
 ENV GOVUK_APP_NAME=asset-manager GOVUK_ASSET_ROOT=http://assets-origin.dev.gov.uk
 
 RUN apt update && \
-    apt install -y clamav shared-mime-info
+    apt install -y clamav clamav-daemon shared-mime-info
```
sengi, line 25: best practice: avoid including the apt database in the image and avoid installing optional "recommended" packages. (If there were "recommended" packages that were actually needed, it's best to specify them explicitly — there probably aren't, though.)
```suggestion
    apt install -y --no-install-recommends clamav clamav-daemon shared-mime-info && \
    rm -fr /var/lib/apt/lists/*
```

sengi (approved):
Need to update the title/description but otherwise LGTM 👍

Dockerfile
fredericfran-gds: thanks, I can fix that.

PR #944, opened 2022-08-12
Add a temporary task to update the legacy urls for topical event assets

BeckaL:
The ClassificationFeaturingImageData model was renamed in Whitehall, and this is used on the fly to generate the asset url, meaning that when a topical event was republished, its asset urls pointed to the new location (/topical_event_featuring_image_data) rather than the legacy url those assets are currently found at (/classification_featuring_image_data). This temporary rake task fixes this issue by updating all the legacy url paths for classsification_featuing_image_data to be topical_event_image_data. 

A long term fix for this would be to store the image urls in Whitehall, so they are not dynamically generated each time. But this is an immediate fix these document types for now.

The fix flow should be - run this task, and then republish all topical events immediately.

https://trello.com/c/3q6al3ol/207-fix-issue-with-topical-event-featured-document-image-urls 

### DIFF
--- a/lib/tasks/assets.rake
+++ b/lib/tasks/assets.rake
@@ -25,4 +25,21 @@ namespace :assets do
     asset = WhitehallAsset.find_by!(legacy_url_path: args.fetch(:legacy_url_path))
     Rake::Task["assets:redirect"].invoke(asset.id, args.fetch(:redirect_url))
   end
+
+  desc "Update a Whitehall asset legacy url path for a topical event featuring image"
+  task update_topical_event_legacy_url_paths: :environment do |_t|
+    legacy_urls_to_update = WhitehallAsset.where("legacy_url_path": /\/government\/uploads\/system\/uploads\/classification_featuring_image_data/).pluck(:legacy_url_path)
+
+    total = legacy_urls_to_update.size
+    puts "updating #{legacy_urls_to_update.size} assets"
+    legacy_urls_to_update.each_with_index do |legacy_url, index|
+      if (index % 1000).zero?
+        puts "#{index}/#{total} completed"
+      end
+      new_url = legacy_url.gsub(/\/classification_featuring_image_data/, "/topical_event_featuring_image_data")
+      WhitehallAsset
+        .where(legacy_url_path: legacy_url)
+        .update_all({ "$set" => { legacy_url_path: new_url } })
+    end
+  end
 end
--- a/spec/tasks/assets_tasks_spec.rb
+++ b/spec/tasks/assets_tasks_spec.rb
@@ -1,0 +1,41 @@
+require "rails_helper"
+
+RSpec.describe "asset tasks" do
+  describe "assets:update_topical_event_legacy_url_paths" do
+    let(:task) { Rake::Task["assets:update_topical_event_legacy_url_paths"] }
+
+    before { task.reenable }
+
+    def legacy_url(asset_name, model_name)
+      "/government/uploads/system/uploads/#{model_name}/file/1234/#{asset_name}.jpeg"
+    end
+
+    it "updates any assets with legacy urls matching /classification_featuring_image_data to /topical_event_image_data" do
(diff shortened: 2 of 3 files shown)


spec/tasks/assets_tasks_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  describe "update_topical_event_legacy_url_paths" do
+    let(:task) { Rake::Task["assets:update_topical_event_legacy_url_paths"] }
+
+    def legacy_url(asset_name, model_name)
+      "/government/uploads/system/uploads/#{model_name}/file/1234/#{asset_name}.jpeg"
+    end
+
+    def assert_assets_the_same_except_legacy_url_path(actual_assets, expected_assets)
+      ignored_keys = %w[legacy_url_path updated_at last_modified created_at] # annoyingly there's a really small difference in created at which is I think due to mongo truncating
```
BeckaL, line 14: Now I think about it, I am a bit worried about the difference in created at here, maybe there is an issue?

spec/tasks/assets_tasks_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+      task.invoke
+
+      ## There should now be 5 assets with an updated path, whose urls match and which are equivalent to the assets before they were renamed
+      topical_event_featuring_assets = WhitehallAsset.where(legacy_url_path: /\/government\/uploads\/system\/uploads\/topical_event_featuring_image_data/)
+      expected_urls = (1..5).map { |i| legacy_url("asset-#{i}", "topical_event_featuring_image_data") }
+      expect(topical_event_featuring_assets.pluck(:legacy_url_path)).to eq expected_urls
+      assert_assets_the_same_except_legacy_url_path(topical_event_featuring_assets.to_a, classification_featuring_assets)
+
+      ## We shouldn't have any assets left in the database with the classification featuring image data path
```
kevindew, line 33: I feel like having to write these comments is probably an indication that you're performing multiple tests in one `it` block. 

I imagine it may be distinct tests like:

```
it "replaces the legacy_url_path of topical event featurings" do
  # create x matching items
  task.invoke
  # expect count of old to be 0 
  # expect count of new to be x
end

it "changes only the legacy_url_path" do
# create one item
task.invoke
# compare attributes
end
```

I'm not sure you need one for that it doesn't change other items, I think that'd be covered implicitly in the first test

kevindew (commented):
This looks good, nice work Becka.

My comments are a bit more thorough that's what's needed for a temporary task, since we'd remove it after - however I thought it'd be good to share some rake task testing tips. Do feel free to be pragmatic in applying them.

I'm most interested in knowing what we might expect performance wise, do we have any idea? Just hoping it's not scary.

lib/tasks/assets.rake
```diff
@@ -25,4 +25,17 @@ namespace :assets do
...
+  task update_topical_event_legacy_url_paths: :environment do |_t|
+    legacy_urls_to_update = WhitehallAsset.where("legacy_url_path": /\/government\/uploads\/system\/uploads\/classification_featuring_image_data/).pluck(:legacy_url_path)
+
+    puts "updating #{legacy_urls_to_update.size} assets"
+    legacy_urls_to_update.each do |legacy_url|
+      new_url = legacy_url.gsub(/\/classification_featuring_image_data/, "/topical_event_featuring_image_data")
+      WhitehallAsset
+        .where(legacy_url_path: legacy_url)
+        .update_all({ "$set" => { legacy_url_path: new_url } })
+    end
```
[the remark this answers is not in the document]
BeckaL: Thanks for all your comments! All very useful. Good point. I've been a bit optimistic about this not being a super big task given only ~20,000 records to change, but I also know nothing about mongo + mongoid performance. Have added this logging in as you suggested, and will try it out on integration and see.

brucebolt (approved):
Looks good to me 👍.

PR #1005, opened 2022-10-21
Don't virus scan when redirect url is present

BeckaL:
If we receive a request [1] to create an asset, and then very quickly after that receive a request [2] to redirect the asset (i.e. the asset is changed in a publishing app, resulting in a new asset being created and the existing asset redirecting to the new asset), we can receive a `StateMachine:InvalidTransition` error.

This is because a concurrency issue to do with what happens when an asset is saved. We have an after save callback which a) schedules a virus scan if the asset is unscanned, which b) marks the asset as clean when this virus scan is complete and passes.

When we receive a create [1] and redirect [2] in short succession, the following happens:

1) we receive request [1] and schedule a virus check since the asset is unscanned
2) we receive request [2] and schedule a virus check since the asset is still unscanned
3) we complete the check from request [1] and mark the asset as clean
4) we complete the check from request [2] and try to mark the asset as clean

However, the process in step 4 will raise an error, since we are trying to mark a clean asset as clean again.

This commit fixes the issue by not scheduling virus scan checks for assets with redirects.
This works because we should have already scanned these assets, but also because we do not need to upload these assets to the S3 bucket since we will only serve redirects for them.

[Trello](https://trello.com/c/AwECBIdD/313-spike-into-fix-for-statemachinesinvalidtransition)

### DIFF
--- a/spec/requests/asset_requests_spec.rb
+++ b/spec/requests/asset_requests_spec.rb
@@ -147,4 +147,30 @@ RSpec.describe "Asset requests", type: :request do
       expect(response).to be_successful
     end
   end
+
+  describe "creating then redirecting an asset" do
+    it "does not result in an invalid transition error when a redirect is received in short succession after a create" do
+      # use threads to simulate multiple sidekiq workers
+      threads = []
+      allow(VirusScanWorker).to receive(:perform_async) do |asset_id|
+        threads << Thread.new do
+          sleep(0.5)
+          asset = Asset.find(asset_id)
+          asset.scanned_clean!
+        end
+      end
+
+      post "/assets", params: { asset: { file: load_fixture_file("lorem.txt") } }
+      asset_id = Asset.last.id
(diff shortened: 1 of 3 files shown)


spec/requests/asset_requests_spec.rb
```diff
@@ -147,4 +147,30 @@
...
+      allow(VirusScanWorker).to receive(:perform_async) do |asset_id|
+        threads << Thread.new do
+          sleep(0.5)
+          asset = Asset.find(asset_id)
+          asset.scanned_clean!
+        end
+      end
+
+      post "/assets", params: { asset: { file: load_fixture_file("lorem.txt") } }
+      asset_id = JSON.parse(response.body)["id"].split("/").last
```
brucebolt, line 164: I wonder whether we could use `Asset.last` rather than parsing the response?

brucebolt (approved):
Looks good to me 👍 .

I've made a non-blocking comment regarding a test, but feel free to ignore if it's not possible or reading the response is a better solution.

PR #1021, opened 2022-11-05
Minor improvements to examples in API docs.

sengi:
- Illustrate how to use a bearer token with `curl`. This is sometimes useful when troubleshooting.
- Add a paragraph break between the point about return format and the point about authentication.
- Remove unnecessary use of `echo` in commands for creating temporary files.


nsabri1 (approved):
OK

PR #1022, opened 2022-11-07
Fix clamdscan in container image. 

sengi:
- Install clamdscan package, which was missing.
- Patch the config files for clamd and freshclam to fix logging and permissions.
- Avoid installing shared-mime-info package twice, as it doesn't seem to be needed in the build stage.
- Fix a couple of Dockerfile lints.

Really we want ClamAV to run as a separate service so that we can sandbox it more restrictively and run it with the absolute minimum of privileges, but we'll tackle that in another PR.

This change only affects Kubernetes/Docker and not EC2 (i.e. current production).

Tested: ran locally in Docker:

```
$ docker build -t asset-manager
...
$ docker run -it -v clam-sigs:/var/lib/clamav asset-manager bash

app@242503f4a4c4:~$ clamd &
...
app@242503f4a4c4:~$ clamdscan Gemfile
/app/Gemfile: OK
app@242503f4a4c4:~$ cat >eicar.txt
...EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*
^D
app@242503f4a4c4:/tmp$ clamdscan eicar.txt 
Sat Nov  5 16:28:06 2022 -> /tmp/eicar.txt: Eicar-Signature(69630e4574ec6798239b091cda43dca0:69) FOUND
/tmp/eicar.txt: Eicar-Signature FOUND
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,8 +13,8 @@ FROM $base_image
 
 ENV GOVUK_APP_NAME=asset-manager
 # TODO: move ClamAV into a completely separate service.
-RUN apt update && \
-    apt install -y --no-install-recommends \
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
         clamav clamav-daemon clamdscan shared-mime-info && \
     rm -fr /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


nsabri1 (approved):
LGTM

PR #1023, opened 2022-11-07
Avoid writing to the image filesystem by default.

sengi:
Essentially the same change as https://github.com/alphagov/whitehall/commit/3d13ac8.

This only affects running in Docker / Kubernetes. No change in EC2 / current production.

Also some minor cleanup in `.dockerignore` while we're there.

Tested: [deployed to integration](https://github.com/alphagov/asset-manager/actions/runs/3410602597) and was able to [create an asset](https://github.com/alphagov/asset-manager/blob/main/docs/api.md#create-an-asset).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,8 +1,12 @@
+.dockerignore
 .git
 .gitignore
+Dockerfile
 Jenkinsfile
+Procfile
 README.md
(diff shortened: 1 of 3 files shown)


nsabri1 (approved):
OK

PR #1025, opened 2022-11-07
Stop patching `/etc/clamav/*.conf` in the container build.

sengi:
Turns out it's not really worth doing this, since nobody would want to run with this config anyway. We replace the whole of `/etc/clamav` via a configmap when running in k8s, and that's part of the Helm chart so there's no real portability issue here.

Patching this config in the Dockerfile just adds a potential source of confusion.

Instead let's just remove the default configs for ClamAV, since those won't work properly in a container.

Partially reverts 575363a.

Also avoid setting `GOVUK_UPLOADS_ROOT`, as it makes more sense to do that in the Helm chart, where we're configuring the volumeMount that it uses.

https://trello.com/c/moEiT5BG


nsabri1 (approved):
OK

PR #1193, opened 2023-09-05
Fix static URL for assets (local dev env)

JessyTW:
In local development using govuk-docker, we want to be able to preview Assets in Whitehall - this means rather than relying on Nginx we need to use a redirect to serve assets. The cloud-storage-proxy does not work in local env because of using fake s3, which serves the assets from the local file system.

We have also had to change the virus scanning tests to match the new behaviour. 

This PR is dependant of https://github.com/alphagov/govuk-docker/pull/685. Please make sure to update govuk-docker.

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -97,6 +97,9 @@ protected
 
     if request.fresh?(response)
       head :not_modified
+    elsif AssetManager.s3.fake?
+      url = Services.cloud_storage.presigned_url_for(asset, http_method: request.request_method)
+      redirect_to url
     else
       url = Services.cloud_storage.presigned_url_for(asset, http_method: request.request_method)
       headers["X-Accel-Redirect"] = "/cloud-storage-proxy/#{url}"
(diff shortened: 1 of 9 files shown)


spec/controllers/media_controller_spec.rb
```diff
@@ -38,142 +39,167 @@ def download
         allow(asset).to receive(:last_modified).and_return(last_modified)
         allow(AssetManager).to receive(:content_disposition).and_return(content_disposition)
         allow(content_disposition).to receive(:header_for).with(asset).and_return("content-disposition")
+        allow(AssetManager).to receive(:s3).and_return(s3)
       end
 
-      shared_examples "a download response" do
```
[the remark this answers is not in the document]
JonathanHallam: dw, doesn't look bad at all when you use the whitespace filter :)

app/controllers/media_controller.rb
```diff
@@ -97,6 +97,10 @@ def proxy_to_s3_via_nginx(asset)
 
     if request.fresh?(response)
       head :not_modified
+    elsif AssetManager.s3.fake?
+      url = Services.cloud_storage.presigned_url_for(asset, http_method: request.request_method)
+      redirect_to url
+      nil
```
JonathanHallam, line 103: what's this nil return value for? I'm assuming this would never get hit if it's redirecting?

JessyTW: Thanks for pointing that out. After reviewing the other usages for redirect url, we noticed that it wasn't necessary to return nil. We have now removed it.

JonathanHallam: Do you mind pushing that change up? The `nil` is still there

JessyTW: The change has been pushed now, thank you!!

JonathanHallam (approved):
LGTM 👍

PR #1218, opened 2023-10-13
Bump `sidekiq` to 6.5.12

brucebolt:
This resolves the security alert in https://github.com/alphagov/transition/security/dependabot/37.

Dependabot was unable to raise a PR to bump this gem, so I've done it manually.


brucebolt:
We cannot upgrade beyond `sidekiq` version `6.5.5` as [`6.5.6`](https://github.com/sidekiq/sidekiq/blob/main/Changes.md#656) requires `redis <5` and we are already using version 5.

This will be resolved when we upgrade to Sidekiq 7.

PR #1241, opened 2023-11-22
Add logging to help diagnose ongoing issue with virus scanning

jkempster34:
https://trello.com/c/9YZ5VItp

We have an ongoing issue where we are seeing the following errors related to missing files coming from the VirusScanWorker:
- Errno::ENOENT: No such file or directory
- StateMachines::InvalidTransition: Cannot transition state via :scanned_clean from :unscanned (Reason(s): File can't be blank)

Assets should go through the following stages:
1. unscanned => saved to the file system and scanned by ClamAV in a background job.
2. scanned  => Saved to AWS.
3. uploaded => removed from the file system.

However, it seems like the file has already been removed during the virus scanning stage.

This adds logging so that we can trace the history of assets which raise the above errors to try to diagnose the issue.

---

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -85,6 +85,11 @@ class Asset
   scope :undeleted, -> { where(deleted_at: nil) }
 
   state_machine :state, initial: :unscanned do
+    around_transition do |asset, transition, block|
+      Rails.logger.info("#{asset.id} - Asset#state_machine - event: #{transition.event}")
+      block.call
+    end
+
     event :scanned_clean do
       transition unscanned: :clean
     end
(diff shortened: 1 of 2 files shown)


app/models/asset.rb
```diff
@@ -85,6 +85,11 @@ class Asset
   scope :undeleted, -> { where(deleted_at: nil) }
 
   state_machine :state, initial: :unscanned do
+    around_transition do |asset, transition, block|
+      Rails.logger.info("#{asset.id} - Asset#state_machine - event: #{transition.event}")
```
mike29736, line 89: I'm not familiar with this state machine, so I'll assume that all Assets have an `#id` for all transitions (i.e. there's no point in using safe navigation here)

mike29736: (And I see that `Transition#event` comes from https://github.com/state-machines/state_machines/blob/v0.6.0/lib/state_machines/transition.rb#L65, so that's nothing to worry about)

mike29736 (approved):
As long as it's true about `Asset#id`, it looks good to me :+1:

PR #1248, opened 2023-11-27
Updating ClamAV engine with offical version 1.2.1

nsabri1:
Currently adding latest version found on https://www.clamav.net/downloads
 
 As a todo:
 Can improve this by using curl to ascertain the latest release then
 injecting that as variable `$LATEST` rather then hardcoding the current
 version.
 
 Thus the wget line will change to:
` wget https://www.clamav.net/downloads/production/clamav-$LATEST.linux.x86_64.deb`
 
https://trello.com/c/TicAq9G1/1238-update-clamav-engine-to-latest-version

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 ARG ruby_version=3.2.2
 ARG base_image=ghcr.io/alphagov/govuk-ruby-base:$ruby_version
 ARG builder_image=ghcr.io/alphagov/govuk-ruby-builder:$ruby_version
-
+ARG clam_engine=clamav-1.2.1.linux.x86_64.deb
 
 FROM $builder_image AS builder
 
@@ -14,13 +14,15 @@ RUN bootsnap precompile --gemfile .
 
 FROM $base_image
 
+ARG clam_engine
 ENV GOVUK_APP_NAME=asset-manager
 
 # TODO: move ClamAV into a completely separate service.
-RUN install_packages clamav clamav-daemon clamdscan shared-mime-info && \
-    rm -fr /etc/clamav/* && \
-    mkdir -p /var/run/clamav && \
-    chown app:app /var/run/clamav /var/lib/clamav
+RUN install_packages wget shared-mime-info && \
+    wget https://www.clamav.net/downloads/production/$clam_engine && \
+    apt install ./$clam_engine && rm ./$clam_engine && \
+    mkdir -p /var/run/clamav /var/lib/clamav /usr/local/share/clamav && \
+    chown app:app /var/run/clamav /var/lib/clamav /usr/local/share/clamav
 
 WORKDIR $APP_HOME
 COPY --from=builder $BUNDLE_PATH $BUNDLE_PATH


Dockerfile
```diff
@@ -17,9 +17,10 @@ FROM $base_image
 ENV GOVUK_APP_NAME=asset-manager
 
 # TODO: move ClamAV into a completely separate service.
-RUN install_packages clamav clamav-daemon clamdscan shared-mime-info && \
-    rm -fr /etc/clamav/* && \
-    mkdir -p /var/run/clamav && \
+RUN install_packages wget shared-mime-info && \
+    wget https://www.clamav.net/downloads/production/clamav-1.2.1.linux.x86_64.deb && \
+    apt install ./clamav-1.2.1.linux.x86_64.deb && \
```
sengi, line 22: Don't forget to `rm` the package file after installing it, so it doesn't increase the image size unnecessarily.

Dockerfile
```diff
@@ -17,9 +17,10 @@ FROM $base_image
 ENV GOVUK_APP_NAME=asset-manager
 
 # TODO: move ClamAV into a completely separate service.
-RUN install_packages clamav clamav-daemon clamdscan shared-mime-info && \
-    rm -fr /etc/clamav/* && \
-    mkdir -p /var/run/clamav && \
+RUN install_packages wget shared-mime-info && \
+    wget https://www.clamav.net/downloads/production/clamav-1.2.1.linux.x86_64.deb && \
```
sengi, line 21: I'd recommend making the filename a variable (`ARG`) so that you can update in one place.

sengi, line 21: Looks like you're hardcoding amd64 here — this means developers will no longer be able to build natively on arm64 laptops (which they pretty much all have these days). Can you make it work for both architectures?

If not, that's ok for now (not ideal but ok) but you'd need to just check this cross-builds an amd64 image ok from an M1 Mac. (If that's already how you're working then you've already got it covered!)

sengi (approved):

Dockerfile
sengi: Just had a quick look myself and disappointingly Cisco doesn't seem to release arm64 packages itself :/ Their Docker images sometimes (but weirdly inconsistently) have arm64 e.g. [1.2.0](https://hub.docker.com/layers/clamav/clamav-debian/1.2.0-6/images/sha256-54a7e4404aff642d6fd611f8f2f73bf6c198e1dfcc0e533d29c226622f50407d?context=explore). That might be usable (`COPY --from`) but it's a different approach. Probably don't worry about it for now I guess.

nsabri1: Good point here - i could probably download the source https://www.clamav.net/downloads/production/clamav-1.2.1.tar.gz - but building it make introduce complexity 
but i feel they will be able to build the dockerfile on M processors as i was able to do so.

sengi: Oof yeah agree not worth trying to build it ourselves, that'd be a real extra maintenance burden that we could do without.

PR #1254, opened 2023-11-29
Resolve `Rails.application.secrets` deprecation warning

brucebolt:
In Rails 7.2, `Rails.application.secrets` is being deprecated.

We can solve this issue by using the `config_for` method, which will rerieve the relevant secret from the `config/secrets.yml` file.

[Trello card](https://trello.com/c/F8tMzJUp)


jkempster34 (approved):
Happy to approve this, though it might be worth linking the other solution and saying why it was unsuitable.

Another option might be drop the following:

```ruby
unless Rails.application.secrets.jwt_auth_secret
    raise "JWT auth secret is not configured. See config/secrets.yml"
end
```

Then, use the bang method of accessing the JWT credential. That should raise an error according to this https://www.codingzeal.com/post/managing-secrets-credentials-yml-vs-environment-variables-in-ruby-on-rails#:~:text=Handling%20Missing%20Configuration

brucebolt:
> Another option might be drop the following:
> 
> ```ruby
> unless Rails.application.secrets.jwt_auth_secret
>     raise "JWT auth secret is not configured. See config/secrets.yml"
> end
> ```
> 
> Then, use the bang method of accessing the JWT credential. That should raise an error according to this https://www.codingzeal.com/post/managing-secrets-credentials-yml-vs-environment-variables-in-ruby-on-rails#:~:text=Handling%20Missing%20Configuration

I'm assuming the purpose of checking for the JWT secret in `application.rb` is to catch the absence of the value before the application loads, rather than waiting until runtime.

jkempster34:
> > Another option might be drop the following:
> > ```ruby
> > unless Rails.application.secrets.jwt_auth_secret
> >     raise "JWT auth secret is not configured. See config/secrets.yml"
> > end
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then, use the bang method of accessing the JWT credential. That should raise an error according to this https://www.codingzeal.com/post/managing-secrets-credentials-yml-vs-environment-variables-in-ruby-on-rails#:~:text=Handling%20Missing%20Configuration
> 
> I'm assuming the purpose of checking for the JWT secret in `application.rb` is to catch the absence of the value before the application loads, rather than waiting until runtime.

Ah, yes. That makes sense!

PR #1263, opened 2023-12-07
Constrain `VirusScanWorker` to one concurrent worker per asset

brucebolt:
This will prevent multiple jobs becoming enqueued for a single asset until the original job has finished executing.

We are currently seeing an issue where a user can replace the file after uploading but before the virus scan has completed. This results in a second virus scan being enqueued, which subsequently fails as the asset has been deleted from the filesystem. This should prevent that from happening.

[Trello card](https://trello.com/c/UWS9qEtj)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,6 +32,7 @@ group :development, :test do
   gem "pact", require: false
   gem "pact_broker-client", require: false
   gem "rspec-rails"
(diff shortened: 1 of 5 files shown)


aldavidson (approved):
LGTM - nice solution!

PR #1266, opened 2023-12-11
Remove Whitehall Asset

yuetylauiris:
The whitehall_assets endpoints are no longer used by `whitehall` app and are moved to assets endpoints like others, so we removing the routes and controllers form asset manager.

Whitehall download endpoint are kept as there are still valid use cases 

This PR also have a dependency on [this gds-api-adapters PR](https://github.com/alphagov/gds-api-adapters/pull/1220)

[Trello card](https://trello.com/c/Kp9nquOr/71-task-remove-whitehallasset-concept-from-asset-manager-and-gds-api-adapters)


JonathanHallam (approved):
nice! thanks

PR #1311, opened 2024-02-13
Added some documentation of the various authorisation options available for assets

ryanb-gds:
I've been attempting to rationalise the handling of attachment assets in Whitehall. To do that I needed to understand what some of the parameters Whitehall includes in the requests that it sends to Asset Manager do. I found that there wasn't any documentation describing those parameters, so I thought I would add some. Let me know if there are any inaccuracies.

I did think about including this documentation within the API document, but the way that is formatted at the moment doesn't lend itself brilliantly to describing individual keys within a request body, so I went for a separate doc in the end.

### DIFF
--- a/docs/authorisation.md
+++ b/docs/authorisation.md
@@ -1,0 +1,36 @@
+# Authorisation
+
+In some cases, assets should not be publicly accessible. This only applies when they are in the "draft" state, which typically
+means that they are associated with a content item which is present on the "draft" stack in content store, but not present on the "live" stack.
+All assets associated with "live" content must be publicly available.
+
+There are three types of authorisation that can be applied to assets in the "draft" state:
+
+- Authorisation based on the user ID
+- Authorisation based on the user's organisation
+- Authorisation based on a bypass token
+
+To apply any of the above authorisation protocols, the `draft` key must have a value of `true` in the request body sent to the
+create or update asset API endpoints.
+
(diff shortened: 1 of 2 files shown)


docs/authorisation.md
```diff
@@ -0,0 +1,36 @@
...
+- Authorisation based on the user's organisation
+- Authorisation based on a bypass token
+
+To apply any of the above authorisation protocols, the `draft` key must have a value of `true` in the request body sent to the
+create or update asset API endpoints.
+
+## Authorisation based on the user ID
+
+To provide an allowlist of users that should be able to access an asset, include the `access_limited` key in the request body
+when creating or updating an asset. The value should be an array of Signon user IDs. An empty array means "no restrictions".
```
brucebolt, line 19:
```suggestion
when creating or updating an asset. The value should be an array of Signon user IDs. An empty array means "no restrictions of this type".
```

brucebolt (approved):
Looks good to me, thanks for writing this.

I've just made a comment (which applies to all three cases) just to clarify that an empty array could still have other types of authorisation in place.

ryanb-gds:
> Looks good to me, thanks for writing this.
> 
> I've just made a comment (which applies to all three cases) just to clarify that an empty array could still have other types of authorisation in place.

Good shout, thanks Bruce!

PR #1400, opened 2024-06-14
Build and Bundle ClamAV to get Native Containers

dj-maisy:
## What?
This adds another build stage for the Docker Build process so we can bundle a natively compiled version of ClamAV for the respective architecture and version of Ubuntu that our Ruby apps are built on.

## Why?
For some odd reason, Cisco builds separate ARM images of ClamAV but does not provide up to date ARM-supported packages via APTitude. This way, we can build and bundle a version of ClamAV that we know is going to work with our app, regardless of which platform we are trying to launch it on.

The original plan was to spin ClamAV into a "sidecar" container but this still would have required a local copy of `clamdscan` and so this approach would not have worked-around or solved the problem of suitable pre-compiled binaries not being available.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
-ARG clamav_binary_src=https://www.clamav.net/downloads/production/clamav-1.3.1.tar.gz
+ARG clamav_version=1.3.1
 ARG ruby_version=3.3
 ARG base_image=ghcr.io/alphagov/govuk-ruby-base:$ruby_version
 ARG builder_image=ghcr.io/alphagov/govuk-ruby-builder:$ruby_version
 
 FROM --platform=$TARGETPLATFORM index.docker.io/library/ubuntu:22.04 AS clam_builder
 
-ARG clamav_binary_src
+ARG clamav_version
 
 WORKDIR /src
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,33 +1,140 @@
+ARG clamav_version=1.3.1
 ARG ruby_version=3.3
 ARG base_image=ghcr.io/alphagov/govuk-ruby-base:$ruby_version
 ARG builder_image=ghcr.io/alphagov/govuk-ruby-builder:$ruby_version
-ARG clam_engine=clamav-1.2.1.linux.x86_64.deb
 
-FROM --platform=$TARGETPLATFORM $builder_image AS builder
+FROM --platform=$TARGETPLATFORM index.docker.io/library/ubuntu:22.04 AS clam_builder
```
theseanything, line 6:
```suggestion
FROM --platform=$TARGETPLATFORM public.ecr.aws/lts/ubuntu:22.04_stable AS clam_builder
```

Maybe keep the base the same as our ruby images?
https://github.com/alphagov/govuk-ruby-images/blob/e0e9bd1aaf52ba4d5d5c7f82c10662922d9a1d41/base.Dockerfile#L1

theseanything (approved):

PR #1402, opened 2024-06-17
Netcat package has been renamed in Ubuntu 24

dj-maisy:
## What?
Switching to Ubuntu 24 to align with the base images. Also, the `netcat` package needs to be switched to `netcat-traditional` in this version of Ubuntu.


Dockerfile
```diff
@@ -125,7 +125,7 @@ RUN apt update && apt-get install -y libbz2-1.0 \
         libxml2 \
         zlib1g \
         tzdata \
-        netcat && \
+        netcat-traditional && \
```
sengi, line 128: I think the direct replacement for what we had before is `netcat-openbsd` but it probably doesn't matter much 🤷

sengi (approved):

PR #1408, opened 2024-06-19
Symlink the new AV path to avoid breakages.

dj-maisy:
## What?
The new build process leaves the various ClamAV binaries in `/usr/bin`, whereas they were previously linked to `/usr/local/bin`. Let's symlink these binaries so they can still be found in the right place by Asset Manager.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -133,6 +133,16 @@ RUN apt update && apt-get install -y libbz2-1.0 \
 WORKDIR $APP_HOME
 
 COPY --from=clam_builder "/clamav" "/"
+
+RUN ln -s /usr/bin/clamav-config /usr/local/bin/clamav-config && \
+    ln -s /usr/bin/clambc /usr/local/bin/clambc && \
+    ln -s /usr/bin/clamconf /usr/local/bin/clamconf && \
+    ln -s /usr/bin/clamdscan /usr/local/bin/clamdscan && \
+    ln -s /usr/bin/clamdtop /usr/local/bin/clamdtop && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -133,6 +133,16 @@ RUN apt update && apt-get install -y libbz2-1.0 \
...
 
 COPY --from=clam_builder "/clamav" "/"
+
+RUN ln -s /usr/bin/clamav-config /usr/local/bin/clamav-config && \
+    ln -s /usr/bin/clambc /usr/local/bin/clambc && \
+    ln -s /usr/bin/clamconf /usr/local/bin/clamconf && \
+    ln -s /usr/bin/clamdscan /usr/local/bin/clamdscan && \
+    ln -s /usr/bin/clamdtop /usr/local/bin/clamdtop && \
+    ln -s /usr/bin/clamscan /usr/local/bin/clamscan && \
+    ln -s /usr/bin/clamsubmit /usr/local/bin/clamsubmit && \
```
sengi, line 143: How about:

```suggestion
RUN ln -s /usr/bin/clam* /usr/local/bin \
```

Dockerfile
```diff
@@ -133,6 +133,16 @@ RUN apt update && apt-get install -y libbz2-1.0 \
...
+    chmod 0755 /usr/local/bin/clam*
```
sengi, line 144: This changes the mode of the link targets rather than the links themselves. Is that what you intend here? (and have you checked that it's necessary?)

sengi (approved):

sengi:
It'd also be cool to fix the app not to specify the full path so that we can remove this workaround.

PR #1411, opened 2024-06-20
Build ClamAV to run as "app" user for container.

dj-maisy:
## What?
This changes the clam configuration files to try and run the Clam services as the "app" user in the container to try and avoid any permissions problems being experienced. This may explain why we are getting permissions issues when freshclam is trying to run in integration.

Also set the permissions for the symlinks in /usr/local/bin to match the permissions elsewhere (the ones we created seem to default to 0777).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -135,8 +135,7 @@ WORKDIR $APP_HOME
 COPY --from=clam_builder "/clamav" "/"
 
 RUN ln -s /usr/bin/clam* /usr/local/bin && \
-    ln -s /usr/bin/freshclam /usr/local/bin && \
-    chown 0755 /usr/local/bin/clam* /usr/local/bin/freshclam
+    ln -s /usr/bin/freshclam /usr/local/bin
 
 COPY --from=app_builder $BUNDLE_PATH $BUNDLE_PATH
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -134,7 +134,9 @@ WORKDIR $APP_HOME
 
 COPY --from=clam_builder "/clamav" "/"
 
-RUN ln -s /usr/bin/clam* /usr/local/bin
+RUN ln -s /usr/bin/clam* /usr/local/bin && \
+    ln -s /usr/bin/freshclam /usr/local/bin && \
+    chown 0755 /usr/local/bin/clam* /usr/local/bin/freshclam
```
sengi, line 139: Did you mean chmod? Not sure what's intended there though because the mode of the binaries should already be fine (like 0755 or something) and the mode of the links themselves is also gonna be fine by default. If the binaries being symlinked to are 0700 or something then `chmod 755` on them would make sense (but preferably not via the links, because that'd be a little confusing to someone reading the script).

sengi (approved):

Dockerfile
dj-maisy: Whoops yes, wrong command name. 🤦

PR #1415, opened 2024-06-21
Set correct config directory when building clamav

samsimpson1:
The config path appears to be embedded in the clamav binaries, and it wasn't pointing to where we're putting our config files. This seems to fix it (tested uploading a PDF to whitehall in integration)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,7 +57,7 @@ RUN apt update && apt install -y \
           -DCMAKE_BUILD_TYPE="Release" \
           -DCMAKE_INSTALL_PREFIX="/usr" \
           -DCMAKE_INSTALL_LIBDIR="/usr/lib" \
-          -DAPP_CONFIG_DIRECTORY="/etc/clamav" \
+          -DAPP_CONFIG_DIRECTORY="/usr/local/etc" \
           -DDATABASE_DIRECTORY="/var/lib/clamav" \
           -DENABLE_CLAMONACC=OFF \
           -DENABLE_EXAMPLES=OFF \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -72,32 +72,6 @@ RUN apt update && apt install -y \
...
-        -e "s|^\#\(ScriptedUpdates\).*|\1 yes|" \
-        "/clamav/etc/clamav/freshclam.conf.sample" > "/clamav/etc/clamav/freshclam.conf" && \
-    sed -e "s|^\(Example\)|\# \1|" \
-        -e "s|.*\(PidFile\) .*|\1 /tmp/clamav-milter.pid|" \
-        -e "s|.*\(MilterSocket\) .*|\1 inet:7357|" \
-        -e "s|.*\(User\) .*|\1 clamav|" \
-        -e "s|^\#\(LogFile\) .*|\1 /dev/stdout|" \
-        -e "s|^\#\(LogTime\).*|\1 yes|" \
-        -e "s|.*\(\ClamdSocket\) .*|\1 unix:/tmp/clamd.sock|" \
-        "/clamav/etc/clamav/clamav-milter.conf.sample" > "/clamav/etc/clamav/clamav-milter.conf" || \
```
theseanything, line 98: Where are we specifying this config instead?

theseanything: (as in where in our codebase)

samsimpson1: This configmap is being mounted at `/usr/local/etc`: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/asset-manager/templates/clamav-configmap.yaml

theseanything (approved):

PR #1439, opened 2024-07-12
Add `assets:get_id_by_legacy_url_path` rake task

ChrisBAshton:
It seems unnecessarily laborious to have to run rails console commands (https://docs.publishing.service.gov.uk/manual/manage-assets.html#get-an-assets-id) in order to know what to pass to a rake task
(https://docs.publishing.service.gov.uk/manual/manage-assets.html#redirect-an-asset).

It should just be a rake task.

### DIFF
--- a/spec/lib/tasks/assets_spec.rb
+++ b/spec/lib/tasks/assets_spec.rb
@@ -1,0 +1,32 @@
+require "rails_helper"
+require "rake"
+
+RSpec.describe "assets.rake" do
+  before do
+    AssetManager::Application.load_tasks if Rake::Task.tasks.empty?
+  end
+
+  describe "get_id_by_legacy_url_path" do
+    before do
+      task.reenable # without this, calling `invoke` does nothing after first test
+    end
+
+    let(:task) { Rake::Task["assets:get_id_by_legacy_url_path"] }
+
+    it "returns ID of asset by its legacy URL path" do
+      id = "abc123def456ghi789"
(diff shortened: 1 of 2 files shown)


spec/lib/tasks/assets_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+RSpec.describe "assets.rake" do
+  before do
+    AssetManager::Application.load_tasks if Rake::Task.tasks.empty?
+  end
+
+  describe "get_id_by_legacy_url_path" do
+    before do
+      task.reenable # without this, calling `invoke` does nothing after first test
+      allow($stdout).to receive(:puts) # supress Rake task output for cleaner tests
```
unoduetre, line 12: I don't see any difference with and without this, as you're capturing the output.

ChrisBAshton: It suppresses the output in the console (so we don't see lots of rake task output when running `rspec`). But we still want to have tests checking what the output will contain when we run it for real.

unoduetre: Have you tried to run it with and without this line? What was the difference?

ChrisBAshton: Huh, you're right - it made no difference! I copied it over from https://github.com/alphagov/manuals-publisher/pull/2427 (where it _did_ make a difference) - not sure why it's not needed here, but have now removed it.

unoduetre (approved):
Approved

PR #1457, opened 2024-08-14
Upgrade to Rails 7.2

brucebolt:
Note: We can’t yet upgrade the apps using mongo, since mongoid isn’t compatible with Rails 7.2. See [issue here](https://jira.mongodb.org/browse/MONGOID-5806).

[Trello card](https://trello.com/c/KtQiL6h1)


brucebolt:
Closing as the current version of mongoid isn't compatible with Rails 7.2.  We will wait until dependabot raises a PR.

PR #1475, opened 2024-09-06
Validate public assets have public document URLs

ryanb-gds:
There are over 1000 assets in the Asset Manager database which are not drafts, but have a parent document URL that points to the draft stack. Whilst this doesn't seem to break anything, that shouldn't really be possible so it would seem to make sense to apply some validation.

We expect this may help us to diagnose an issue where at least some of those 1000+ assets were unexpectedly made public ahead of their parent document being published. If this validation had been in place we think it would have prevented the issue.

I'm not really sure that we should be keeping parent document URLs in Asset Manager as it seems like a rather circular relationship, but anyway, it might prove useful in this case!

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -280,5 +280,9 @@ protected
     unless uri && %w[http https].include?(uri.scheme)
       errors.add(:parent_document_url, "must be an http(s) URL")
     end
+
+    if uri && uri.host.start_with?("draft-origin") && !draft?
+      errors.add(:parent_document_url, "must be a public GOV.UK URL")
+    end
   end
 end
--- a/spec/controllers/assets_controller_spec.rb
+++ b/spec/controllers/assets_controller_spec.rb
@@ -68,10 +68,10 @@ RSpec.describe AssetsController, type: :controller do
       end
 
       it "stores parent_document_url on asset" do
-        attributes = valid_attributes.merge(parent_document_url: "parent-document-url")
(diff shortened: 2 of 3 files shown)


app/models/asset.rb
```diff
@@ -280,5 +280,9 @@ def ensure_parent_document_url_is_valid
     unless uri && %w[http https].include?(uri.scheme)
       errors.add(:parent_document_url, "must be an http(s) URL")
     end
+
+    if uri && uri.host.start_with?("draft-origin") && !draft?
```
ryanb-gds, line 284: Can I find the draft origin host using Plek? There didn't seem to be an obvious way to do it but I suspect it might be possible, and would be better than hard coding it here

JonathanHallam: I honestly don't know. I can see that the govuk-config file uses plek in [a similar way](https://github.com/alphagov/asset-manager/blob/main/lib/govuk_configuration.rb#L15-L23) to what you're describing but not similar enough to actually use here

ryanb-gds: I have had a play around on integration and I can't seem to persuade it to. There are a bunch of PLEK related env vars in the integration container but I'm not sure how they got there, given I can't see them in the helm charts or AWS secrets manager. Are you able to give this a thumbs up for now @JonathanHallam, unless you've got other objections?

JonathanHallam: yep all good :)

JonathanHallam (approved):
lgtm

PR #1493, opened 2024-09-23
Upgrade to Sidekiq 7

brucebolt:
This upgrades `govuk_sidekiq` to 9.0.0, which in turn upgrades Sidekiq to 7.*.

After deployment, the `redis_namespace:remove_namespace` rake task will need to be immediately run, to retain any jobs that were queued before the upgrade was deployed.

[Trello card](https://trello.com/c/K9ds4XZX)


Gweaton (approved):
Thanks for the descriptive commits, definitely helped with my review.

PR #1568, opened 2024-12-11
Add documentation - overview of asset fields

lauraghiorghisor-tw:
Add summary of asset fields. This captures what we've learned during the asset deletion investigation.

Opened on the back of [this other PR](https://github.com/alphagov/govuk-developer-docs/pull/4866).

[Trello](https://trello.com/c/YIs8YZmF/3146-fix-live-assets-on-superseded-editions)

### DIFF
--- a/docs/asset_state_machine_overview.md
+++ b/docs/asset_state_machine_overview.md
@@ -1,0 +1,43 @@
+## Asset state machine overview
+
+Asset Manager stores a a number of fields whose values work together to represent the state of an asset.
+These fields are closely related to various document states from the publishing applications.
+
+1. `draft`
+
+Assets are set to draft (boolean `true`) when they first get created on a draft edition. When the document gets published, the draft state of the asset should be updated to `false` by the publishing application.
+All assets associated with live content must be publicly available. Once `draft` is set to `false`, it should not be set back to `true`. There are other ways to further represent that the asset is no longer live, such as setting a `redirect_url` or a `replacement_id`.
+
+The draft state should always match the state of the `parent_document_url`. Recently, a [validation rule](https://github.com/alphagov/asset-manager/blob/cffce4e0e1323eab138e016de9b33536f62fef60/app/models/asset.rb#L288) has been added to ensure this, though no data patch was run for existing data.
+
+Assets must be in draft for certain authorisation protocols to apply. See [documentation](https://github.com/alphagov/asset-manager/blob/cffce4e0e1323eab138e016de9b33536f62fef60/docs/authorisation.md).
+
+2. `state`
+
+This is a representation of the internal Asset Manager processing of the asset, particularly around uploading and virus scanning status.
+The state machine includes `scanned_clean`, `clean`, `scanned_infected`, `upload_success`, `uploaded`.
+
+NB: There are some invalid remnants of a previous state machine, including state values such as `deleted`, in the database. These should be removed.
+
+3. `deleted_at`
+
+The deletion timestamp should be set by the publishing application when an asset is deleted.
+A `nil` value means the asset has not been deleted (default).
+
+4. `replacement_id`
+
+This is a BSON object ID, set by the publishing app when the user uploads a new file in the place of an old one, typically with the intention of preserving some contextual document metadata. 
+Default is `nil`. The replaced asset redirects to its replacement, provided the replacement is not in draft.
+
+Whilst deletion and replacements are mutually exclusive (an asset should not logically have both), they do coexist in the database.
+Deletion, replacement, and draft work together to support previewing of draft content. An asset will not redirect to its replacement if the replacement is in draft.
+This ensures that until the publish event is fired, the users can preview images and documents on the draft stack.
+
+Publishing applications only send the next-in-line `replacement_id`. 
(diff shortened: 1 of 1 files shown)


docs/asset_fields_overview.md
```diff
@@ -0,0 +1,44 @@
+## Asset fields overview
+
+Asset Manager stores a a number of fields whose values work together to enforce how the asset is served.
+These fields are closely related to various document states from the publishing applications.
+Whitehall uses most of the fields, since it implements a full "editioned" flow. The following examples are thus based on Whitehall's behaviour.
```
ryanb-gds, line 5: I'd probably drop this line about Whitehall - we should have some docs in Whitehall about how it interacts with Asset Manager (I think Korny wrote some).

docs/asset_fields_overview.md
```diff
@@ -0,0 +1,44 @@
...
+
+1. `draft`
+
+Assets are set to draft (boolean `true`) when they first get created on a draft edition. When the document gets published, the draft state of the asset is also updated to `false`.
```
ryanb-gds, line 9:
```suggestion
Assets are set to draft (boolean `true`) when they first get created on a draft edition. When the document gets published, the draft state of the asset should be updated to `false` by the publishing application.
```

I think it's important to make it clear that Asset Manager isn't doing some kind of magic document state tracking here.

docs/asset_fields_overview.md
```diff
@@ -0,0 +1,44 @@
...
+2. `state`
+
+This is a representation of the internal Asset Manager processing of the asset, particularly around uploading and virus scanning status.
+The state machine includes `scanned_clean`, `clean`, `scanned_infected`, `upload_success`, `uploaded`.
+
+NB: There are some invalid remnants of a previous state machine, including state values such as `deleted`, in the database. These should be removed.
+
+3. `deleted_at`
+
+The deletion timestamp gets set when an asset is deleted in the corresponding publishing application, either by deleting an attachment or discarding a document.
```
ryanb-gds, line 25:
```suggestion
The deletion timestamp should be set by the publishing application when an asset is deleted.
```
Again, emphasising that the publishing app manages the state. Also suggest dropping the bit about deleting docs and discarding attachments as that's not really relevant to Asset Manager

docs/asset_fields_overview.md
```diff
@@ -0,0 +1,44 @@
+## Asset fields overview
+
+Asset Manager stores a a number of fields whose values work together to enforce how the asset is served.
```
ryanb-gds, line 3:
```suggestion
## Asset state machine overview

Asset Manager stores a a number of fields whose values work together to represent the state of an asset.
```
I know the Asset has a single `state` field, but in reality its state is comprised of all of these fields, as you correctly identify. I think being clear that what we're really documenting here is the state machine is helpful

docs/asset_fields_overview.md
```diff
@@ -0,0 +1,44 @@
...
+Deletion, replacement, and draft work together to support previewing of draft content. An asset will not redirect to its replacement if the replacement is in draft.
+This ensures that until the publish event is fired, the users can preview images and documents on the draft stack.
+
+Publishing applications only send the next-in-line `replacement_id`. 
+It is Asset Manager that deals with backpropagating the replacement - see the `update_indirect_replacements_on_publish` [callback](https://github.com/alphagov/asset-manager/blob/13d68aee4d0c8cf2c81be7fafef309dc36436ca3/app/models/asset.rb#L174).
+This ensures that all assets in a chain of replacements redirect to the latest one.
+
+5. `redirect_url`
+
+This is typically set when the parent edition is unpublished. Default is `nil`. It is usually set to the parent document URL. Because of this, most redirected assets will return a `404` status.
```
ryanb-gds, line 43: It doesn't follow to me that most redirected assets should return a 404 - if they were redirected to the parent document I would expect the parent document to be redirected as well and therefore to return a 301/302 response

docs/asset_fields_overview.md
```diff
@@ -0,0 +1,44 @@
...
+This ensures that until the publish event is fired, the users can preview images and documents on the draft stack.
+
+Publishing applications only send the next-in-line `replacement_id`. 
+It is Asset Manager that deals with backpropagating the replacement - see the `update_indirect_replacements_on_publish` [callback](https://github.com/alphagov/asset-manager/blob/13d68aee4d0c8cf2c81be7fafef309dc36436ca3/app/models/asset.rb#L174).
+This ensures that all assets in a chain of replacements redirect to the latest one.
+
+5. `redirect_url`
+
+This is typically set when the parent edition is unpublished. Default is `nil`. It is usually set to the parent document URL. Because of this, most redirected assets will return a `404` status. 
+Upon re-drafting the parent edition, the redirect URL gets set back to `nil`.
```
ryanb-gds, line 44:
```suggestion
Upon creating a new draft of the parent edition, the publishing app should set the redirect URL back to `nil`.
```
Publishing app emphasis

ryanb-gds (commented):
A few thoughts

docs/asset_fields_overview.md
lauraghiorghisor-tw: That's correct.

PR #1588, opened 2025-01-07
Update Rails default config for 8.0

mike3985:
I forgot to do this the other day [when I upgraded the Rails gems](https://github.com/alphagov/asset-manager/pull/1572).

https://trello.com/c/JJ02il7P/1536-properly-complete-rails-8-upgrades-for-transition-asset-manager-and-signon


config/environments/test.rb
```diff
@@ -32,18 +28,13 @@
...
   # Tell Action Mailer not to deliver emails to the real world.
   # The :test delivery method accumulates sent emails in the
   # ActionMailer::Base.deliveries array.
   config.action_mailer.delivery_method = :test
 
-  # Unlike controllers, the mailer instance doesn't have any context about the
-  # incoming request so you'll need to provide the :host parameter yourself.
-  # config.action_mailer.default_url_options = { host: "www.example.com" }
+  # Set host to be used by links generated in mailer templates.
+  config.action_mailer.default_url_options = { host: "example.com" }
```
yndajas, line 37: **issue:** example config

I don't think we want this default host?

config/application.rb
```diff
@@ -46,6 +46,19 @@ class Application < Rails::Application
...
+    # Configures whether the ActionDispatch::ETag middleware should prefer the
+    # ETag header over the Last-Modified header when both are present in the
+    # response.
+    #
+    # If set to true, when both headers are present only the ETag is considered
+    # as specified by RFC 7232 section 6.
+    #
+    # If set to false, when both headers are present, both headers are checked
+    # and both need to match for the response to be considered fresh.
+    config.action_dispatch.strict_freshness = false
```
yndajas, line 60: **question:** other apps

Is this something we should check in other apps?

config/environments/test.rb
mike3985: I wondered if it really matters because this app doesn't have any mailers and its the test env, so I just accepted the change 🤷

yndajas: Yeah I suppose. Might be best to comment it out to avoid any confusion about mailers in this app though?

config/application.rb
mike3985: I was thinking that this is specifically a concern for Asset Manager because of the service it provides to clients. It might be true for Content Store and a couple of others, too?

config/environments/test.rb
mike3985: Sure thing. Confusion = bad

mike3985: Done!

yndajas (approved):

config/application.rb
mike3985: BTW, here's some evidence of Asset Manager caring about this stuff https://github.com/alphagov/asset-manager/blob/main/spec/controllers/media_controller_spec.rb#L76-L86

PR #1946, opened 2026-05-06
Upgrade Asset-Manager to Ruby 4

callumknights:
2nd time after missing a step on the initial PR


.ruby-version
```diff
@@ -1 +1 @@
-3.4.4
+4.0.3
```
brucebolt, line 1: Version 4.0.3 isn't yet available in [`govuk-ruby-images`](https://github.com/alphagov/govuk-ruby-images/blob/main/build-matrix.json), so this may fail on deployment.

callumknights:
Asset-manager running fine in integration with correct release tag 
image

https://github.com/alphagov/asset-manager/actions/runs/25426747041/job/74583231886

.ruby-version
callumknights: Discussed outside of Github, but this is fine, `govuk-ruby-images` uses the latest available image >4, as defined by the Dockerfile

brucebolt (approved):

.ruby-version
brucebolt: Brilliant, that's good to know for future and definitely improve my understanding of the process.

PR #1948, opened 2026-05-12
[PP-7726] Add new 'scanning' state to the Asset state machine

callumknights:
This is to prevent a possible race condition where a file could be uploaded to be scanned, then while scanning is still going on, the file is updated. Then despite the virus scan being queued again, the original scan would pass and flag the now updated asset as clean. This could be leveraged to pass infected assets through the clamAV scanning.

Implementing the 'scanning' state prevents this by having a transition rule in place that means every asset must have a 'scanning' state before having a 'clean' or 'infected' state. Then if an asset is updated mid-scan, it's state will be reset to 'unscanned' and the original scan will fail with a InvalidTransition Exception.

I've also included a change to the logic of the VirusScanWorker, where it will now check that the hex_digest of the file it's scanning has not changed during the course of the scan. As though the scanning state mostly closed this loophole and stopped infected assets that were still waiting on the queue from being uploaded, if the infected asset scan started while the initial clean asset was still being scanned, a race condition was still possible. This change prevents that by checking the file itself has not changed during the course of the scan.

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -90,8 +90,12 @@ class Asset
       block.call
     end
 
+    event :begun_scan do
+      transition unscanned: :scanning
+    end
+
     event :scanned_clean do
-      transition unscanned: :clean
+      transition scanning: :clean
     end
 
     after_transition to: :clean do |asset, _|
@@ -99,7 +103,7 @@ class Asset
     end
 
     event :scanned_infected do
-      transition unscanned: :infected
+      transition scanning: :infected
     end
 
     event :upload_success do
--- a/app/sidekiq/virus_scan_job.rb
+++ b/app/sidekiq/virus_scan_job.rb
@@ -7,11 +7,13 @@ class VirusScanJob
 
   def perform(asset_id)
     asset = Asset.find(asset_id)
+    initial_digest = asset.md5_hexdigest
     if asset.unscanned?
       begin
+        asset.begun_scan!
         Rails.logger.info("#{asset_id} - VirusScanJob#perform - Virus scan started")
         Services.virus_scanner.scan(asset.file.path)
-        asset.scanned_clean!
+        asset.scanned_clean! if asset.reload.md5_hexdigest == initial_digest
       rescue VirusScanner::InfectedFile => e
         GovukError.notify(e, extra: { id: asset.id, filename: asset.filename })
         asset.scanned_infected!
--- a/spec/factories/factories.rb
+++ b/spec/factories/factories.rb
@@ -3,10 +3,16 @@ FactoryBot.define do
     file { Rack::Test::UploadedFile.new(Rails.root.join("spec/fixtures/files/asset.png")) }
   end
   factory :clean_asset, parent: :asset do
-    after :create, &:scanned_clean!
+    after :create do |a|
+      a.begun_scan!
+      a.scanned_clean!
+    end
   end
   factory :infected_asset, parent: :asset do
-    after :create, &:scanned_infected!
+    after :create do |a|
+      a.begun_scan!
+      a.scanned_infected!
+    end
   end
   factory :uploaded_asset, parent: :clean_asset do
     after :create, &:upload_success!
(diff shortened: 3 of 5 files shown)


yndajas (commented):
This change gives me confidence that we can't go from `:unscanned` to `:clean`, but I'm unclear on where we reset to `:unscanned` when updating an asset

> if an asset is updated mid-scan, it's state [will] be reset to 'unscanned'

Which bit of the code does that, and do we need a (new) test to show that an asset gets reset to `:unscanned` if updated while in the `:scanning` state?

yndajas:
> Which bit of the code does that

Reset state on updating an asset's file: https://github.com/alphagov/asset-manager/blob/a143026284c06587399cf520679f0b6c1a0bc829/app/models/asset.rb#L134

We have a model test for scheduling a new virus scan which is a side effect of setting the state to `:unscanned`, but I can't see a test specifically for resetting the state on file change. It might be nice to add from all of `:scanning`,`:clean`, `:infected`, and `:uploaded`

One more question: do we have multiple workers? If so, is it possible that one worker could be scanning the good/first upload and another scanning the bad/second upload at the same time, in which case the state could have gone from `:scanning` to `:unscanned` and back to `:scanning` before the good finishes? Perhaps that's too much of an edge case 🤷🏻‍♀️ (not an expert!)

callumknights:
> > Which bit of the code does that
> 
> Reset state on updating an asset's file:
> 
> https://github.com/alphagov/asset-manager/blob/a143026284c06587399cf520679f0b6c1a0bc829/app/models/asset.rb#L134
> 
> We have a model test for scheduling a new virus scan which is a side effect of setting the state to `:unscanned`, but I can't see a test specifically for resetting the state on file change. It might be nice to add from all of `:scanning`,`:clean`, `:infected`, and `:uploaded`
> 
> One more question: do we have multiple workers? If so, is it possible that one worker could be scanning the good/first upload and another scanning the bad/second upload at the same time, in which case the state could have gone from `:scanning` to `:unscanned` and back to `:scanning` before the good finishes? Perhaps that's too much of an edge case 🤷🏻‍♀️ (not an expert!)

Good shout on adding a test for the file state being reset. I'll add one in as it does look like that's missing from the coverage.

We do have multiple workers, and funnily enough this wouldn't be a problem if we didn't. As all the virus scan jobs are async, if we had a single worker, the scan post-update would just get queued behind the original, then the state would be reset and the file would be re-scanned. However with multiple workers we run the risk that one worker does the initial scan, then another runs the updated scan on a potentially infected file, but the file itself is updated outside of the workers. So should the first scan pass first, while the second (infected) scan is still running, then the file will be marked as clean and uploaded, despite the fact it's been updated. The 2nd scan would then later throw us the file is infected, then remove it, but it will have already been uploaded to S3.

This removes the case you have described by setting the state to `scanning` before any potentially time consuming logic is performed. So then if the first scan takes a while, then the asset is updated with an infected file, a subsequent scan won't begin until the first is finished, as the asset is no longer `unscanned`.

callumknights:
@yndajas My bad, just realised there is a test for resetting the Asset state on update. It's not a model test, but still fulfils the same case. https://github.com/alphagov/asset-manager/blob/ceb842a2def85c7705ad4b8defe8354d6046b0e1/spec/requests/asset_requests_spec.rb#L52

yndajas:
> This removes the case you have described by setting the state to `scanning` before any potentially time consuming logic is performed. So then if the first scan takes a while, then the asset is updated with an infected file, a subsequent scan won't begin until the first is finished, as the asset is no longer `unscanned`.

I'm a little confused - I guess this is the sequence I'm thinking of:

1. Upload good file -> `:unscanned`
2. Start scan on good file -> `:scanning`
3. Upload bad file -> `:unscanned`
4. Start scan on bad file -> `:scanning`
5. Finish scan on good file -> `:clean`
6. Upload good file -> `:uploaded`
7. Finish scan on bad file -> ?

Does the bad file still get uploaded after the good file (both before and after this change)? Does the introduction of `:scanning` somehow prevent this (can two tests not run simultaneously)? If so, is there a test we could add for this (both before and after this change)?

> there is a test for resetting the Asset state on update

Nice. It doesn't cover reverting to unscanned from each subsequent state - we could maybe do that in model tests - but at least there is some coverage of that behaviour

callumknights:
Superseded by https://github.com/alphagov/asset-manager/pull/1951

PR #1951, opened 2026-05-14
[PP-7726] Fix possible race condition is Asset virus scanning

callumknights:
This fixes a possible race condition in the asset_manager virus scanning where an Asset could be updated shortly after it was initially uploaded, and while the initial virus scan is taking place. Theoretically an infected file could then replace the original, as when the original virus scan reports success (these are run asyncronously), the asset will be marked as clean and uploaded, despite having been altered.

This change prevents that by saving the original hex_digest at the start of the job, then verifying it's the same before granting `clean` status. If not it raises an `AssetReplaced` exception.

Previous approach [here](https://github.com/alphagov/asset-manager/pull/1948) for context 

### DIFF
--- a/app/sidekiq/virus_scan_job.rb
+++ b/app/sidekiq/virus_scan_job.rb
@@ -3,15 +3,18 @@ require "services"
 class VirusScanJob
   include Sidekiq::Job
 
-  sidekiq_options lock: :until_and_while_executing
+  sidekiq_options lock: :until_executing
+
+  class AssetReplaced < StandardError; end
 
   def perform(asset_id)
     asset = Asset.find(asset_id)
     if asset.unscanned?
       begin
+        initial_digest = asset.md5_hexdigest
         Rails.logger.info("#{asset_id} - VirusScanJob#perform - Virus scan started")
         Services.virus_scanner.scan(asset.file.path)
-        asset.scanned_clean!
+        asset.reload.md5_hexdigest == initial_digest ? asset.scanned_clean! : Rails.logger.info("#{asset.id} VirusScanJob checksum failed")
       rescue VirusScanner::InfectedFile => e
         GovukError.notify(e, extra: { id: asset.id, filename: asset.filename })
         asset.scanned_infected!
--- a/spec/sidekiq/virus_scan_job_spec.rb
+++ b/spec/sidekiq/virus_scan_job_spec.rb
@@ -31,6 +31,20 @@ RSpec.describe VirusScanJob do
       allow(scanner).to receive(:scan).and_return(true)
     end
 
+    context "but no longer matches the file associated with the asset" do
+      before do
+        allow(asset).to receive(:md5_hexdigest).twice.and_return("foo", "bar")
+        allow(Asset).to receive(:find).with(asset.id).and_return(asset)
+        allow(Rails.logger).to receive(:info).at_least(:once)
+      end
+
+      it "logs the job failure and does not update the asset's state" do
+        worker.perform(asset.id)
+        expect(Rails.logger).to have_received(:info).with("#{asset.id} VirusScanJob checksum failed").once
+        expect(asset.reload).not_to be_clean
+      end
+    end
+
     it "sets the state to clean" do
       worker.perform(asset.id)
 


app/sidekiq/virus_scan_job.rb
```diff
@@ -5,13 +5,16 @@ class VirusScanJob
...
+
   def perform(asset_id)
     asset = Asset.find(asset_id)
     if asset.unscanned?
       begin
+        initial_digest = asset.md5_hexdigest
         Rails.logger.info("#{asset_id} - VirusScanJob#perform - Virus scan started")
         Services.virus_scanner.scan(asset.file.path)
-        asset.scanned_clean!
+        asset.reload.md5_hexdigest == initial_digest ? asset.scanned_clean! : raise(AssetReplaced)
```
yndajas: Is there a way we can test this? I wonder if we can intercept the `Services.virus_scanner.scan` call, change the file there, and test that the exception is thrown?

Changing the file would be a more realistic test, but if that's not possible - or in addition - you could use something like this to return a different digest the second time without worrying about the actual file:

```rb
allow(asset).to receive(:md5_hexdigest).twice.and_return("abc", "def")

expect { virus_scan_job.perform }.to raise_error(AssetReplaced)
```

callumknights: Had to go with the above suggestion as a test on the VirusScanJob. Spent quite a few hours messing with `Threads` to see if I could figure out a way to intercept a concurrent run of `scan`, then make a post request to alter the asset, and didn't get anywhere.

spec/sidekiq/virus_scan_job_spec.rb
```diff
@@ -49,6 +49,21 @@
...
+      allow(scanner).to receive(:scan).and_raise(exception)
+      allow(asset).to receive(:md5_hexdigest).twice.and_return("foo", "bar")
+    end
+
+    it "raises an exception" do
+      expect(worker).to receive(:perform)
+
+      worker.perform(asset.id)
+    end
+  end
```
yndajas: I think this is making `Services.virus_scanner` raise the exception via the test code rather than testing that an exception is raised by `VirusScanJob` per the app code as a result of the digest change

Did you try something more like this?

```rb
  context "when the file is updated mid scan" do
    before do
      allow(asset).to receive(:md5_hexdigest).twice.and_return("foo", "bar")
    end

    it "raises an exception" do
      expect { worker.perform(asset.id) }.to_raise(AssetError)
    end
  end
```

app/sidekiq/virus_scan_job.rb
brucebolt: What happens when this exception is raised?  Is it logged to Sentry? Will Sidekiq retry the job?

Also, with the lock type being `until_and_while_executing`, will the new file ever be queued for scanning?

callumknights: It should be logged to Sentry yes. And as far as retrying the job goes, when the `update` post request that triggered change of digest comes in, a new run of the job is queued anyway. So further down the queue there will already be a retry waiting to go, with the new updated file. This just prevents the job changing the virus scan status based on outdated file contents.

brucebolt: > It should be logged to Sentry yes.

Do we actually want that to happen? The alerts will be completely unactionable by us, so there's probably little benefit to alerting us.

> And as far as retrying the job goes, when the update post request that triggered change of digest comes in, a new run of the job is queued anyway.

When the file is updated, we'll attempt to queue another virus scan but this will fail because sidekiq-unique-jobs will have locked new jobs from being created until the existing one completes.  The only parameter to the job is `asset_id`, which will be the same even if the file is updated.

Maybe the lock type needs changing to `until_executing`, rather than `until_and_while_executing`, so we do queue a new job when the existing one is running?  Or the job parameters could include the hex digest and lock based on that?

yndajas: Alternatively we could just info log when this happens rather than raise an exception?

But even if we don't raise an exception, would the second scan ever get queued if the first is in progress when first trying to queue it? Maybe we should be doing some of this ^ anyway

callumknights: Ah, I hadn't considered Sidekiq-unique-jobs impact on this, that definitely changes things.

I hear what you're saying about the Sentry alert, me and Ynda had a chat about this and the end non-conclusion was more or less "if they are noisy, we could silence later". But given what you've said, maybe I could change this to just log the alert, and for the hex_digest, explicitly retry the job when the checksum fails. Will the lock status still need changing? As the retry could be being called from inside the executing job.

brucebolt: I do think logging the alert is a better approach than reporting to Sentry, as we'll likely have forgotten context when it does trigger which will lead to unnecessary investigation.

> explicitly retry the job when the checksum fails

I think changing the locking mechanism would solve this, so a job for the same asset can be queued whilst a virus scan is already running for the same asset.

callumknights: Works for me! I'll get on that now

callumknights: Done. I'm not a fan of testing log output, but seems like the only thing to do in this situation. I've squashed and added you as a co-author @yndajas

brucebolt: The logging looks good to me, thanks for changing that.  I think the only thing left is to change the locking mechanism on the job, so we can schedule another virus scan if the file is updated whilst another scan is running.

callumknights: My bad, lost it rebasing, should be there now 👍

brucebolt (approved):
Thanks for changing the lock type.  The code all looks good to me.

Could you add a short note to the commit message explaining why the lock type has changed.  This will make it easier to understand why we changed it in the future, without needing to read through all the PR discussions.

PR #1962, opened 2026-05-21
Add state transition diagram to the docs

AgaDufrat:
To visualise states and transition events on POST /assets API call.

[✨  Rendered doc](https://github.com/alphagov/asset-manager/blob/47e98192df450f64880f986f7e6ea1eab44fb5ff/docs/asset_state_machine_overview.md)

### DIFF
--- a/docs/asset_state_machine_overview.md
+++ b/docs/asset_state_machine_overview.md
@@ -15,9 +15,46 @@ Assets must be in draft for certain authorisation protocols to apply. See [docum
 2. `state`
 
 This is a representation of the internal Asset Manager processing of the asset, particularly around uploading and virus scanning status.
-The state machine includes `scanned_clean`, `clean`, `scanned_infected`, `upload_success`, `uploaded`.
+The state machine includes `scanned_clean`, `clean`, `scanned_infected`, `infected`, `upload_success`, `uploaded`.
 
 NB: There are some invalid remnants of a previous state machine, including state values such as `deleted`, in the database. These should be removed.
+
+```
+┌──────────────┐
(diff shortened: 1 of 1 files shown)


callumknights (approved):
This is fantastic!

docs/asset_state_machine_overview.md
```diff
@@ -15,10 +15,47 @@ Assets must be in draft for certain authorisation protocols to apply. See [docum
 2. `state`
 
 This is a representation of the internal Asset Manager processing of the asset, particularly around uploading and virus scanning status.
-The state machine includes `scanned_clean`, `clean`, `scanned_infected`, `upload_success`, `uploaded`.
+The state machine includes `scanned_clean`, `clean`, `scanned_infected`, `infected`, `upload_success`, `uploaded`.
 
 NB: There are some invalid remnants of a previous state machine, including state values such as `deleted`, in the database. These should be removed.
 
+```
```
yndajas, line 22: **suggestion:** this works for me, but we do use Mermaid diagrams in a few docs, so that might be worth considering

yndajas (approved):

docs/asset_state_machine_overview.md
AgaDufrat: oh! I will have a look. If we're planning to make changes, it will probably be easier to maintain as well.

AgaDufrat: I thought we are already using it in this repo when I wrote the above comment. There's a bit of set up and introduction of JavaScript dependencies. So I think adding mermaid as a dependency it's out of scope at the moment.

PR #1964, opened 2026-06-02
[PP-7719] Add svg scanning to Asset manager

callumknights:
The PR adds the functionality to Asset Manager to scan SVG's for undesirable elements, mainly, but not limited to, script tags. 

The scanning itself relies on the Nokogiri Parser, to which we can pass in undesirable elements, in this case those are defined as `script foreignobject iframe object`.

The new scanner integrates with the existing state machine for each Asset, that was previously used just to check virus scanning and upload status. In addition, we've implemented Mimetype checking for the SVG scanner, so that files that are not SVG's are not unnecessarily scanned. The scanning job itself is run asynchronously in a similar fashion to the existing Virus Scanning job, and includes the hex_digest checksum validation, to ensure it's not vulnerable to a concurrent run/update race condition. 

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -90,8 +90,20 @@ class Asset
       block.call
     end
 
-    event :scanned_clean do
-      transition unscanned: :clean
+    event :virus_scanned_clean do
+      transition unscanned: :virus_scanned_clean
+    end
+
+    after_transition to: :virus_scanned_clean do |asset, _|
+      asset.schedule_svg_scan
+    end
+
+    event :svg_scan_skipped do
+      transition virus_scanned_clean: :clean
+    end
+
+    event :svg_scanned_clean do
+      transition virus_scanned_clean: :clean
     end
 
     after_transition to: :clean do |asset, _|
@@ -100,6 +112,10 @@ class Asset
 
     event :scanned_infected do
       transition unscanned: :infected
+    end
+
+    event :scanned_infected do
+      transition virus_scanned_clean: :infected
     end
 
     event :upload_success do
@@ -219,6 +235,10 @@ class Asset
   def initialize_dup(other)
     @_mounters = nil
     super
+  end
+
+  def schedule_svg_scan
+    Marcel::MimeType.for(Pathname.new(file.path)) == "image/svg+xml" ? SvgScanJob.perform_async(id.to_s) : svg_scan_skipped!
   end
 
 protected
--- a/app/sidekiq/svg_scan_job.rb
+++ b/app/sidekiq/svg_scan_job.rb
@@ -1,0 +1,23 @@
+require "services"
+
+class SvgScanJob
+  include Sidekiq::Job
+  include EnsureFile
+
+  sidekiq_options lock: :until_executing
+
+  def perform(asset_id)
+    asset = Asset.find(asset_id)
+    if asset.scanned_clean?
+      begin
+        Rails.logger.info("#{asset_id} - SvgScanJob#perform - SVG scan started")
+        ensure_file_is_same_after_scan(asset, "SvgScanJob", :svg_scanned_clean!) do
+          Services.svg_scanner.scan(asset.file.path)
+        end
+      rescue SvgDocument::UnsafeSvg => e
+        GovukError.notify(e, extra: { id: asset.id, filename: asset.filename })
+        asset.scanned_infected!
+      end
+    end
+  end
+end
--- a/app/sidekiq/svg_scan_job.rb
+++ b/app/sidekiq/svg_scan_job.rb
@@ -8,7 +8,7 @@ class SvgScanJob
 
   def perform(asset_id)
     asset = Asset.find(asset_id)
-    if asset.scanned_clean?
+    if asset.virus_scanned_clean?
       begin
         Rails.logger.info("#{asset_id} - SvgScanJob#perform - SVG scan started")
         ensure_file_is_same_after_scan(asset, "SvgScanJob", :svg_scanned_clean!) do
--- a/app/sidekiq/virus_scan_job.rb
+++ b/app/sidekiq/virus_scan_job.rb
@@ -22,5 +22,3 @@ class VirusScanJob
     end
   end
 end
-
-VirusScanWorker = VirusScanJob
--- a/app/sidekiq/virus_scan_job.rb
+++ b/app/sidekiq/virus_scan_job.rb
@@ -13,7 +13,7 @@ class VirusScanJob
     if asset.unscanned?
       begin
         Rails.logger.info("#{asset_id} - VirusScanJob#perform - Virus scan started")
-        ensure_file_is_same_after_scan(asset, "VirusScanJob", :scanned_clean!) do
+        ensure_file_is_same_after_scan(asset, "VirusScanJob", :virus_scanned_clean!) do
           Services.virus_scanner.scan(asset.file.path)
         end
       rescue VirusScanner::InfectedFile => e
--- a/app/sidekiq/virus_scan_job.rb
+++ b/app/sidekiq/virus_scan_job.rb
@@ -2,6 +2,7 @@ require "services"
 
 class VirusScanJob
   include Sidekiq::Job
+  include EnsureFile
 
   sidekiq_options lock: :until_executing
 
@@ -11,10 +12,10 @@ class VirusScanJob
     asset = Asset.find(asset_id)
     if asset.unscanned?
       begin
-        initial_digest = asset.md5_hexdigest
         Rails.logger.info("#{asset_id} - VirusScanJob#perform - Virus scan started")
-        Services.virus_scanner.scan(asset.file.path)
-        asset.reload.md5_hexdigest == initial_digest ? asset.scanned_clean! : Rails.logger.info("#{asset.id} VirusScanJob checksum failed")
+        ensure_file_is_same_after_scan(asset, "VirusScanJob", :scanned_clean!) do
+          Services.virus_scanner.scan(asset.file.path)
+        end
       rescue VirusScanner::InfectedFile => e
         GovukError.notify(e, extra: { id: asset.id, filename: asset.filename })
         asset.scanned_infected!
--- a/lib/svg_scanner.rb
+++ b/lib/svg_scanner.rb
@@ -1,0 +1,36 @@
+class SvgDocument < Nokogiri::XML::SAX::Document
+  class UnsafeSvg < StandardError; end
+
+  UNSAFE_ELEMENTS = %w[script foreignobject iframe object].freeze
+
+  def start_element(name, attrs = [])
+    element_name = name.downcase
+
+    if UNSAFE_ELEMENTS.include?(element_name)
+      raise UnsafeSvg, "SVG: Unsafe element detected: <#{name}>"
+    end
+
+    attrs.each do |attr_name, attr_value|
+      attr_name_lower = attr_name.downcase
+
+      if attr_name_lower.start_with?("on") # e.g. onload, onerror, onclick
+        raise UnsafeSvg, "SVG: Unsafe event handler detected: '#{attr_name}' on <#{name}>"
+      end
+
+      if attr_value.to_s.strip.downcase.start_with?("javascript:") # e.g., href="javascript:alert(1)"
+        raise UnsafeSvg, "SVG: Unsafe javascript URI detected in '#{attr_name}' attribute on <#{name}>"
+      end
+    end
+  end
+end
+
+class SvgScanner
+  def scan(file_path)
+    parser = Nokogiri::XML::SAX::Parser.new(SvgDocument.new)
+
+    File.open(file_path) do |file|
+      parser.parse_io(file)
+    end
+    true
+  end
+end
--- a/spec/lib/svg_scanner_spec.rb
+++ b/spec/lib/svg_scanner_spec.rb
@@ -1,0 +1,49 @@
+require "rails_helper"
+require "svg_scanner"
+
+RSpec.describe SvgScanner, type: :file_handler do
+  describe "scanning an SVG file" do
+    subject(:scanner) { described_class.new }
+
+    let(:safe_file_path) { fixture_file_path("asset-safe.svg") }
+    let(:unsafe_element_file_path) { fixture_file_path("asset-unsafe-element.svg") }
+    let(:unsafe_event_handler_file_path) { fixture_file_path("asset-unsafe-event-handler.svg") }
+    let(:unsafe_uri_file_path) { fixture_file_path("asset-unsafe-uri.svg") }
+
+    context "when SVG is safe" do
+      it "returns true" do
+        expect(scanner.scan(safe_file_path)).to be true
+      end
+    end
+
+    context "when SVG contains an unsafe element" do
+      it "raises a relevant error" do
+        expect {
+          scanner.scan(unsafe_element_file_path)
+        }.to raise_error(
+          SvgDocument::UnsafeSvg, "SVG: Unsafe element detected: <script>"
+        )
+      end
+    end
+
+    context "when SVG contains an unsafe event handler" do
+      it "raises a relevant error" do
+        expect {
+          scanner.scan(unsafe_event_handler_file_path)
+        }.to raise_error(
+          SvgDocument::UnsafeSvg, "SVG: Unsafe event handler detected: 'onload' on <svg>"
+        )
+      end
+    end
+
+    context "when SVG contains an unsafe URI" do
+      it "raises a relevant error" do
+        expect {
+          scanner.scan(unsafe_uri_file_path)
+        }.to raise_error(
+          SvgDocument::UnsafeSvg, "SVG: Unsafe javascript URI detected in 'href' attribute on <a>"
+        )
+      end
+    end
+  end
+end
--- a/spec/models/asset_spec.rb
+++ b/spec/models/asset_spec.rb
@@ -516,8 +516,30 @@ RSpec.describe Asset, type: :model do
     end
   end
 
+  describe "scheduling an SVG scan" do
+    it "schedules a scan after a clean virus scan" do
+      a = described_class.new(file: load_fixture_file("asset-safe.svg"))
+
+      expect(SvgScanJob).to receive(:perform_async).with(a.id)
+
+      a.virus_scanned_clean!
+    end
+
+    it "schedules another scan after saving a file" do
+      a = FactoryBot.create(:svg_asset_clean)
+      a.file = load_fixture_file("asset-safe.svg")
+
+      expect(SvgScanJob).to receive(:perform_async).with(a.id)
+
+      a.save!
+
+      allow(Services.virus_scanner).to receive(:scan)
+      VirusScanJob.drain
+    end
+  end
+
   describe "when an asset is marked as clean" do
-    let(:state) { "unscanned" }
+    let(:state) { "virus_scanned_clean" }
     let(:asset) { FactoryBot.build(:asset, state:) }
 
     before do
@@ -525,7 +547,7 @@ RSpec.describe Asset, type: :model do
     end
 
     it "sets the asset state to clean" do
-      asset.scanned_clean!
+      asset.svg_scanned_clean!
 
       expect(asset.reload).to be_clean
     end
@@ -533,14 +555,14 @@ RSpec.describe Asset, type: :model do
     it "schedules saving the asset to cloud storage" do
       expect(SaveToCloudStorageJob).to receive(:perform_async).with(asset.id)
 
-      asset.scanned_clean!
+      asset.svg_scanned_clean!
     end
 
     context "when asset is already clean" do
       let(:state) { "clean" }
 
       it "does not allow the state transition" do
-        expect { asset.scanned_clean! }
+        expect { asset.virus_scanned_clean! }
           .to raise_error(StateMachines::InvalidTransition)
       end
     end
@@ -549,7 +571,7 @@ RSpec.describe Asset, type: :model do
       let(:state) { "infected" }
 
       it "does not allow the state transition" do
-        expect { asset.scanned_clean! }
+        expect { asset.virus_scanned_clean! }
           .to raise_error(StateMachines::InvalidTransition)
       end
     end
@@ -558,7 +580,7 @@ RSpec.describe Asset, type: :model do
       let(:state) { "uploaded" }
 
       it "does not allow the state transition" do
-        expect { asset.scanned_clean! }
+        expect { asset.virus_scanned_clean! }
           .to raise_error(StateMachines::InvalidTransition)
       end
     end
--- a/app/sidekiq/concerns/ensure_file.rb
+++ b/app/sidekiq/concerns/ensure_file.rb
@@ -1,0 +1,10 @@
+module EnsureFile
+  extend ActiveSupport::Concern
+
+  def ensure_file_is_same_after_scan(asset, job_name, next_state)
+    initial_digest = asset.md5_hexdigest
+    Rails.logger.info("#{asset.id} - #{job_name}#perform - scan started")
+    yield
+    asset.reload.md5_hexdigest == initial_digest ? asset.send(next_state) : Rails.logger.info("#{asset.id} #{job_name} checksum failed")
+  end
+end
--- a/docs/asset_state_machine_overview.md
+++ b/docs/asset_state_machine_overview.md
@@ -15,45 +15,26 @@ Assets must be in draft for certain authorisation protocols to apply. See [docum
 2. `state`
 
 This is a representation of the internal Asset Manager processing of the asset, particularly around uploading and virus scanning status.
-The state machine includes `scanned_clean`, `clean`, `scanned_infected`, `infected`, `upload_success`, `uploaded`.
 
 NB: There are some invalid remnants of a previous state machine, including state values such as `deleted`, in the database. These should be removed.
 
-```
-┌──────────────┐
-│  Unscanned   │
-│  (Initial)   │
-└──────┬───────┘
-       │
-       ▼
-[VirusScanJob triggered]
-       │
-       ▼
-[Virus scan performed]
-           │
-   ┌───────┴────────┐
-   │                │
-   ▼                ▼
-(scanned_clean)   (scanned_infected)
-   │                │
-   ▼                ▼
-┌──────────┐    ┌────────────┐
-│  Clean   │    │  Infected  │
-└────┬─────┘    └────────────┘
-     │
-     ▼
-[SaveToCloudStorageJob triggered]
-     │
-     ▼
-[Asset uploaded to AWS S3 bucket]
-     │
-     ▼
-(upload_success)
-     │
-     ▼
-┌──────────┐
-│ Uploaded │
-└──────────┘
+```mermaid
+flowchart TD
+    Unscanned("Unscanned\n(Initial)") --> TriggerVirusScan["VirusScanJob triggered"]
+    TriggerVirusScan --> VirusScanPerformed["Virus scan performed"]
+
+    VirusScanPerformed -->|virus_scanned_clean| VirusScannedClean("Virus Scanned Clean")
+    VirusScanPerformed -->|scanned_infected| Infected("Infected")
+
+    VirusScannedClean --> TriggerSvgScan["SvgScanJob triggered"]
+    TriggerSvgScan --> SvgScanPerformed["SVG scan performed"]
+
+    SvgScanPerformed -->|svg_scanned_clean| Clean("Clean")
+    SvgScanPerformed -->|scanned_infected| Infected("Infected")
+
+    Clean --> TriggerSaveToCloudStorageJob["SaveToCloudStorageJob triggered"]
+    TriggerSaveToCloudStorageJob --> Upload["Asset uploaded to AWS S3 bucket"]
+    Upload -->|upload_success| Uploaded("Uploaded")
 ```
 
 3. `deleted_at`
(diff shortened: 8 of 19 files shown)


app/sidekiq/svg_scan_job.rb
```diff
@@ -0,0 +1,22 @@
...
+class SvgScanJob
+  include Sidekiq::Job
+
+  sidekiq_options lock: :until_and_while_executing
+
+  def perform(asset_id)
+    asset = Asset.find(asset_id)
+    begin
+      initial_digest = asset.md5_hexdigest
+      Rails.logger.info("#{asset_id} - SvgScanJob#perform - SVG scan started")
```
AgaDufrat: Is there a particular need for surfacing `asset_id` in the logs? 
We already log start and end of a job with a job id ([see in Kibana](https://kibana.logit.io/s/42f4d2d5-e9ce-451f-8ffc-cdb25bd624f8/app/data-explorer/discover#?_a=(discover:(columns:!(logjson_message,message),isDirty:!f,sort:!()),metadata:(indexPattern:'filebeat-*',view:discover))&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15d,to:now))&_q=(filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'filebeat-*',key:logjson.class,negate:!f,params:(query:VirusScanJob),type:phrase),query:(match_phrase:(logjson.class:VirusScanJob)))),query:(language:kuery,query:'')))). For debugging we can get the asset_id via Sidekiq UI or API.

spec/lib/mimetype_inferrer_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    let(:svg_file_path) { fixture_file_path("asset.svg") }
+    let(:non_svg_file_path) { fixture_file_path("asset.png") }
+    let(:non_existant_file_path) { fixture_file_path("none") }
+    let(:svg_without_extension) { fixture_file_path("asset-svg-without-extension") }
+    let(:svg_mime_type) { "image/svg+xml" }
+
+    # asset-with-capitalised-extension.TXT
+    # asset-with-unregistered-mimetype-extension.doc
+    # asset-without-extension
```
mike3985: stray comments?

spec/lib/mimetype_inferrer_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    it "calls out to 'file'" do
+      output = ""
+      status = instance_double(Process::Status, exitstatus: 0)
+      allow(Open3).to receive(:capture2e).and_return([output, status])
```
mike3985: Could these setup lines before the `expect` be replaced with `allow(Open3).to receive(:capture2e).and_call_original`?

app/sidekiq/svg_scan_job.rb
```diff
@@ -0,0 +1,22 @@
...
+      Services.svg_scanner.scan(asset.file.path)
+      asset.reload.md5_hexdigest == initial_digest ? asset.svg_scanned_clean! : Rails.logger.info("#{asset.id} SvgScanJob checksum failed")
```
AgaDufrat: It maybe nice to add a note in the commit message that MD5 is only used to detect whether the file changed during scanning as it's not obvious without prior knowledge of the work added in another PR.

spec/lib/mimetype_inferrer_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+        output = "ERROR: cannot stat `#{non_existant_file_path}' (No such file or directory)"
+        expect {
+          inferrer.infer(non_existant_file_path)
+        }.to raise_error(MimetypeInferrer::MimetypeInferenceError, output)
+      end
+    end
+
+    context "when the file is an SVG" do
+      it "returns the SVG mimetype" do
+        expect(inferrer.infer(svg_file_path)).to eq(svg_mime_type)
```
mike3985: It could just be me, but the duplication of the string literal itself might read quite well in these `expect`s? "SVG mimetype" is mentioned in all of these examples' descriptions already, so it isn't like the magic string and its name would be far apart 🤷

app/sidekiq/svg_scan_job.rb
```diff
@@ -0,0 +1,22 @@
...
+    rescue SvgScanner::UnsafeSvgError => e
+      GovukError.notify(e, extra: { id: asset.id, filename: asset.filename })
+      asset.scanned_infected!
+    end
+  end
+end
+
+SvgScanWorker = SvgScanJob
```
AgaDufrat: This alias is no longer needed. It was only required to safely upgrade to Sidekiq 7.

app/sidekiq/svg_scan_job.rb
```diff
@@ -0,0 +1,22 @@
+require "services"
+
+class SvgScanJob
+  include Sidekiq::Job
+
+  sidekiq_options lock: :until_and_while_executing
```
mike3985: I have to admit, I've never understood the use case for this lock type. I can't see an explanation anywhere for the choice in this case?

app/sidekiq/svg_scan_job.rb
mike3985: What's the scenario where this check would fail?

app/sidekiq/svg_scan_job.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def perform(asset_id)
+    asset = Asset.find(asset_id)
+    begin
+      initial_digest = asset.md5_hexdigest
+      Rails.logger.info("#{asset_id} - SvgScanJob#perform - SVG scan started")
+      Services.svg_scanner.scan(asset.file.path)
+      asset.reload.md5_hexdigest == initial_digest ? asset.svg_scanned_clean! : Rails.logger.info("#{asset.id} SvgScanJob checksum failed")
+    rescue SvgScanner::UnsafeSvgError => e
```
AgaDufrat: Can you pls add tests for this job? 
I can't see this working as intended given `SvgScanner::UnsafeSvgError ` is not defined anywhere.

app/sidekiq/svg_scan_job.rb
mike3985: Ah, I just realised that this is similar to what was added to the virus scan job and I hadn't caught up on since getting back from my hols. I do worry that the design still leaves an opening for a race condition (that I expected the new `scanning` state to mitigate), but I'm happy to talk about that separately to this pull request!

spec/lib/svg_scanner_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    let(:unsafe_event_handler_file_path) { fixture_file_path("asset-unsafe-event-handler.svg") }
+    let(:unsafe_uri_file_path) { fixture_file_path("asset-unsafe-uri.svg") }
+    let(:nokogiri_parser) { instance_double(Nokogiri::XML::SAX::Parser) }
+
+    before do
+      allow(Nokogiri::XML::SAX::Parser).to receive(:new).and_return(nokogiri_parser)
+    end
+
+    it "calls out to Nokogiri" do
+      allow(nokogiri_parser).to receive(:parse).with(File.read(safe_file_path))
```
mike3985: I think the `expect` sets up the stub as well as the assertion, so this `allow` should be redundant?

spec/lib/svg_scanner_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+
+      expect(nokogiri_parser).to receive(:parse).with(File.read(safe_file_path))
+      scanner.scan(safe_file_path)
+    end
+
+    context "when SVG is safe" do
+      it "returns true" do
+        allow(nokogiri_parser).to receive(:parse).with(File.read(safe_file_path))
```
mike3985: I've re-read it a couple of times and I'm still confused... it looks to me like most of the examples are stubbing-out the thing I expected them to be testing. (Like, `it "calls out to Nokogiri"` is the only one that actually needs the `nokogiri_parser`.) What am I missing?

lib/svg_scanner.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+  end
+end
+
+class SvgScanner
+  def scan(file_path)
+    svg_scanner = SvgDocument.new
+    parser = Nokogiri::XML::SAX::Parser.new(svg_scanner)
+
+    parser.parse(File.read(file_path))
```
AgaDufrat: I’m a bit confused about the implementation here. We’re using SAX in `SvgDocument` to try to avoid reading the entire file 🎉 but then load the the entire file to the memory anyway.
Have you considered parsing an IO object, which will be more memory-efficient and in line with the streaming nature of SAX parsing.

``` suggestion
  File.open(file_path) do |file|
    parser.parse_io(file)
  end
```

app/models/asset.rb
```diff
@@ -221,6 +233,17 @@ def initialize_dup(other)
...
+  # must be a public method for the state machine to call it
+  def schedule_svg_scan
+    begin
+      mimetype = Services.mimetype_inferrer.infer(file.path)
+    rescue MimetypeInferrer::MimetypeInferenceError => e
+      GovukError.notify(e, extra: { id: asset.id, filename: asset.filename })
+      svg_scan_skipped!
+    end
+    mimetype == "image/svg+xml" ? SvgScanJob.perform_async(id.to_s) : svg_scan_skipped!
+  end
```
mike3985, line 242: Do you think it'd be too expensive to perform this check at the beginning of the job itself?

spec/lib/svg_scanner_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    before do
+      allow(Nokogiri::XML::SAX::Parser).to receive(:new).and_return(nokogiri_parser)
+    end
+
+    it "calls out to Nokogiri" do
+      allow(nokogiri_parser).to receive(:parse).with(File.read(safe_file_path))
+
+      expect(nokogiri_parser).to receive(:parse).with(File.read(safe_file_path))
+      scanner.scan(safe_file_path)
+    end
```
AgaDufrat: I can't see what behaviour we are testing here? It doesn't assert anything.

spec/lib/svg_scanner_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    subject(:scanner) { described_class.new }
+
+    let(:safe_file_path) { fixture_file_path("asset-safe.svg") }
+    let(:unsafe_element_file_path) { fixture_file_path("asset-unsafe-element.svg") }
+    let(:unsafe_event_handler_file_path) { fixture_file_path("asset-unsafe-event-handler.svg") }
+    let(:unsafe_uri_file_path) { fixture_file_path("asset-unsafe-uri.svg") }
+    let(:nokogiri_parser) { instance_double(Nokogiri::XML::SAX::Parser) }
+
+    before do
+      allow(Nokogiri::XML::SAX::Parser).to receive(:new).and_return(nokogiri_parser)
```
AgaDufrat: We seem to be over mocking in this test suite given we have `allow(nokogiri_parser).to receive(:parse)` later.
We don't seem to be testing the real parsing anywhere. The only thing we are effectively testing is `parser.parse(...)`

spec/lib/svg_scanner_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    context "when SVG contains an unsafe element" do
+      it "raises a relevant error" do
+        allow(nokogiri_parser).to receive(:parse).with(File.read(unsafe_element_file_path)).and_raise(SvgDocument::UnsafeSvg, "SVG: Unsafe element detected: <script>")
+
+        expect {
+          scanner.scan(unsafe_element_file_path)
+        }.to raise_error(
+          SvgDocument::UnsafeSvg, "SVG: Unsafe element detected: <script>"
+        )
```
AgaDufrat: What do you think of removing the Nokogiri mocking entirely to test the parsing and event handler:
``` suggestion
  expect { scanner.scan(unsafe_element_file_path) }.to raise_error(SvgDocument::UnsafeSvg)
```

lib/mimetype_inferrer.rb
```diff
@@ -0,0 +1,36 @@
+require "open3"
+
+# Simple wrapper around the Unix 'file' utility
+#
+# This expects AssetManager.govuk.file_utility_path to be an executable command
+# that is compatible with 'file'.
+#
+# Note that this does not try to look inside encrypted or compressed files.
+class MimetypeInferrer
+  class Error < StandardError; end
```
AgaDufrat: This class doesn't seem to be used anywhere.

lib/mimetype_inferrer.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    case status.exitstatus
+    when 0
+      out_str
+    else
+      # It might have failed to infer a mimetype,
+      # or the file might be missing,
+      # or we might lack permission to access it.
+      raise MimetypeInferenceError, out_str
+    end
```
AgaDufrat: Suggestion to avoid case statement using `.success?` mentioned in the [Open3 docs](https://ruby-doc.org/stdlib-3.0.0/libdoc/open3/rdoc/Open3.html)

``` suggestion
    return out_str if status.success?

    raise MimetypeInferenceError, "Failed to infer MIME type: #{out_str}"
```

spec/lib/mimetype_inferrer_spec.rb
AgaDufrat: Some development notes still hanging around 😅

spec/lib/mimetype_inferrer_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    # asset-with-capitalised-extension.TXT
+    # asset-with-unregistered-mimetype-extension.doc
+    # asset-without-extension
+
+    it "calls out to 'file'" do
+      output = ""
+      status = instance_double(Process::Status, exitstatus: 0)
+      allow(Open3).to receive(:capture2e).and_return([output, status])
+      expect(Open3).to receive(:capture2e).with("file", "--mime-type", "--brief", "-E", svg_file_path.to_s)
```
AgaDufrat: Do we need both of those?

app/models/asset.rb
```diff
@@ -221,6 +233,17 @@ def initialize_dup(other)
     super
   end
 
+  # must be a public method for the state machine to call it
```
AgaDufrat: I think this comment can go into the commit message.

AgaDufrat (commented):
I left a few comments. I think we should test it in integration. We should also update the state machine docs

We will also need to update Sentry alerts for Asset Manager to notify us on every instance.

app/sidekiq/svg_scan_job.rb
```diff
@@ -0,0 +1,22 @@
...
+      GovukError.notify(e, extra: { id: asset.id, filename: asset.filename })
```
brucebolt: Will there be a related PR in the dev docs, to explain to us what action should be taken when we're alerted about this?

app/sidekiq/svg_scan_job.rb
brucebolt: I wonder if this (and the other MD5 check) could be moved into a shared method, to avoid duplication?

Something like:
```rb
def ensure_file_is_same_after_scan(asset, job_name)
  initial_digest = asset.md5_hexdigest
  Rails.logger.info("#{asset_id} - #{job_name}#perform - SVG scan started") 
  yield
  asset.reload.md5_hexdigest == initial_digest ? asset.svg_scanned_clean! : Rails.logger.info("#{asset.id} #{job_name} checksum failed")
end
```

And then replace this with (plus the same in the other virus scan job):
```rb
ensure_file_is_same_after_scan(asset, "SvgScanJob") do
  Services.svg_scanner.scan(asset.file.path)
end
```

brucebolt: I wonder if this (and the other MD5 check) could be moved into a shared method, to avoid duplication?

Something like:
```rb
def ensure_file_is_same_after_scan(asset, job_name)
  initial_digest = asset.md5_hexdigest
  yield
  asset.reload.md5_hexdigest == initial_digest ? asset.svg_scanned_clean! : Rails.logger.info("#{asset.id} #{job_name} checksum failed")
end
```

And then replace this with (plus the same in the other virus scan job):
```rb
ensure_file_is_same_after_scan(asset, "SvgScanJob") do
  Rails.logger.info("#{asset_id} - SvgScanJob#perform - SVG scan started") 
  Services.svg_scanner.scan(asset.file.path)
end
```

yndajas: > I do worry that the design still leaves an opening for a race condition (that I expected the new scanning state to mitigate), but I'm happy to talk about that separately to this pull request!

I'd be curious to hear more. I expressed concern that the proposed scanning state wouldn't prevent a race condition: https://github.com/alphagov/asset-manager/pull/1948#issuecomment-4439518293

app/sidekiq/svg_scan_job.rb
nacnudus: This should be `:until_executing`, as in the virus scanner job, which was recently changed, but after this new code had already been cribbed from it. See https://github.com/alphagov/asset-manager/pull/1951/changes/90f050b2600ec0d86e7de71fd003e4c7c1373155

app/sidekiq/svg_scan_job.rb
nacnudus: Do you mean that we could use the job start/end timestamps to find a job in Sidekiq, and get the `asset_id` from the job? If there's no harm in including the ID in the logs, then I for one would find that much easier and less ambiguous.

nacnudus: app/sidekiq/virus_scan_job.rb also logs the `asset_id`

app/sidekiq/svg_scan_job.rb
nacnudus: Good spot! This shoudl be `UnsafeSvg` not `UnsafeSvgError`.

app/sidekiq/svg_scan_job.rb
nacnudus: Good question. There doesn't seem to be any documentation of the corresponding notification by the virus scanner. We ought to discuss whether a notification is useful.

lib/svg_scanner.rb
nacnudus: Thanks, your're right.

lib/mimetype_inferrer.rb
nacnudus: I like `return out_str if status.exitstatus == 0` but I'm wary of using `.success?` because it doesn't actually seem to be documented, or at least I can't see where.

app/models/asset.rb
nacnudus: I don't think it would be computationally expensive, but an SVG scan job would be logged for every file instead of only relevant files. Would that be a good thing?

spec/models/asset_spec.rb
```diff
@@ -516,31 +516,53 @@
...
+  describe "scheduling an SVG scan" do
+    it "schedules a scan after a clean virus scan" do
+      a = described_class.new(file: load_fixture_file("asset.svg"))
+
+      expect(SvgScanJob).to receive(:perform_async).with(a.id)
+
+      a.virus_scanned_clean!
+    end
+
+    it "schedules a scan after save if the file is changed" do
```
nacnudus:
```suggestion
    it "schedules another scan after saving a file" do
```

app/models/asset.rb
mike3985: It's a personal taste thing. To me, it'd be an improvement to have this code live closer to the job because this code has more in common with the job than it does with the state machine definition or the Asset model

callumknights: My main view on this was log pollution as well to be honest. Given we log the start and end of the SVG job, I _personally_ think it sounds undesirable to fill the logs with instances of "Job started". In fact without a bit more bespoke logging, if we were to mimetype check as part of the job currently, to someone reading logit it would seem like every file is an SVG, as the logs would be "SVG Scan started", then a state transition to "clean", so on and so forth.

lib/mimetype_inferrer.rb
AgaDufrat: It's documented here https://ruby-doc.org/3.4.1/Process/Status.html#method-i-success-3F

AgaDufrat: Reasons why I prefer it is because:
- it's more ruby idiomatic
- the intent is clearer "succeeded or failed", you don't need to know that `exit 0` means success (although it is common knowledge with experience)
- it's shorter, less code
- easier to read, as the happy path is immediately visible
- avoids unnecessary `case` with only two outcomes

but I'll leave it up to you to decide as it's not blocking.

spec/lib/mimetype_inferrer_spec.rb
nacnudus: Oh, nice 🙇

spec/lib/mimetype_inferrer_spec.rb
nacnudus: Apparently not. This works:
```
expect(Open3).to receive(:capture2e).with("file", "--mime-type", "--brief", "-E", svg_file_path.to_s).and_call_original
```

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,8 @@
 module ApplicationHelper
+  def ensure_file_is_same_after_scan(asset, job_name, next_state)
```
brucebolt: Would this be better placed in `app/sidekiq/concerns`/ rather than the Rails view helpers?

app/sidekiq/virus_scan_job.rb
```diff
@@ -11,10 +12,10 @@ def perform(asset_id)
     asset = Asset.find(asset_id)
     if asset.unscanned?
       begin
-        initial_digest = asset.md5_hexdigest
         Rails.logger.info("#{asset_id} - VirusScanJob#perform - Virus scan started")
-        Services.virus_scanner.scan(asset.file.path)
-        asset.reload.md5_hexdigest == initial_digest ? asset.scanned_clean! : Rails.logger.info("#{asset.id} VirusScanJob checksum failed")
+        ensure_file_is_same_after_scan(asset, "VirusScanJob", :virus_scanned_clean!) do
```
brucebolt, line 16: Should this be `:scanned_clean!` in this commit, then updated later, as I don't think `:virus_scanned_clean!` exists at this point?

lib/svg_scanner.rb
```diff
@@ -0,0 +1,41 @@
...
+  def start_element(name, attrs = [])
+    element_name = name.downcase
+
+    if UNSAFE_ELEMENTS.include?(element_name)
+      raise UnsafeSvg, "SVG: Unsafe element detected: <#{name}>"
+    end
+
+    attrs_hash = attrs.to_h
+
+    attrs_hash.each do |attr_name, attr_value|
```
brucebolt: Is it possible to iterate over `attrs` without converting it to a hash first?

lib/svg_scanner.rb
```diff
@@ -0,0 +1,41 @@
...
+      if attr_value.to_s.strip.downcase.start_with?("javascript:") # e.g., href="javascript:alert(1)"
+        raise UnsafeSvg, "SVG: Unsafe javascript URI detected in '#{attr_name}' attribute on <#{name}>"
+      end
+    end
+  end
+end
+
+class SvgScanner
+  def scan(file_path)
+    svg_scanner = SvgDocument.new
```
brucebolt: Does this need to be a variable?  It's only used once.

lib/mimetype_inferrer.rb
```diff
@@ -0,0 +1,28 @@
...
+      "-E", # Return a non-zero exit status on failure
+      file_path.to_s,
+    ]
+
+    out_str, status = Open3.capture2e(*cmd)
+    out_str.strip!
+
+    return out_str if status.success?
+
+    raise MimetypeInferenceError, "Failed to infer MIME type: #{out_str}"
```
brucebolt: Have you considered using `Marcel::MimeType`, which does much the same check and is a default library in Rails?

spec/lib/svg_scanner_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    it "calls out to Nokogiri" do
+      nokogiri_parser = instance_double(Nokogiri::XML::SAX::Parser)
+      allow(Nokogiri::XML::SAX::Parser).to receive(:new).and_return(nokogiri_parser)
+      allow(nokogiri_parser).to receive(:parse_io).with(an_object_having_attributes(path: safe_file_path))
+
+      expect(nokogiri_parser).to receive(:parse_io)
+
+      scanner.scan(safe_file_path)
+    end
```
brucebolt: I wonder if this test is needed? It doesn't really matter how `SvgScanner` determines the result, it's the outcome we want to test (which is already covered by the other tests).

app/sidekiq/virus_scan_job.rb
```diff
@@ -22,5 +22,3 @@ def perform(asset_id)
     end
   end
 end
-
-VirusScanWorker = VirusScanJob
```
brucebolt, line 26: Is it worth adding a brief sentence to the commit message explaining why this alias was there in the first place (to allow existing enqueued jobs with the old name to complete whilst the worker was renamed)?

brucebolt (commented):
This is looking good to me, thanks for making changes after the previous reviews.

I've left a few small comments, but otherwise this should be ready to merge very soon.

Could you also update the docs in `docs/asset_state_machine_overview.md` to reflect the changes to the state machine.

spec/lib/svg_scanner_spec.rb
nacnudus: I agree. I've removed it.

lib/svg_scanner.rb
nacnudus: Good point.

lib/svg_scanner.rb
nacnudus: Yes.

app/sidekiq/virus_scan_job.rb
nacnudus: That's a good spot.

app/helpers/application_helper.rb
nacnudus: I've moved it into `app/sidekiq/concerns/ensure_file.rb` in a module called `EnsureFile`.

lib/mimetype_inferrer.rb
nacnudus: I didn't know about marcel - somehow it didn't show up in any search results.

I've tried it, and it detects the `asset.svg` as `"application/octet-stream"`. The contents of that file are

```
<svg xmlns="http://www.w3.org/2000/svg"/>
```

That seems to be a bug that was [fixed](https://github.com/rails/marcel/commit/631b328180298a5193f2092042f4910e8b38d2d9) in October, and [released](https://github.com/rails/marcel/releases/tag/v1.2.0) three weeks ago (a fortnight after beginning this ticket!). Unfortunately v1.2.0 seems to be incompatible with CarrierWave <= v2, in a way that I don't understand but that involves the mimemagic gem, which is why bundler installs the earlier, buggier version.

app/sidekiq/virus_scan_job.rb
```diff
@@ -15,8 +15,8 @@ def perform(asset_id)
         Rails.logger.info("#{asset_id} - VirusScanJob#perform - Virus scan started")
         Services.virus_scanner.scan(asset.file.path)
         asset.reload.md5_hexdigest == initial_digest ? asset.scanned_clean! : Rails.logger.info("#{asset.id} VirusScanJob checksum failed")
-      rescue VirusScanner::InfectedFile => e
-        GovukError.notify(e, extra: { id: asset.id, filename: asset.filename })
+      rescue VirusScanner::InfectedFile
+        Rails.logger.info("#{asset_id} - VirusScanJob#perform - Virus detected")
```
brucebolt: I didn't spot this before: should this be `debug` instead of `info`? It means we'll get it logged in production, which could be useful if we need to do investigation.

brucebolt: Although I wonder (thinking with [this comment](https://github.com/alphagov/asset-manager/pull/1964#discussion_r3371203294)) whether we should stick with the alerts but actually document what to do when one occurs (e.g. find the user and notify them, or at least notify the team who own the publishing app it was submitted with)?

app/sidekiq/virus_scan_job.rb
```diff
@@ -1,7 +1,9 @@
 require "services"
 
 class VirusScanJob
+  include ApplicationHelper
```
brucebolt: Is this needed now that the helper has been moved to `EnsureFile`?

app/sidekiq/svg_scan_job.rb
```diff
@@ -0,0 +1,24 @@
+require "services"
+
+class SvgScanJob
+  include ApplicationHelper
```
brucebolt: Same as my comment on the previous commit, I don't think this is needed here anymore.

lib/svg_scanner.rb
```diff
@@ -0,0 +1,38 @@
+require "nokogiri"
```
brucebolt: Is this needed?  I think nokogiri will get included because it's in Gemfile.

lib/mimetype_inferrer.rb
brucebolt: That's strange because when I replace the `infer` method with just `Marcel::MimeType.for(Pathname.new(file_path))`, all the tests pass for me, except two.

The two failing tests are the one which tests for the call to Open3 (which is understandable, as we're not using that anymore) and the one testing for `none` (which probably just needs the error handling updated).

brucebolt (commented):
This is looking good.  I've left a few more comments.

lib/mimetype_inferrer.rb
nacnudus: Odd. It now passes all tests for me, and all I've done is replace the `infer` method as you have described. The two tests that fail for you don't fail for me.

app/sidekiq/virus_scan_job.rb
nacnudus: We discussed this offline and decided to 
- leave the virus scan job notification as before
- make the svg job notify in the same way as the virus scan job

I have removed the commit that changed the virus scan job notification/logging, and I've changed the svg job to notify in the same way as the virus scan job.

brucebolt (approved):
Thanks for making the changes.  This all looks good to me now.

PR #1991, opened 2026-06-30
Pp 7805 scan existing svg files for potentially malicious content

nacnudus:
This creates a rake task that will put a batch of already-uploaded assets into a new, lowest-priority queue, to be scanned for SVG vulnerabilities. If the queue isn't empty, then it will do nothing, so that the queue won't grow too big for redis.

This is intended to be used by a cron job, yet to be created in govuk-helm-charts.

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -53,6 +53,10 @@ class Asset
 
   field :deleted_at, type: Time
 
+  field :svg_scanned_at, type: Time
+
+  field :svg_scanned_safe, type: Boolean
+
   validates :file, presence: true, if: :unscanned?
 
   validates :uuid,
@@ -239,6 +243,10 @@ class Asset
 
   def schedule_svg_scan
     Marcel::MimeType.for(Pathname.new(file.path)) == "image/svg+xml" ? SvgScanJob.perform_async(id.to_s) : svg_scan_skipped!
+  end
+
+  def schedule_svg_batch_scan
+    SvgScanBatchJob.perform_async(id.to_s)
   end
 
 protected
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -114,11 +114,11 @@ class Asset
       SaveToCloudStorageJob.perform_async(asset.id.to_s)
     end
 
-    event :scanned_infected do
+    event :virus_scanned_infected do
       transition unscanned: :infected
     end
 
-    event :scanned_infected do
+    event :svg_scanned_infected do
       transition virus_scanned_clean: :infected
     end
 
@@ -129,6 +129,21 @@ class Asset
     after_transition to: :uploaded do |asset, _|
       asset.save!
       DeleteAssetFileFromNfsJob.perform_in(5.minutes, asset.id.to_s)
+    end
+
+    after_transition on: :virus_scanned_infected do |asset, _|
+      Rails.logger.warn("#{asset.id} - Virus Scan - File #{asset.filename} marked as infected")
+    end
+
+    after_transition on: :svg_scanned_clean do |asset, _|
+      asset.set(svg_scanned_at: Time.zone.now)
+      asset.set(svg_scanned_safe: true)
+    end
+
+    after_transition on: :svg_scanned_infected do |asset, _|
+      asset.set(svg_scanned_safe: false)
+      asset.set(svg_scanned_at: Time.zone.now)
+      Rails.logger.warn("#{asset.id} - SVG Scan - File #{asset.filename} marked as unsafe")
     end
   end
 
--- a/app/sidekiq/concerns/ensure_file.rb
+++ b/app/sidekiq/concerns/ensure_file.rb
@@ -1,14 +1,12 @@
 module EnsureFile
   extend ActiveSupport::Concern
 
-  def ensure_file_is_same_after_scan(asset, job_name, next_state)
+  def ensure_file_is_same_after_scan(asset, job_name, success_callback)
     initial_digest = asset.md5_hexdigest
     Rails.logger.info("#{asset.id} - #{job_name}#perform - scan started")
     yield
     if asset.reload.md5_hexdigest == initial_digest
-      asset.send(next_state)
-      asset.set(svg_scanned_safe: true)
-      asset.set(svg_scanned_at: Time.zone.now)
+      success_callback.call
     else
       Rails.logger.info("#{asset.id} #{job_name} checksum failed")
     end
--- a/app/sidekiq/concerns/ensure_file.rb
+++ b/app/sidekiq/concerns/ensure_file.rb
@@ -5,6 +5,12 @@ module EnsureFile
     initial_digest = asset.md5_hexdigest
     Rails.logger.info("#{asset.id} - #{job_name}#perform - scan started")
     yield
-    asset.reload.md5_hexdigest == initial_digest ? asset.send(next_state) : Rails.logger.info("#{asset.id} #{job_name} checksum failed")
+    if asset.reload.md5_hexdigest == initial_digest
+      asset.send(next_state)
+      asset.set(svg_scanned_safe: true)
+      asset.set(svg_scanned_at: Time.zone.now)
+    else
+      Rails.logger.info("#{asset.id} #{job_name} checksum failed")
+    end
   end
 end
--- a/app/sidekiq/svg_scan_batch_job.rb
+++ b/app/sidekiq/svg_scan_batch_job.rb
@@ -31,7 +31,7 @@ private
     asset.set(svg_scanned_safe: true)
   rescue SvgDocument::UnsafeSvg => e
     asset.set(svg_scanned_safe: false)
-    Rails.logger.info("#{asset.id} - SvgScanBatchJob#perform - SVG unsafe")
+    Rails.logger.warn("#{asset.id} - SVG Scan - File #{asset.filename} marked as unsafe")
     GovukError.notify(e, extra: { id: asset.id, filename: asset.filename })
   ensure
     asset.set(svg_scanned_at: Time.zone.now)
--- a/app/sidekiq/svg_scan_batch_job.rb
+++ b/app/sidekiq/svg_scan_batch_job.rb
@@ -1,0 +1,39 @@
+require "services"
+
+class SvgScanBatchJob
+  include Sidekiq::Job
+
+  sidekiq_options lock: :until_executing, queue: "batch"
+
+  def perform(asset_id)
+    asset = Asset.find(asset_id)
+
+    file = Services.cloud_storage.download(asset)
+
+    if Marcel::MimeType.for(Pathname.new(file.path)) == "image/svg+xml"
+      scan_svg(asset, file.path)
+    end
+  rescue Aws::S3::Errors::NoSuchKey
+    # File doesn't exist in S3. Nothing to do.
+    Rails.logger.info("#{asset_id} - SvgScanBatchJob#perform - Asset missing from S3")
+  ensure
+    if file.respond_to?(:path) && File.exist?(file.path)
+      file.close
+      file.unlink
+    end
+  end
+
+private
+
+  def scan_svg(asset, file_path)
+    Rails.logger.info("#{asset.id} - SvgScanBatchJob#perform - SVG scan started")
+    Services.svg_scanner.scan(file_path)
+    asset.set(svg_scanned_safe: true)
+  rescue SvgDocument::UnsafeSvg => e
+    asset.set(svg_scanned_safe: false)
+    Rails.logger.info("#{asset.id} - SvgScanBatchJob#perform - SVG unsafe")
+    GovukError.notify(e, extra: { id: asset.id, filename: asset.filename })
+  ensure
+    asset.set(svg_scanned_at: Time.zone.now)
+  end
+end
--- a/app/sidekiq/svg_scan_job.rb
+++ b/app/sidekiq/svg_scan_job.rb
@@ -11,15 +11,12 @@ class SvgScanJob
     if asset.virus_scanned_clean?
       begin
         Rails.logger.info("#{asset_id} - SvgScanJob#perform - SVG scan started")
-        ensure_file_is_same_after_scan(asset, "SvgScanJob", :svg_scanned_clean!) do
+        ensure_file_is_same_after_scan(asset, "SvgScanJob", -> { asset.svg_scanned_clean! }) do
           Services.svg_scanner.scan(asset.file.path)
         end
       rescue SvgDocument::UnsafeSvg
-        Rails.logger.warn("#{asset_id} - SvgScanJob#perform - File #{asset.filename} marked as unsafe")
-        asset.scanned_infected!
-        asset.set(svg_scanned_safe: false)
+        asset.svg_scanned_infected!
       end
-      asset.set(svg_scanned_at: Time.zone.now)
     end
   end
 end
--- a/app/sidekiq/svg_scan_job.rb
+++ b/app/sidekiq/svg_scan_job.rb
@@ -17,7 +17,9 @@ class SvgScanJob
       rescue SvgDocument::UnsafeSvg
         Rails.logger.warn("#{asset_id} - SvgScanJob#perform - File #{asset.filename} marked as unsafe")
         asset.scanned_infected!
+        asset.set(svg_scanned_safe: false)
       end
+      asset.set(svg_scanned_at: Time.zone.now)
     end
   end
 end
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -3,3 +3,4 @@
 :queues:
   - default
   - low_priority
+  - batch
--- a/lib/s3_storage/fake.rb
+++ b/lib/s3_storage/fake.rb
@@ -12,6 +12,13 @@ class S3Storage
       File.write(target_path, File.read(source_path))
     end
 
+    def download(asset, **_args)
+      source_path = source_path_for(asset)
+      download_path = download_path_for(asset)
+      FileUtils.mkdir_p(File.dirname(download_path))
+      File.write(download_path, File.read(source_path))
+    end
+
     def delete(asset, **_args)
       target_path = target_path_for(asset)
       File.delete(target_path)
@@ -48,6 +55,10 @@ class S3Storage
     def target_path_for(asset)
       relative_path = relative_path_for(asset)
       @target_root.join(relative_path)
+    end
+
+    def download_path_for(asset)
+      Tempfile.new(asset.uuid).path
     end
 
     def relative_path_for(asset)
--- a/lib/tasks/assets.rake
+++ b/lib/tasks/assets.rake
@@ -119,6 +119,21 @@ namespace :assets do
         end
       end
     end
+
+    desc "Scan a batch of assets for SVG vulnerabilites that haven't yet been scanned"
+    task :scan_next_svg_batch, %i[batch_size] => :environment do |_t, args|
+      if SvgScanBatchJob.jobs.empty?
+        Rails.logger.info("Enqueuing next SVG scan batch")
+        puts "Enqueuing next SVG scan batch"
+        batch_size = args.fetch(:batch_size)
+        Asset.where(state: "uploaded", deleted_at: nil, redirect_url: nil, svg_scanned_at: nil)
+          .limit(batch_size)
+          .each(&:schedule_svg_batch_scan)
+      else
+        Rails.logger.info("Not enqueuing next SVG scan batch: previous batch still in progress")
+        puts "Not enqueuing next SVG scan batch: previous batch still in progress"
+      end
+    end
   end
 end
 
--- a/spec/lib/s3_storage_spec.rb
+++ b/spec/lib/s3_storage_spec.rb
@@ -157,6 +157,15 @@ RSpec.describe S3Storage do
     end
   end
 
+  describe "#download" do
+    it "downloads the file from the S3 bucket" do
+      allow(s3_client).to receive(:get_object)
+      file = storage.download(asset)
+      expect(s3_client).to have_received(:get_object)
+      expect(File).to exist(file.path)
+    end
+  end
+
   describe "#delete" do
     it "deletes the file from the S3 bucket" do
       allow(s3_object).to receive(:delete).and_return(true)
--- a/spec/lib/tasks/assets_spec.rb
+++ b/spec/lib/tasks/assets_spec.rb
@@ -399,6 +399,34 @@ RSpec.describe "assets.rake" do
         task.invoke(filepath)
         expect(File.read(filepath)).to eq "6592008029c8c3e4dc76256c,DONE\n"
       end
+
+      describe "assets:bulk_fix:scan_next_svg_batch" do
+        let(:task) { Rake::Task["assets:bulk_fix:scan_next_svg_batch"] }
+
+        Sidekiq::Testing.fake! do
+          it "does nothing if the batch queue is not empty" do
+            SvgScanBatchJob.perform_async("a")
+            allow(Rails.logger).to receive(:info).at_least(:once)
+            task.invoke(1)
+            expect(Rails.logger).to have_received(:info).with("Not enqueuing next SVG scan batch: previous batch still in progress").once
+          end
+
+          it "schedules the next batch if the batch queue is empty" do
+            allow(Rails.logger).to receive(:info).at_least(:once)
+            task.invoke(1)
+            expect(Rails.logger).to have_received(:info).with("Enqueuing next SVG scan batch").once
+          end
+
+          it "limits the batch size to the given number" do
+            FactoryBot.create(:uploaded_asset)
+            FactoryBot.create(:uploaded_asset)
+            allow(Rails.logger).to receive(:info).at_least(:once)
+            task.invoke(1)
+            expect(Rails.logger).to have_received(:info).with("Enqueuing next SVG scan batch").once
+            expect(SvgScanBatchJob.jobs.size).to eq(1)
+          end
+        end
+      end
     end
   end
   # rubocop:enable RSpec/AnyInstance
--- a/spec/sidekiq/svg_scan_batch_job_spec.rb
+++ b/spec/sidekiq/svg_scan_batch_job_spec.rb
@@ -84,7 +84,7 @@ RSpec.describe SvgScanBatchJob do
 
       before do
         allow(scanner).to receive(:scan).and_raise(exception)
-        allow(Rails.logger).to receive(:info).at_least(:once)
+        allow(Rails.logger).to receive(:warn).at_least(:once)
         allow(GovukError).to receive(:notify).at_least(:once)
       end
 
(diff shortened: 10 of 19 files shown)


lib/s3_storage/fake.rb
```diff
@@ -12,6 +12,13 @@ def upload(asset, **_args)
       File.write(target_path, File.read(source_path))
     end
 
+    def download(asset, **_args)
+      source_path = source_path_for(asset)
+      download_path = download_path_for(asset)
+      FileUtils.mkdir_p(File.dirname(download_path))
+      File.write(download_path, File.read(source_path))
+    end
```
yndajas, line 20: **question:** do we need tests for the two new methods in this file?

spec/lib/s3_storage_spec.rb
```diff
@@ -157,6 +157,14 @@
     end
   end
 
+  describe "#download" do
+    it "downloads the file from the S3 bucket" do
+      allow(s3_client).to receive(:get_object)
+      storage.download(asset)
+      expect(s3_client).to have_received(:get_object)
```
yndajas, line 164: **note:** looks like this is following a pattern from other tests, but the `allow` implicitly `expect`s, so the last line is kind of redundant

spec/lib/s3_storage_spec.rb
```diff
@@ -157,6 +157,14 @@
     end
   end
 
+  describe "#download" do
+    it "downloads the file from the S3 bucket" do
```
yndajas, line 161: **question:** would it be too challenging to test that the tempfile changes as a result of this method? It might also be good to assert on the return value

lib/s3_storage/fake.rb
yndajas, line 20: **question:** the commit message mentions the `S3Storage` change but not the changes in this file. What are they for? (Can it be explained in the commit message?)

app/models/asset.rb
```diff
@@ -53,6 +53,10 @@ class Asset
```
yndajas, line 53: **question:** I would assume the system the app is running on rather than the worker?

> assumes that the asset exists as a file on the worker

yndajas, line 53: **suggestion:** 

> So a new job is needed

I'm not sure the "So" quite follows. I suppose the context is "Add a job to scan SVG assets in bulk", but it would be good to explain why we need this

yndajas, line 53: **suggestion:** 

> yet-to-be-scanned assets can easily be retrieved, and ones that were found to have vulnerabilities can be dealt with.

->

> we can easily _identify_ yet-to-be-scanned assets and ones that were found to have vulnerabilities.

app/sidekiq/concerns/ensure_file.rb
```diff
@@ -5,6 +5,12 @@ def ensure_file_is_same_after_scan(asset, job_name, next_state)
     initial_digest = asset.md5_hexdigest
     Rails.logger.info("#{asset.id} - #{job_name}#perform - scan started")
     yield
-    asset.reload.md5_hexdigest == initial_digest ? asset.send(next_state) : Rails.logger.info("#{asset.id} #{job_name} checksum failed")
+    if asset.reload.md5_hexdigest == initial_digest
+      asset.send(next_state)
+      asset.set(svg_scanned_safe: true)
+      asset.set(svg_scanned_at: Time.zone.now)
```
yndajas: [still relevant but note reply below]

**issue:** this and the scan started log make this method feel like a bit of a single responsibility principle violation, or at least a little misleadingly named. If its responsibility is ensuring a file is unchanged after a scan, should it really have these side effects?

I wonder if this could be revised to something like this:

```rb
# app/sidekiq/concerns/ensure_file.rb

def ensure_file_is_unchanged_by_job(asset, job_name, success_callback, &block)
  initial_digest = asset.md5_hexdigest
  block.call
  success_callback.call if asset.reload.md5_hexdigest == initial_digest
end

# app/sidekiq/svg_scan_job.rb

def success_callback(asset)
  asset.svg_scanned_clean!
  asset.set(svg_scanned_safe: true)
  # probably not the following line since it's done elsewhere in the job already
  # asset.set(svg_scanned_at: Time.zone.now)
end

[...]

ensure_file_is_unchanged_by_job(asset, "SvgScanJob", method(:success_callback)) do
  Rails.logger.info("#{asset.id} - #{job_name}#perform - scan started")
  Services.svg_scanner.scan(asset.file.path)
end
```

Or alternatively, the `ensure...` method could just return `true` or `false`, then you use that in the job to fire what are currently side effects?

app/sidekiq/svg_scan_batch_job.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    begin
+      file = Services.cloud_storage.download(asset)
+    rescue Aws::S3::Errors::NoSuchKey
+      # File doesn't exist in S3. Nothing to do.
+      return
+    end
+
+    begin
+      if Marcel::MimeType.for(Pathname.new(file.path)) == "image/svg+xml"
```
yndajas: **question:** would it be useful to store something on the asset to indicate whether it's an SVG, so we know not to include it in future bulk scans? It would need resetting when the file changes, maybe in https://github.com/alphagov/asset-manager/blob/07ec080fd6497f35b69f67f59772cd2f0e9d3b00/app/models/asset.rb#L257-L261 (called in https://github.com/alphagov/asset-manager/blob/07ec080fd6497f35b69f67f59772cd2f0e9d3b00/app/models/asset.rb#L145-L151)

app/sidekiq/svg_scan_batch_job.rb
```diff
@@ -0,0 +1,38 @@
...
+        begin
+          Rails.logger.info("#{asset_id} - SvgScanBatchJob#perform - SVG scan started")
+          Services.svg_scanner.scan(file.path)
+          asset.set(svg_scanned_safe: true)
```
yndajas: **question:** do we need to care about the potential for assets to be updated (with a new file) mid-scan here too?

app/sidekiq/svg_scan_job.rb
```diff
@@ -17,7 +17,9 @@ def perform(asset_id)
       rescue SvgDocument::UnsafeSvg
         Rails.logger.warn("#{asset_id} - SvgScanJob#perform - File #{asset.filename} marked as unsafe")
         asset.scanned_infected!
+        asset.set(svg_scanned_safe: false)
       end
+      asset.set(svg_scanned_at: Time.zone.now)
```
yndajas: **note:** you also set the scanned at field in the `ensure...` method, but I'd favour doing it here (see comment on that file)

app/sidekiq/svg_scan_batch_job.rb
yndajas: **question:** is this outer `begin` necessary? Is `Marcel::MimeType.for` likely to error? Where does the `ensure` need to go?

app/models/asset.rb
```diff
@@ -53,6 +53,10 @@ class Asset
 
   field :deleted_at, type: Time
 
+  field :svg_scanned_at, type: Time
+
+  field :svg_scanned_safe, type: Boolean
```
yndajas, line 58: **suggestion:** I'd probably split this second commit into two to keep it small:
1. the model and current job changes
2. the new job

config/sidekiq.yml
```diff
@@ -3,3 +3,4 @@
 :queues:
   - default
   - low_priority
+  - batch
```
yndajas, line 6: **suggestion:** could this be more specific, like `bulk/batch_svg_scan`?

app/sidekiq/concerns/ensure_file.rb
yndajas: Actually I reckon the whole success callback/side effects should go in the event itself (but still be duplicated in the batch job since it doesn't use the state machine): https://github.com/alphagov/asset-manager/blob/07ec080fd6497f35b69f67f59772cd2f0e9d3b00/app/models/asset.rb#L105-L107

I still think the scan started log belongs in the job though (or even the scanner service)

spec/sidekiq/svg_scan_batch_job_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    SidekiqUniqueJobs.use_config(enabled: true) do
+      expect { described_class.perform_in(1.minute, asset.id.to_s) }.to enqueue_sidekiq_job(described_class)
+      expect { described_class.perform_in(1.minute, asset.id.to_s) }.not_to enqueue_sidekiq_job(described_class)
+    end
+  end
+
+  it "does nothing unless the mimetype is 'image/svg+xml'" do
+    allow(Marcel::MimeType).to receive(:for).and_return("not-svg")
+    expect(scanner).not_to receive(:scan)
+  end
```
yndajas: **issue:** I think you need to call `worker.perform(asset.id)` here?

spec/sidekiq/svg_scan_batch_job_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+
+    it "calls out to the SvgScanner to scan the file" do
+      expect(scanner).to receive(:scan).with(file.path)
+      worker.perform(asset.id)
+    end
+
+    context "when the file is clean" do
+      before do
+        allow(scanner).to receive(:scan).and_return(true)
```
yndajas: **note:** it's interesting that we don't do anything with this return value 👀

spec/sidekiq/svg_scan_batch_job_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+
+    context "when the SVG asset is unsafe" do
+      let(:exception_message) { "SVG: Unsafe element detected: <script>" }
+      let(:exception) { SvgDocument::UnsafeSvg.new(exception_message) }
+
+      before do
+        allow(scanner).to receive(:scan).and_raise(exception)
+        allow(Rails.logger).to receive(:info).at_least(:once)
+        allow(GovukError).to receive(:notify).at_least(:once)
```
yndajas: **question:** I _think_ `at_least(:once)` is redundant. If it doesn't receive the message, the test should error out. You could use just `once` if you want to confirm it's _only_ once

app/sidekiq/svg_scan_batch_job.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def perform(asset_id)
+    asset = Asset.find(asset_id)
+
+    begin
+      file = Services.cloud_storage.download(asset)
+    rescue Aws::S3::Errors::NoSuchKey
+      # File doesn't exist in S3. Nothing to do.
+      return
+    end
```
yndajas: **suggestion:** it might be good to include a test for this?

lib/tasks/assets.rake
```diff
@@ -119,6 +119,21 @@ namespace :assets do
         end
       end
     end
+
+    desc "Scan a batch of assets for SVG vulnerabilites that haven't yet been scanned"
+    task :scan_next_svg_batch, %i[batch_size] => :environment do |_t, args|
+      # rubocop:disable Style/ZeroLengthPredicate
+      if SvgScanBatchJob.jobs.size.zero?
+        # rubocop:enable Style/ZeroLengthPredicate
```
yndajas: **question:** did RuboCop have a suggestion here, like `empty?`? I'd go with that if so

lib/tasks/assets.rake
```diff
@@ -119,6 +119,21 @@ namespace :assets do
...
+        Rails.logger.info("Enqueuing next SVG scan batch")
+        puts "Enqueuing next SVG scan batch"
```
yndajas, line 127: **todo:** looks like you have duplicate logging?

lib/tasks/assets.rake
```diff
@@ -119,6 +119,21 @@ namespace :assets do
...
+        batch_size = args.fetch(:batch_size)
+        Asset.where(state: "uploaded", deleted_at: nil, redirect_url: nil, svg_scanned_at: nil).limit(batch_size).each(&:schedule_svg_batch_scan)
```
yndajas: **nitpick:** might be nice to multi-line this chain!

spec/lib/tasks/assets_spec.rb
```diff
@@ -399,6 +399,34 @@
...
+          it "schedules the next batch if the batch queue is empty" do
+            allow(Rails.logger).to receive(:info).at_least(:once)
+            task.invoke(1)
+            expect(Rails.logger).to have_received(:info).with("Enqueuing next SVG scan batch").once
+          end
+
+          it "limits the batch size to the given number" do
+            FactoryBot.create(:uploaded_asset)
+            FactoryBot.create(:uploaded_asset)
+            allow(Rails.logger).to receive(:info).at_least(:once)
```
yndajas, line 423: **question:** are the allows needed? It doesn't look like a stub

spec/sidekiq/svg_scan_batch_job_spec.rb
```diff
@@ -0,0 +1,93 @@
+require "rails_helper"
+require "services"
+require "sidekiq_unique_jobs/testing"
+
+RSpec.describe SvgScanBatchJob do
+  let(:worker) { described_class.new }
+  let(:asset) { FactoryBot.create(:svg_asset_safe) }
```
yndajas, line 7: **thought:** not really related to this PR, but I wonder if the factory names could be clearer. It's a little hard - without reading the factory code - to work out that the difference between `svg_asset_safe` and `svg_asset_clean` is which scan (virus vs SVG) has been successful. Here I was confused that we start a job that scans SVGs with a 'safe' asset

app/models/asset.rb
yndajas: ... this is in the next commit message, but I think the context would be useful in the earlier one

yndajas (commented):
Generally looks like a sensible approach. A few comments/suggestions on the details

lib/tasks/assets.rake
nacnudus: It suggested both `.size.zero?` and `.empty?`, but `.empty?` doesn't work.

lib/s3_storage/fake.rb
nacnudus: I actually think the whole spec of this file is overkill. Not every existing method is tested anyway. An example of one that is tested:

```
def exists?(asset)
  target_path = target_path_for(asset)
  File.exist?(target_path)
end
```

```
describe "#exists?" do                                       
  let(:asset_path) { root_path.join(relative_path_to_asset) }
                                                             
  context "when file is not saved in storage" do             
    it "returns falsey" do                                   
      expect(storage).not_to exist(asset)                    
    end                                                      
  end                                                        
                                                             
  context "when file is saved in storage" do                 
    before do                                                
      storage.upload(asset)                                  
    end                                                      
                                                             
    it "returns truthy" do                                   
      expect(storage).to exist(asset)                        
    end                                                      
  end                                                        
end                                                          
```

spec/lib/s3_storage_spec.rb
nacnudus: I looked this up, and it seems that `expect` impllies `allow`, rather than the other way around. Given the fairly widespread confusion, I think it's worth following the pattern of being explicit rather than implicit.

spec/lib/s3_storage_spec.rb
nacnudus: I've added an expectation of the existence of the the file that `download` returns.

lib/s3_storage/fake.rb
nacnudus: I've added to the commit message
> fake.rb is updated to support the new method, so that it can be used in tests.

app/models/asset.rb
nacnudus: I've rewritten the commit message to
> assumes that the asset can be assessed as a local file on the system that the app is running on, because it hasn't yet been uploaded to S3.

lib/tasks/assets.rake
nacnudus: I wasn't sure what to do here. The rake tasks generally use `puts`, presumbably so that someone running the task in a console can read the messages there. But I also put in the logs, because our immediate use of this task will be scheduled, not interactive.

lib/tasks/assets.rake
nacnudus: Rubocop looked the other way 🙈

spec/lib/tasks/assets_spec.rb
nacnudus: See my reply to a similar comment. `allow` vs `expect` seems to cause confusion, whereas this is explicit, and there are other instances already in the code.

spec/sidekiq/svg_scan_batch_job_spec.rb
nacnudus: I agree. Shall I rename them in this PR or separately?

config/sidekiq.yml
nacnudus: I considered that, but the more generic name would allow it to be used for a batch virus scan too, should we decided rescan everything with the latest virus definitions. I'm also reconsidering the name `lowest_priority`, which is what it is after all.

A specific name would be useful if it were configured to have the same priority as another queue, so that the purpose of separating the queues is for some other reason than priority.

spec/sidekiq/svg_scan_batch_job_spec.rb
nacnudus: Without `.at_least(:once)`, it defaults to `.once`. I suspect the reason why the virus scanner spec uses `.at_least(:once)` is so that the test wouldn't fail were more logs to be introduced.

lib/tasks/assets.rake
nacnudus: Now it works 🤷

app/sidekiq/svg_scan_batch_job.rb
nacnudus: They would be scanned before upload. Unless I've missed something?

app/models/asset.rb
yndajas: Accessed?

app/sidekiq/svg_scan_batch_job.rb
nacnudus: Actually it's possible for a safe replacement file to be marked as unsafe.

1. Scanning of an original, unsafe file begins.
2. A replacement file is scanned, found to be safe, and is uploaded.
3. Scanning of the original file is completed, and finds it to be unsafe.

This seems very unlikely, and we would probably discover it when dealing with the files marked as unsafe.

yndajas: I suppose it would be covered by the normal `SvgScanJob`

Say:
- an `SvgScanBatchJob` starts
- you upload a bad replacement file, queueing an `SvgScanJob`
- the `SvgScanJob` somehow completes first, marking the result as bad
- the `SvgScanBatchJob` on the old file finishes second, marking the asset as good

That seems very, very unlikely, and I think even if it happened, the failed `SvgScanJob` would prevent the bad file from uploading, so the results of the `SvgScanBatchJob` would effectively be accurate?

If the `SvgScanJob` finishes second with a failure though, would the metadata reflect a file that failed to replace the one in S3 that we return in public requests?

app/sidekiq/concerns/ensure_file.rb
nacnudus: I've started work on this in https://github.com/alphagov/asset-manager/pull/1991/changes/f9c841d559f8e1722502d18422a7619e02071364. If we go ahead with it, I think most of the refactoring should be done on just the virus scanning, in an initial commit, and then the subsequent SVG commits should be rewritten to include these changes.

It's quite a lot of change to code that was written very recently, so we might consider whether it's worthwhile.

I've left the scan-started log in the job, because the scanner service doesn't know the asset ID, and I don't think it ought to know anything except the file path.

spec/sidekiq/svg_scan_batch_job_spec.rb
yndajas: I don't think you actually need to `allow` these at all, since they aren't set up as test doubles/stubbed. The tests pass with the two `allow(Rails.logger).to receive(:info).at_least(:once)` lines removed

lib/tasks/assets.rake
yndajas: I'd have assumed `Rails.logger.info` logs would show when running in a console (making the `puts` redundant), but it would be worth checking

spec/lib/tasks/assets_spec.rb
yndajas: Yeah per https://github.com/alphagov/asset-manager/pull/1991#discussion_r3626084342 I'd avoid this unless it's a test double. I don't think you'd usually `allow` on non-double unless you want to stub a specific response

app/sidekiq/svg_scan_batch_job.rb
yndajas: Ah, didn't see your recent comment before replying. Agree that this is probably a rare and relatively inconsequential edge case

app/models/asset.rb
```diff
@@ -241,6 +245,10 @@ def schedule_svg_scan
     Marcel::MimeType.for(Pathname.new(file.path)) == "image/svg+xml" ? SvgScanJob.perform_async(id.to_s) : svg_scan_skipped!
   end
 
+  def schedule_svg_batch_scan
+    SvgScanBatchJob.perform_async(id.to_s)
+  end
```
yndajas, line 265: **todo:** this belongs in the next commit

app/sidekiq/svg_scan_job.rb
yndajas: **todo:** looks like this still needs deduplicating pending decision on whether to reduce the scope of `EnsureFile`

https://github.com/alphagov/asset-manager/pull/1991#discussion_r3586448728

spec/sidekiq/svg_scan_batch_job_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+      context = Seahorse::Client::RequestContext.new
+      error = Aws::S3::Errors::NoSuchKey.new(context, "The specified key does not exist.")
+      allow(storage).to receive(:download).and_raise(error)
+      allow(Rails.logger).to receive(:info).at_least(:once)
+    end
+
+    it "does nothing" do
+      worker.perform(asset.id)
+      expect(Rails.logger).to have_received(:info).with("#{asset.id} - SvgScanBatchJob#perform - Asset missing from S3").once
+    end
```
yndajas, line 37: **suggestion:** perhaps worth more explicitly saying and testing that it doesn't call `scan`, like below?

`expect(scanner).not_to receive(:scan)`

Probably in a separate test to the one expecting on `Rails.logger`?

spec/sidekiq/svg_scan_batch_job_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+
+  context "when the file's mimetype is not 'image/svg+xml'" do
+    before do
+      allow(Marcel::MimeType).to receive(:for).and_return("not-svg")
+    end
+
+    it "does nothing" do
+      expect(scanner).not_to receive(:scan)
+      worker.perform(asset.id)
+    end
```
yndajas, line 48: **suggestion:** `it "does not call out to the SvgScanner to scan the file"`

lib/tasks/assets.rake
```diff
@@ -119,6 +119,21 @@ namespace :assets do
         end
       end
     end
+
+    desc "Scan a batch of assets for SVG vulnerabilites that haven't yet been scanned"
+    task :scan_next_svg_batch, %i[batch_size] => :environment do |_t, args|
+      if SvgScanBatchJob.jobs.empty?
+        Rails.logger.info("Enqueuing next SVG scan batch")
+        puts "Enqueuing next SVG scan batch"
+        batch_size = args.fetch(:batch_size)
```
yndajas, line 128: **question:** does this need a `.to_i`?

spec/sidekiq/svg_scan_batch_job_spec.rb
yndajas: Same should apply in the Rake task tests

spec/lib/tasks/assets_spec.rb
```diff
@@ -399,6 +399,34 @@
...
+            SvgScanBatchJob.perform_async("a")
+            allow(Rails.logger).to receive(:info).at_least(:once)
+            task.invoke(1)
+            expect(Rails.logger).to have_received(:info).with("Not enqueuing next SVG scan batch: previous batch still in progress").once
+          end
+
+          it "schedules the next batch if the batch queue is empty" do
+            allow(Rails.logger).to receive(:info).at_least(:once)
+            task.invoke(1)
+            expect(Rails.logger).to have_received(:info).with("Enqueuing next SVG scan batch").once
```
yndajas, line 417: **suggestion:** could be good to assert something like

`expect_any_instance_of(Asset).to receive(:schedule_svg_batch_scan)`

and

`expect_any_instance_of(Asset).not_to receive(:schedule_svg_batch_scan)`

app/sidekiq/concerns/ensure_file.rb
yndajas: I hadn't really expected this to require changes to the virus scan flow. I guess that's because of the shared `scanned_infected!` event

I think a simpler approach of just moving the `asset.set(svg_scanned_safe: true)` call into `event :svg_scanned_clean` and letting the job set the `asset.set(svg_scanned_at: Time.zone.now)` could work?

```rb
# app/sidekiq/svg_scan_job.rb
begin
  ...
  ensure_file_is_same_after_scan(asset, "SvgScanJob", :svg_scanned_clean!) do
    # provided there's no error and the file is unchanged, this calls
    # svg_scanned_clean!
    Services.svg_scanner.scan(asset.file.path)
  end
rescue SvgDocument::UnsafeSvg
   ...
  asset.scanned_infected!
  # you could instead introduce an `:svg_scanned_infected` event which performs
  # the line below before calling `scanned_infected!`. Then you could move the
  # timestamp setting line into the two events
  asset.set(svg_scanned_safe: false)
end
asset.set(svg_scanned_at: Time.zone.now)

# app/models/asset.rb
event :svg_scanned_clean do
  asset.set(svg_scanned_safe: true)
  transition virus_scanned_clean: :clean
end
```

---

Alternatively, if it's easier than tinkering with the state machine, you could 'just' replace the `next_state` parameter of `ensure_file_is_same_after_scan` with a callback and pass in e.g.

```
def scanned_clean_callback
  asset.set(svg_scanned_safe: true)
  asset.svg_scanned_clean!
end

# def ensure_file_is_same_after_scan(asset, job_name, success_callback)
ensure_file_is_same_after_scan(asset, "SvgScanJob", -> scanned_clean_callback)
```

I think it's cleaner in the state machine, but the job at least feels a better place to specify the side effects than the `EnsureFile` concern

---

All that said, I don't know that it's worth spending too much time on this. We can always refactor later!

app/sidekiq/svg_scan_batch_job.rb
yndajas: This is still an open question but it's non-blocking

app/sidekiq/svg_scan_job.rb
yndajas: And this should probably be fixed pending the outcome of the SRP/refactor discussion

spec/sidekiq/svg_scan_batch_job_spec.rb
yndajas: Per our chat, a final commit would be fine if it's not too much trouble, but otherwise separately

app/sidekiq/svg_scan_batch_job.rb
yndajas: This might actually be blocking:
- the Rake task will schedule say 1000 jobs
- the jobs will do nothing if the files are not SVGs
- the next run will schedule the same 1000 jobs...

PR #1996, opened 2026-07-15
Handle formatless requests with a 404

callumknights:
This is a reasonably quick fix for an issue that has cropped up since Rails 8.1, where formatless requests, which were intended to be handled as `NullType`, we're actually throwing 500's as `NullType` is not a registered MimeType, so subsequently `requests.formats` fails a lookup deeper in the rails code, then defaulting to `nil`. This `nil` then causes template rendering to fail, resulting in a 500 error.

The decision was made that given there are basically no assets without a format extension, and all the traffic causing these 500's since the upgrade are not legitimate users, that having this throw a 404 is more sensible response.

### DIFF
--- a/app/controllers/whitehall_media_controller.rb
+++ b/app/controllers/whitehall_media_controller.rb
@@ -1,6 +1,6 @@
 class WhitehallMediaController < MediaController
   def download
-    request.formats = [Mime::NullType] if params[:format].blank?
+    error_404 if params[:format].blank?
     super
   end
 
--- a/spec/controllers/whitehall_media_controller_spec.rb
+++ b/spec/controllers/whitehall_media_controller_spec.rb
@@ -67,7 +67,7 @@ RSpec.describe WhitehallMediaController, type: :controller do
 
           expect(request.fullpath).to eq(legacy_url_path)
           expect(request.query_parameters).to be_empty
-          expect(request.format).to be_an_instance_of(Mime::NullType)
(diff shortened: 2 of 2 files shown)


app/controllers/whitehall_media_controller.rb
```diff
@@ -1,6 +1,6 @@
 class WhitehallMediaController < MediaController
   def download
-    request.formats = [Mime::NullType] if params[:format].blank?
+    error_404 if params[:format].blank?
```
brucebolt, line 3: I think my idea was more to check if the asset actually exists first, before serving the 404 error.

However if you've confirmed there are no `WhitehallAsset` records without a format in the `legacy_url_path`, then I'd be happy to approve as is (since these cannot be created, except through one rake task).

callumknights: In which case there might be a better way to do this. I had a different idea while implementing this, and that might support the check you've asked for, I'll have a look 👍

brucebolt: I was thinking more just logging into a Rails console and running a one-off query to see if there's any WhitehallAssets without an extension.

callumknights: @nacnudus Kindly provided me with this info. We have 8 Assets overall with no extension, none of those have a legacy_url_path. So this should be fine.

brucebolt (approved):

PR #2000, opened 2026-07-20
[PP-7782] Add 'Payload Too Large' response for file uploads

brucebolt:
Currently, files larger than the ClamAV scanning limit may be accepted by the API but cannot be processed or scanned correctly downstream. This can lead to inconsistent behaviour, failed virus scans, and users needing to contact us via Zendesk.

This updates Asset Manager file upload endpoints to return an HTTP 413 Payload Too Large response when an uploaded file exceeds the maximum allowed file size configured in ClamAV (`MaxFileSize`).

This will allow publishers to get immediate feedback that a file is too large, rather than waiting for the virus scan to run.

Depends on https://github.com/alphagov/govuk-helm-charts/pull/4364.

Note: this does not expand ZIP files to check whether the `MaxScanSize` is exceeded. It would be very difficult to make this check, without reverse engineering how ClamAV works through documents.

### DIFF
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -1,5 +1,8 @@
 class AssetsController < ApplicationController
+  MAX_FILE_SIZE = Rails.configuration.max_file_size
+
   before_action :restrict_request_format
+  before_action :check_request_size, only: %i[create update]
 
   def show
     @asset = find_asset(include_deleted: true)
@@ -81,5 +84,11 @@ private
 
   def build_asset
     Asset.new(asset_params)
+  end
+
+  def check_request_size
+    if request.content_length.to_i > MAX_FILE_SIZE
+      head :payload_too_large
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


app/controllers/assets_controller.rb
```diff
@@ -82,4 +85,10 @@ def find_asset(include_deleted: false)
   def build_asset
     Asset.new(asset_params)
   end
+
+  def check_request_size
+    if request.content_length.to_i > MAX_FILE_SIZE
```
yndajas, line 90: **question:** is it definitely the `Content-Length` header we need to check or is it something like `params.dig(:asset, :file)`? Is something setting that header to reflect only the file rather than the full request body?

brucebolt: I went down this route for simplicity of implementation and testing, on the basis that the remainder of the request would just be a few kilobytes, meaning the limit wouldn't be substantially lower than 750MB.

I'll add this to the commit message.

yndajas: Seems reasonable to me

yndajas (approved):

PR #2012, opened 2026-07-28
[PP-7886] Allow `auth_bypass_ids` to expire

brucebolt:
In HMRC Manuals API, we are currently building functionality to allow upload of assets (https://github.com/alphagov/hmrc-manuals-api/pull/1316).  Part of this requires draft asset's `auth_bypass_ids` to expire after 30 days.

This PR adds that functionality into Asset Manager, so publishing applications can specify this value.

There is no change for assets that do not have an expiry date specified.

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -137,7 +137,7 @@ class Asset
   end
 
   def valid_auth_bypass_token?(auth_bypass_id)
-    auth_bypass_ids.include?(auth_bypass_id)
+    auth_bypass_ids.include?(auth_bypass_id) && (auth_bypass_ids_expiry.nil? || auth_bypass_ids_expiry > Time.zone.now)
   end
 
   def public_url_path
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -49,6 +49,8 @@ class Asset
 
   field :auth_bypass_ids, type: Array, default: []
 
+  field :auth_bypass_ids_expiry, type: Time
+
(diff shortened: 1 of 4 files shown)


app/models/asset.rb
```diff
@@ -49,6 +49,8 @@ class Asset
 
   field :auth_bypass_ids, type: Array, default: []
 
+  field :auth_bypass_ids_expiry, type: Time
```
callumknights, line 52: Am I correct in saying this means that all auth_bypass_ids for a given asset will have the same global expiry time? Is this the right approach?

brucebolt: Yes, all will have the same expiry time.

It'd probably be better to have a separate time for each ID, but that would need a lot more work (including in every publishing app, as the value would need to be a hash instead of an array, to associate the expiry with the token).

In HMRC Manuals API's use case, we'll only ever have one valid ID, so it should be fine to expire them all at the same time.

callumknights: Fantastic! Thanks for explaining.

callumknights (approved):

yndajas (approved):

app/models/asset.rb
yndajas: Is it worth noting this in the commit message in case our future selves question why they all have the same expiry?

brucebolt: I've updated the commit message.

PR #2013, opened 2026-07-29
[PP-7805] Refactor the asset state machine and its tests

yndajas:
This is groundwork for bulk-scanning previously-uploaded SVGs. It splits the `scanned_infected` event into separate virus and SVG events, the first step towards a later change of moving scan side effects out of the `EnsureFile` concern and into the jobs. It also orders the state machine events by their from-state, and restructures and colocates the scanning and state machine tests so the coverage is easier to follow

First of five stacked PRs reworking #1991

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -110,11 +110,11 @@ class Asset
       SaveToCloudStorageJob.perform_async(asset.id.to_s)
     end
 
-    event :scanned_infected do
+    event :virus_scanned_infected do
       transition unscanned: :infected
     end
 
-    event :scanned_infected do
+    event :svg_scanned_infected do
       transition virus_scanned_clean: :infected
     end
 
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -90,6 +90,10 @@ class Asset
       block.call
     end
 
+    event :virus_scanned_infected do
+      transition unscanned: :infected
+    end
+
     event :virus_scanned_clean do
       transition unscanned: :virus_scanned_clean
     end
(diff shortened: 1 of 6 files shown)


spec/models/asset_spec.rb
```diff
@@ -470,6 +470,340 @@
...
+        asset.virus_scanned_clean!
+      end
+
+      it "marks the asset as clean" do
+        asset.virus_scanned_clean!
+
+        expect(asset).to be_clean
+      end
+
+      context "but the file is an SVG" do
```
callumknights: I personally find the "but the file is/has foo" context's in this a bit difficult to follow. They match the flow of the state machine well, which perhaps why they are difficult to follow. However I'm wondering it it might be easier to follow with separate test scenarios for when the file is/is not an SVG. But conversely, that doesn't follow the state machine flow as naturally, and will probably mean more test duplication. I don't have a strong enough opinion about this to make it blocking, so will leave as a comment.

app/models/asset.rb
```diff
@@ -98,24 +102,20 @@ class Asset
...
-
-    event :virus_scanned_infected do
-      transition unscanned: :infected
+    event :svg_scan_skipped do
+      transition virus_scanned_clean: :clean
     end
 
-    event :svg_scanned_infected do
-      transition virus_scanned_clean: :infected
+    after_transition to: :clean do |asset, _|
```
callumknights, line 117: I was ambivalent about this until I saw the `SaveToCloudStorageJob` call at the bottom of the state machine. This is much better 👏

spec/models/asset_spec.rb
yndajas: I think you're right - would be better to wrap the other tests in these contexts in `context "and the file is not an SVG"` then change these to `context "and the file is an SVG"`

callumknights: I think that works really well as well. Thanks!

callumknights (approved):
Tests are much easier to follow now! Thanks for that 👍

PR #2016, opened 2026-07-29
[PP-7805] Record asset MIME type and SVG scan details

yndajas:
This records the extra asset metadata that bulk SVG scanning relies on. It persists the file's MIME type and stores the time and result of an SVG scan on the asset, so that a later change can skip assets which have already been scanned or whose files aren't SVGs. It also fixes and adds tests for how an asset's stored file metadata is updated on save

Fourth of five stacked PRs reworking #1991

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -54,6 +54,9 @@ class Asset
 
   field :parent_document_url, type: String
 
+  field :svg_scanned_at, type: Time
+  field :svg_scan_state, type: String
+
   field :deleted_at, type: Time
 
   validates :file, presence: true, if: :unscanned?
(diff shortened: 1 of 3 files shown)


app/models/asset.rb
callumknights: I'm aware this isn't part of the diff, but feels relevant. Why isn't this using`size: size_from_file`? I appreciate I may be missing something here, this PR is generally making my head hurt a bit.

callumknights: I don't know if I did that as a green review by accident but I'm very much not done 🙈

yndajas: Yeah I don't see why it shouldn't, given the logic is the same and we use `last_modified_from_file` here

yndajas: I suspect the reason is that `size_from_file` came later, but I reckon we should update it

https://github.com/alphagov/asset-manager/commit/bfada1a534d441df1c6d999d498b5d4007a39ddb

app/models/asset.rb
```diff
@@ -54,6 +54,9 @@ class Asset
 
   field :parent_document_url, type: String
 
+  field :svg_scanned_at, type: Time
```
callumknights, line 57: This is great! And potentially a really good future pattern for re-scanning for viruses in the same way

callumknights (dismissed):
Took me a fair while to wrap my head around [632e4c3](https://github.com/alphagov/asset-manager/pull/2016/commits/632e4c3a68c459c7de8c9384584828cb80b0be72), but this generally looks really good 👍

app/models/asset.rb
yndajas: I've fixed this - we might as well include it given the assortment of changes in these changed PRs

callumknights (approved):

PR #2017, opened 2026-07-29
[PP-7805] Scan existing SVG assets in bulk

yndajas:
This adds bulk SVG scanning for assets uploaded before scanning existed. S3 storage gains a download method, a new batch job downloads the file from S3 and scans it without altering the asset's state, and a temporary rake task enqueues unscanned in-scope assets up to a given batch size. The job runs on a low-priority "batch" queue, and the task is intended to run on a schedule until the backlog is cleared

Fifth of five stacked PRs reworking #1991

Closes #1991

### DIFF
--- a/app/sidekiq/svg_scan_batch_job.rb
+++ b/app/sidekiq/svg_scan_batch_job.rb
@@ -1,0 +1,45 @@
+require "services"
+
+class SvgScanBatchJob
+  include Sidekiq::Job
+
+  sidekiq_options lock: :until_executing, queue: "batch"
+
+  def perform(asset_id)
+    asset = Asset.find(asset_id)
+    file = Services.cloud_storage.download(asset)
+
+    ensure_mime_type_set(asset, file.path)
+
+    scan_svg(asset, file.path) if asset.mime_type == "image/svg+xml"
+  rescue Aws::S3::Errors::NoSuchKey
+    asset.update!(svg_scan_state: "file_missing_from_s3")
+    Rails.logger.info("#{asset_id} - SvgScanBatchJob - Asset missing from S3")
+  ensure
+    if file.respond_to?(:path) && File.exist?(file.path)
+      file.close
+      file.unlink
+    end
+  end
+
+private
+
+  def ensure_mime_type_set(asset, file_path)
+    if asset.mime_type.nil?
+      asset.set(mime_type: Marcel::MimeType.for(Pathname.new(file_path)))
+      asset.reload
+    end
+  end
+
+  def scan_svg(asset, file_path)
+    Rails.logger.info("#{asset.id} - SvgScanBatchJob - SVG scan started")
+    Services.svg_scanner.scan(file_path)
+    asset.update!(svg_scan_state: "svg_clean")
+  rescue SvgDocument::UnsafeSvg => e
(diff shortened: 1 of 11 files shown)


lib/s3_storage/fake.rb
```diff
@@ -12,6 +12,13 @@ def upload(asset, **_args)
       File.write(target_path, File.read(source_path))
     end
 
+    def download(asset, **_args)
```
callumknights: I think the double splat here is unnecessary. It fits the pattern already there for the fake S3, but the only reason the fake S3 exists is for testing, and in the tests no more arguments other than `asset` are ever passed to `upload` or `delete`.

app/sidekiq/svg_scan_batch_job.rb
```diff
@@ -0,0 +1,45 @@
...
+      asset.set(mime_type: Marcel::MimeType.for(Pathname.new(file_path)))
+      asset.reload
+    end
+  end
+
+  def scan_svg(asset, file_path)
+    Rails.logger.info("#{asset.id} - SvgScanBatchJob - SVG scan started")
+    Services.svg_scanner.scan(file_path)
+    asset.update!(svg_scan_state: "svg_clean")
+  rescue SvgDocument::UnsafeSvg => e
```
callumknights, line 38: What's the plan for handling potential infected scans? I get not updating the state machine, as that would break the state machine flow as the asset is already uploaded, so updating the DB field and logging is the right approach, but what happens then? Is it just a developer action to then check the logs and decide what to do? Or should we play it safe and remove anything that fails from S3?

spec/lib/tasks/assets_spec.rb
```diff
@@ -30,6 +30,106 @@
...
+    context "when there are no in-scope assets to scan" do
+      before do
+        FactoryBot.create(:asset) # not uploaded
+        FactoryBot.create(:uploaded_asset, deleted_at: Time.zone.now) # deleted
+        FactoryBot.create(
+          :uploaded_asset,
+          redirect_url: "https://www.gov.uk/somewhere",
+        ) # redirected
+        FactoryBot.create(:uploaded_svg_asset) # scanned
+        FactoryBot.create(:uploaded_asset).set(mime_type: "application/pdf") # wrong MIME type
```
callumknights, line 57: Helpful comments 👍

app/sidekiq/svg_scan_batch_job.rb
yndajas: (moved conversation to our team channel in Slack)

lib/s3_storage/fake.rb
yndajas: I've haven't really taken the time to fully understand the fake S3 stuff, but I've removed this and everything still passes

yndajas: The real method does take an extra named argument, but we don't actually pass it in anywhere

app/sidekiq/svg_scan_batch_job.rb
yndajas: We've decided to manually review affected SVGs, assuming there aren't too many

callumknights (approved):

PR #2029, opened 2026-08-06
[PP-7924] Add User based restrictions to Asset Manager API endpoints

callumknights:
This is being added to support the integration of HMRC-Manuals-API with Asset Manager

Without these controls HMRC Manuals API would have full management rights with all Assets stored on Asset manager, so we thought it prudent to add restrictions to how publishing apps and API users can manage assets. 

This PR means only API users with the "Manage all Assets" permission in Signon can use #show, #update, #delete and #restore endpoints in asset manager. Otherwise use of those endpoints will be restricted to the original Api user. 

This permission has already been created in Signon, and granted to every Publishing App with an active API user for Asset Manager, so merging this PR should cause no issues.

### DIFF
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -14,6 +14,8 @@ class AssetsController < ApplicationController
   def create
     @asset = build_asset
 
+    @asset.user = current_user
+
     if @asset.save
       render json: AssetPresenter.new(@asset, view_context).as_json(status: :created), status: :created
     else
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -7,6 +7,8 @@ class AssetsController < ApplicationController
   def show
     @asset = find_asset(include_deleted: true)
 
+    return error_403 unless @asset.manageable_by?(current_user)
+
     expires_now
     render json: AssetPresenter.new(@asset, view_context)
   end
@@ -26,6 +28,8 @@ class AssetsController < ApplicationController
   def update
     @asset = Asset.undeleted.or(Asset.where(draft: true)).find(params[:id])
 
+    return error_403 unless @asset.manageable_by?(current_user)
+
     if @asset.update(asset_params)
       render json: AssetPresenter.new(@asset, view_context).as_json(status: :success)
     else
@@ -35,12 +39,18 @@ class AssetsController < ApplicationController
 
   def destroy
     @asset = find_asset
+
+    return error_403 unless @asset.manageable_by?(current_user)
+
     @asset.destroy!
     render json: AssetPresenter.new(@asset, view_context).as_json(status: :success)
   end
 
   def restore
     @asset = find_asset(include_deleted: true)
+
+    return error_403 unless @asset.manageable_by?(current_user)
+
     @asset.restore
     render json: AssetPresenter.new(@asset, view_context).as_json(status: :success)
   end
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -17,6 +17,7 @@ class Asset
   index deleted_at: 1
 
   belongs_to :replacement, class_name: "Asset", optional: true, index: true
+  belongs_to :user, optional: true
 
   field :state, type: String, default: "unscanned"
   field :filename_history, type: Array, default: -> { [] }
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -135,6 +135,10 @@ class Asset
     access_limited.include?(user.uid) || access_limited_organisation_ids.include?(user.organisation_content_id)
   end
 
+  def manageable_by?(user)
+    self.user == user || user.permissions.include?("Manage all Assets")
+  end
+
   def valid_auth_bypass_token?(auth_bypass_id)
     auth_bypass_ids.include?(auth_bypass_id) && (auth_bypass_ids_expiry.nil? || auth_bypass_ids_expiry > Time.zone.now)
   end
--- a/spec/controllers/assets_controller_spec.rb
+++ b/spec/controllers/assets_controller_spec.rb
@@ -5,12 +5,12 @@ RSpec.describe AssetsController, type: :controller do
 
   let(:file) { load_fixture_file("asset.png") }
 
-  before do
-    login_as_stub_user
-  end
-
   describe "POST create" do
     let(:valid_attributes) { { file: } }
+
+    before do
+      login_as_stub_user
+    end
 
     context "when attributes are valid" do
       it "persists asset" do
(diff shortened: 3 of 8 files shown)


app/controllers/assets_controller.rb
```diff
@@ -50,6 +50,8 @@ def restrict_request_format
   end
 
   def asset_params
+    params[:asset].merge!(user_uid: current_user.uid)
```
yndajas: **suggestion:** this could go in the `tap` block below

```rb
asset[:user_uid] = current_user.uid
```

spec/factories/factories.rb
```diff
@@ -53,5 +55,10 @@
   factory :user do
     sequence(:name) { |n| "Winston #{n}" }
     permissions { %w[signin] }
+    uid { "123456" }
+  end
+
+  factory :managing_user, parent: :user do
+    permissions { ["signin", "Manage all Assets"] }
```
yndajas, line 60: **issue:** this factory is introduced in the first commit, but I can't see that it's used?

app/models/asset.rb
```diff
@@ -26,6 +26,7 @@ class Asset
   index uuid: 1
   attr_readonly :uuid
 
+  field :user_uid, type: String
```
yndajas: **question:** would a `belongs_to :user` association be more appropriate?

spec/controllers/assets_controller_spec.rb
```diff
@@ -232,195 +232,239 @@
       let(:file) { load_fixture_file("asset2.jpg") }
       let(:valid_attributes) { { file: } }
 
-      it "persists new attributes on existing asset" do
-        put :update, params: { id: asset.id, asset: valid_attributes }
+      context "with a non-managing user" do
```
yndajas: **suggestion:** I'm a little lost in the spec diff on the second commit. It looks like the ownership/management-based permissions policy is tested as far as success/forbidden response, but the more specific tests just appear in one of the successful contexts? Is it worth using some shared examples to clean this up/test the same things in each successful context? Doing this in a preceding commit would likely make the diff easier to read

app/controllers/assets_controller.rb
yndajas: **question:** do we know that `current_user` is always set for API apps (i.e. across all CRUD endpoints)? I think my assumption was that Rails stores the current user in its sessions, and API-only apps wouldn't use sessions, but we do rely on it elsewhere 🤔

callumknights: In this instance we do because of https://github.com/alphagov/asset-manager/blob/main/app/controllers/application_controller.rb#L8

spec/factories/factories.rb
callumknights: That's fair I could drop this forward to this next commit

spec/controllers/assets_controller_spec.rb
callumknights: It might be yeah. I kept the permissions policy limited to success/forbidden cause forbidden is the only response we'd ever get for more specific scenarios, where the user is incorrect for the given scenario. However, on reflection it's made the tests messy as I got a bit carried away with my approach. I'll see if I can use some shared examples to tidy them up a bit.

spec/factories/factories.rb
yndajas: I couldn't see that it was used in the next commit either 👀

callumknights: https://github.com/alphagov/asset-manager/blob/5b54a3d0aba6c158fbc02922538d42062e775037/spec/support/authentication.rb#L26-L28

app/models/asset.rb
callumknights: It wasn't my first instinct cause of the addition of the "Manage all assets" permission effectively making all Assets belong to all Users with that permission. Which breaks the model a bit. I'd be happy to be proven wrong though.

app/controllers/assets_controller.rb
callumknights: Good spot, the nesting in the `require` threw me off

app/models/asset.rb
brucebolt: +1 for this using `belongs_to`, then Rails will add some magic methods like `.user` automatically, if we ever need them.

app/controllers/assets_controller.rb
brucebolt: Merging in the user ID here will mean we also add a user for update actions (not just create).  Is this definitely something we want to do, as I don't see a test for it?

app/controllers/assets_controller.rb
```diff
@@ -90,6 +100,10 @@ def build_asset
     Asset.new(asset_params)
   end
 
+  def can_manage_asset?
+    @asset.user_uid == current_user.uid || current_user.permissions.include?("Manage all Assets")
+  end
```
brucebolt: Would this be better moved onto the `Asset` model, in a similar way to how [the `accessible_by?` method works](https://github.com/alphagov/asset-manager/blob/2a31089b8dfb30dbee59b528b1871c5f908369cd/app/models/asset.rb#L131-L135)?

spec/controllers/assets_controller_spec.rb
```diff
@@ -232,195 +232,239 @@
...
+            expect(response).to have_http_status(:success)
+          end
+        end
 
-        expect(assigns(:asset).file.path).to match(/asset2\.jpg$/)
-      end
+        context "when the asset is not owned by the user" do
+          before do
+            asset.update!(user_uid: "000000")
+          end
```
brucebolt: Instead of updating the asset's user ID in each test, could the original `let` block for the asset be updated to take `user_uid` as a parameter (with a default value of nil), then you'd just need to add a `let(user_uid) { foo }` into each context (which will remove all the `before` blocks)?  It might reduce the number of lines of setup needed in each context.

app/models/asset.rb
callumknights: Okay fair, I'll get to work changing that round this morning

app/controllers/assets_controller.rb
callumknights: Changing all this to a `belongs_to` association probably makes this a natural refactor anyway, I'll add it in

app/controllers/assets_controller.rb
callumknights: Scratch previous. I'll be removing the parameter editing anyway.

spec/controllers/assets_controller_spec.rb
callumknights: This is only a start, I wanted to make sure this worked on one of the simpler endpoints first. But I'd appreciate a glance at the testing approach [here](https://github.com/alphagov/asset-manager/pull/2029/changes/bfecda8943324d1164df81c2974a2c3e99c97297#diff-2022bd034dee9cc41910d5eb86e8e58694733b5e97166128c35693f56a72f0bcR526-R584) and if you think it's easier to follow.

app/models/asset.rb
```diff
@@ -134,6 +159,10 @@ def accessible_by?(user)
     access_limited.include?(user.uid) || access_limited_organisation_ids.include?(user.organisation_content_id)
   end
 
+  def manageable_by?(user)
+    self.user == user || user.permissions.include?("Manage all Assets")
+  end
```
brucebolt, line 140: Could this be broken down into more than one commit, as it is doing more than detailed in the commit message here.

app/models/asset.rb
```diff
@@ -134,6 +159,10 @@ def accessible_by?(user)
     access_limited.include?(user.uid) || access_limited_organisation_ids.include?(user.organisation_content_id)
   end
 
+  def manageable_by?(user)
+    self.user == user || user.permissions.include?("Manage all Assets")
```
brucebolt, line 139: Is this new permission documented somewhere?  Maybe the commit that first references it should explain why it was created and which apps were given it.

spec/controllers/assets_controller_spec.rb
```diff
@@ -221,304 +221,454 @@
...
+            expect(asset.reload).not_to be_draft
+          end
         end
+      end
 
-        it "changes draft state for draft deleted asset" do
-          asset.update!(deleted_at: 1.hour.ago, draft: true)
+      context "with an existing asset not owned by that user" do
+        let(:user) { FactoryBot.create(:user) }
+        let(:asset) { FactoryBot.create(:asset, user:) }
```
brucebolt: Could this be moved into the parent context block, as it is repeated a couple of  times?

spec/controllers/assets_controller_spec.rb
```diff
@@ -221,304 +221,454 @@
...
+          delete :destroy, params: { id: asset.id }
 
-        expect(response).to have_http_status(:success)
+          expect(response).to have_http_status(:success)
+        end
+      end
+
+      context "with an asset not owned by the user" do
+        let(:user) { FactoryBot.create(:user) }
+        let(:asset) { FactoryBot.create(:asset, user:) }
```
brucebolt: Same here with the repetition on creating these.  Could they be moved into the parent context block, unless this context is changing one of them.

brucebolt (commented):
This is looking good, thanks for making the changes.

I've left a few comments that should be fairly easy to resolve.

spec/controllers/assets_controller_spec.rb
callumknights: Done 👍

spec/controllers/assets_controller_spec.rb
callumknights: Done 👍

app/models/asset.rb
callumknights: I think that's fair actually. I'll break it down and add the permission logic in a subsequent commit. Will make it easier to document per your comment below as well.

app/models/asset.rb
callumknights: Hopefully the commit message is clearer now

spec/models/asset_spec.rb
```diff
@@ -350,6 +350,17 @@
     end
   end
 
+  describe "#manageable_by?" do
+    context "when the user has the correct permissions" do
```
brucebolt: This covers the route of the user having the new permission, but it doesn't cover the user not having that permission (which is the other part of the conditional in the method).

I think this needs another context for the user not having the permission, then two tests: for when they are the creating user and another for when they're not.

callumknights: Missed that, added in

app/models/asset.rb
```diff
@@ -17,6 +17,7 @@ class Asset
   index deleted_at: 1
 
   belongs_to :replacement, class_name: "Asset", optional: true, index: true
+  belongs_to :user, optional: true
```
brucebolt, line 20: Nitpick: there's a typo in the commit message `belong_to` --> `belongs_to`.

spec/models/asset_spec.rb
brucebolt: It's still missing the scenario of the user not having the permission, but being the user who created the asset.

app/controllers/assets_controller.rb
```diff
@@ -7,13 +7,17 @@ class AssetsController < ApplicationController
...
+    return error_403 unless @asset.manageable_by?(current_user)
+
     expires_now
     render json: AssetPresenter.new(@asset, view_context)
   end
 
   def create
     @asset = build_asset
 
+    @asset.user = current_user
```
brucebolt, line 19: I'm wondering if this should also be a separate commit, as this commit is about adding restrictions to the API endpoints, whereas this line associates the user with the asset?

I may have missed it, but I think there's no test for this.

spec/models/asset_spec.rb
callumknights: Oh I misunderstood, fair enough

callumknights: Ignore please

brucebolt: Thanks for adding this.

I think I'd suggest putting the three tests into two context blocks to make it absolutely clear there are two possible scenarios regarding the permission and explicitly naming the permission in the test (otherwise future us may not be sure what permission is being referenced):
- when the user has the manage all assets permission
- when the user does not have the manage all assets permission

Then in the second of those contexts, have two sub-contexts:
- when the user created the asset (there's a typo in the current context name)
- when the user did not create the asset

This will mean you only need to have `let(:user) { FactoryBot.build(:user) }` once (under the 'when the user does not have the manage all assets permission' context).

callumknights: Okay now this is corrected

callumknights: Sorry I misunderstood the comment on first read

app/controllers/assets_controller.rb
callumknights: Here you go https://github.com/alphagov/asset-manager/pull/2029/changes/368fcbc3028421e29ef645ccb64a74254e07f437#diff-2022bd034dee9cc41910d5eb86e8e58694733b5e97166128c35693f56a72f0bcR22-R26

spec/controllers/assets_controller_spec.rb
```diff
@@ -19,6 +19,12 @@
         expect(assigns(:asset)).to be_persisted
       end
 
+      it "associates current user to asset" do
+        post :create, params: { asset: valid_attributes }
+
+        expect(assigns(:asset).user).to match(stub_user)
```
brucebolt, line 25: The `assigns` in this test is testing if the controller assigns an instance variable called `@asset`, rather than testing whether it was persisted in the database.

In other apps, we'd do something like `expect(Asset.last.user).to eq(stub_user)`.  Doing this means the implementation detail can be changed in the method without needing to rewrite the test, as it doesn't matter how we go about doing it in the controller.

Although the `assigns` approach seems to be used consistently in this app for some reason, so this should be fine, but maybe we want to consider changing all the tests in another PR.

.byebug_history
```diff
@@ -0,0 +1,16 @@
+continue
```
brucebolt: I assume you didn't intend to commit this file?

spec/models/asset_spec.rb
```diff
@@ -350,6 +350,46 @@
...
+        end
+      end
+
+      context "and the asset is not owned by the user" do
+        it "returns false" do
+          expect(asset).not_to be_manageable_by(user)
+        end
+      end
+    end
+  end
```
brucebolt, line 395: There are two variables in these tests: the user who is trying to manage the asset and the the user who created the asset.

So that's the only thing that needs to be changed between context blocks.

Instead of declaring `asset` in a few places, I'd declare it once at the top of the `describe` block then modify these two variables as needed in each context.

This way, it makes it very clear what is being changed in each test case.

Maybe something like:

```suggestion
  describe "#manageable_by?" do
    let(:asset) { FactoryBot.build(:asset, user: asset_owner) }

    context "when the user is a managing user" do
      let(:user) { FactoryBot.build(:managing_user) }

      context "and the asset is owned by the user" do
        let(:asset_owner) { user }

        it "returns true" do
          expect(asset).to be_manageable_by(user)
        end
      end

      context "and the asset is not owned by the user" do
        let(:asset_owner) { FactoryBot.build(:user) }

        it "returns true" do
          expect(asset).to be_manageable_by(user)
        end
      end
    end

    context "when the user is a regular user" do
      let(:user) { FactoryBot.build(:user) }

      context "and the asset is owned by the user" do
        let(:asset_owner) { user }

        it "returns true" do
          expect(asset).to be_manageable_by(user)
        end
      end

      context "and the asset is not owned by the user" do
        let(:asset_owner) { FactoryBot.build(:user) }

        it "returns false" do
          expect(asset).not_to be_manageable_by(user)
        end
      end
    end
  end
```

spec/controllers/assets_controller_spec.rb
```diff
@@ -227,304 +227,454 @@
...
 
   describe "DELETE destroy" do
-    context "with an existing asset" do
-      let(:asset) { FactoryBot.create(:asset) }
+    context "when authenticated as a regular user" do
+      before do
+        login_as_stub_user
+      end
+
+      let(:asset) { FactoryBot.create(:asset, user:) }
```
brucebolt: This is being declared here and on line 518.  Could they both be declared just once in the parent block?

spec/controllers/assets_controller_spec.rb
```diff
@@ -227,304 +227,454 @@
...
+    context "when authenticated as a regular user" do
+      before do
+        login_as_stub_user
+      end
 
-      it "responds with success status" do
-        get :show, params: { id: asset.id }
+      context "with an asset owned by the user" do
+        let(:user) { stub_user }
+        let(:asset) { FactoryBot.create(:asset, user:) }
```
brucebolt: Same as my previous comment: this is declared both here and line 569, so can be moved into the parent context.

spec/controllers/assets_controller_spec.rb
```diff
@@ -227,304 +227,454 @@
...
-
-        expect(assigns(:asset)).to be_persisted
+    context "when authenticated as a regular user" do
+      before do
+        login_as_stub_user
       end
 
-      it "stores file on existing asset" do
-        put :update, params: { id: asset.id, asset: valid_attributes }
+      let(:asset) { FactoryBot.create(:asset, user:) }
```
brucebolt: Same as my previous comment, here and line 455 declare the same thing so it can be moved into the parent context.

brucebolt (commented):
I've left a few comments regarding minor changes to the placing of `let` blocks in tests, and there's a development file that you probably didn't intend to add.

spec/models/asset_spec.rb
callumknights: Thanks for the suggestion. It hadn't occurred to me that I could set the `user` for a given `asset` as a yet to be defined in a following example variable. This is really nice and some good learning for me.

spec/controllers/assets_controller_spec.rb
callumknights: Yeah I found the use of `assigns` a bit odd as well. But I thought it would be odd using an entirely different testing approach just for one endpoint. Like you say, maybe one for a refactor PR.

brucebolt (approved):
This looks good to me.  Thanks for making all those changes.

PR #2047, opened 2026-09-14
Remove `.gitkeep` files

brucebolt:
These were added in f5d77057d6c95ef91a5d6ad48bfea04a1a691b15 when the application was first created.

They appear to serve no purpose, as most of the directories are either not required or already contain files.

Therefore removing these to reduce clutter when browsing through the directories.


callumknights (approved):
Good spot. Nice to reduce clutter a bit

